PR #33, opened 2026-01-18
feat: add JSONC config file support

xLillium:
## Summary

Adds support for JSONC (JSON with Comments) configuration files during CLI install.

- **JSONC parsing**: New `stripJsonComments()` function handles single-line (`//`), multi-line (`/* */`) comments, and trailing commas
- **Fallback support**: Reads `opencode.jsonc` if `opencode.json` doesn't exist
- **Preserves user choice**: Writes back to whichever config file exists (`.json` or `.jsonc`)
- **DRY refactor**: Reuses `stripJsonComments` in auto-update-checker instead of duplicating

## Changes

- `src/cli/config-manager.ts`: Core JSONC support with robust comment stripping
- `src/cli/config-manager.test.ts`: Comprehensive test suite (20 tests covering edge cases)
- `src/hooks/auto-update-checker/checker.ts`: Import shared `stripJsonComments` instead of local copy

## Testing

All 20 tests pass:
```
bun test src/cli/config-manager.test.ts
✓ 20 pass | 0 fail
```

### DIFF
--- a/src/cli/config-manager.test.ts
+++ b/src/cli/config-manager.test.ts
@@ -1,0 +1,190 @@
+import { describe, expect, test } from "bun:test"
+import { stripJsonComments } from "./config-manager"
+
+describe("stripJsonComments", () => {
+  test("returns unchanged JSON without comments", () => {
+    const json = '{"key": "value"}'
+    expect(stripJsonComments(json)).toBe(json)
+  })
+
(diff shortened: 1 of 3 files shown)


martokk:
Does this work with `oh-my-opencode-slim.jsonc`?

PR #133, opened 2026-02-16
fix: support variant overrides in agent config

MuenYu:
## Summary
- Fix agent configuration overrides to support `variant` values in addition to `model` and `temperature`.
- Ensure backward-compatible alias configs (e.g. `explore`) can override `variant` correctly.
- Add regression tests to verify variant override behavior for both explorer and orchestrator agents.

## Changes
- Updated override application logic in `src/agents/index.ts` to apply `override.variant` to agent configs.
- Added tests in `src/agents/index.test.ts`:
  - `variant override via old alias`
  - `orchestrator accepts variant override`
- Switched `AgentConfig` type imports to `@opencode-ai/sdk/v2` in:
  - `src/agents/index.ts`
  - `src/agents/orchestrator.ts`
- Applied formatting-only cleanup in `src/tools/ast-grep/cli.ts` and whitespace normalization in `src/agents/orchestrator.ts`.

### DIFF
--- a/src/agents/index.test.ts
+++ b/src/agents/index.test.ts
@@ -66,6 +66,17 @@ describe('agent alias backward compatibility', () => {
     const explorer = agents.find((a) => a.name === 'explorer');
     expect(explorer?.config.temperature).toBe(0.5);
   });
+
+  test('variant override via old alias', () => {
+    const config: PluginConfig = {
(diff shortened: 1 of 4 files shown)


alvinunreal:
thanks @MuenYu

PR #145, opened 2026-02-23
feat: agent delegation rules, skill permission pipeline, and test improvements

pelidan:
## Summary

- intention of these changes is to make OmOS more compatible with superpowers
- Expanded the skill permission pipeline so bundled custom skills (cartography) and externally-managed skills (requesting-code-review) are automatically granted to the correct agents at startup
- Allowed @fixer to spawn @explorer for codebase lookups, and updated the fixer prompt to reflect this

## Changes

- src/cli/skills.ts — introduced PermissionOnlySkill interface and PERMISSION_ONLY_SKILLS array for skills that need permission grants but are not installed by the plugin CLI; extended getSkillPermissionsForAgent to loop over both CUSTOM_SKILLS and PERMISSION_ONLY_SKILLS when building per-agent permissions
- src/cli/custom-skills.ts — added 'explorer' to the cartography skill's allowedAgents list so explorer can invoke the cartography skill
- src/agents/orchestrator.ts — added an Agent Role Mapping section to the orchestrator prompt clarifying that @fixer maps to the implementer role and @oracle maps to the reviewer role in SDD workflows
- src/agents/fixer.ts — updated fixer's constraints to allow spawning @explorer for codebase lookups, removing the hard "NO delegation" rule
- src/agents/index.test.ts — added three skill-permission tests verifying orchestrator gets cartography + wildcard allow, explorer gets cartography, and oracle gets requesting-code-review
- .gitignore — added opencode.jsonc to prevent local config from being committed

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -27,8 +27,8 @@ export type AgentName = (typeof ALL_AGENT_NAMES)[number];
 // Unknown agent types not listed here default to explorer-only access
 export const SUBAGENT_DELEGATION_RULES: Record<AgentName, readonly string[]> = {
   orchestrator: SUBAGENT_NAMES,
-  fixer: [], // leaf node — prompt forbids delegation; use grep/glob for lookups
-  designer: ['explorer'],
+  fixer: [],
+  designer: [],
   explorer: [],
   librarian: [],
   oracle: [],


src/agents/fixer.ts
```diff
@@ -14,9 +14,9 @@ const FIXER_PROMPT = `You are Fixer - a fast, focused implementation specialist.
 
 **Constraints**:
 - NO external research (no websearch, context7, grep_app)
-- NO delegation (no background_task)
+- NO delegation (no background_task) except spawning @explorer for codebase lookups
 - No multi-step research/planning; minimal execution sequence ok
-- If context is insufficient, read the files listed; only ask for missing inputs you cannot retrieve
+- If context is insufficient, read the files listed or delegate a lookup to @explorer; only ask for missing inputs you cannot retrieve
```
alvinunreal: kinda leaves a bit vague "or" is when, maybe we can provide some clarity

pelidan: that's right, fixer can't read any files himself, that was a possible other solution (delegate to explorer or give him permissions for grep etc.)

pelidan: had to update another sentence too as it was mentioning reading files too ... I'm not 100% sure about this, up to you @alvinunreal  to decide - whether it's strict delegation to explorer for any reads (including the files that are going to be updated) oh whether fixer should be able to read some himself ... but then where is the line between own reading and delegation, right?

alvinunreal: fixer is able to read files
Screenshot 2026-02-26 at 13 13 43

Cross agent calling is a different case completely - so far avoiding since it might complicate usage very quickly very soon.

alvinunreal: PR looks good, just for now not sure if we need explorer calling by fixer; let fixer read itself what it needs and avoid extra steps.

pelidan: that is fine I had that as my first solution to the problem, having fixer as a leaf node
but then the rest of the code appeared to me as it was rather prepared for delegation to explorer so I changed it to this to be less invasive :slightly_smiling_face: 

let me dig the original version out from history

alvinunreal:
@greptile-app

PR #200, opened 2026-03-23
chore: remove ripgrep-based grep tool

### LINKED ISSUE #197: Remove Grep Tool
### Discussed in https://github.com/alvinunreal/oh-my-opencode-slim/discussions/196

<div type='discussions-op-text'>

<sup>Originally posted by **adikpb** March 22, 2026</sup>
Opencode already has a grep tool exposed by default, and it seems to use ripgrep underneath, just like the grep tool exposed via this plugin. I suggest we remove it and let opencode handle the grep stuff, and also not download and install ripgrep on behalf of the user.</div>

adikpb:
Closes #197 

## Summary:
- Removed the ripgrep-based grep tool since OpenCode already has a built-in grep tool
- Eliminated ripgrep download/install on behalf of users

## Changes:
- Deleted src/tools/grep/ directory (11 files including tool definition, downloader, constants, and tests)
- Removed grep export and usage from src/tools/index.ts and src/index.ts
- Removed 'grep' from JSON error recovery exclude list in src/hooks/json-error-recovery/hook.ts


alvinunreal:
Do we need at all to mention in prompt? isn't available tools injected by opencode anyway?

adikpb:
> Do we need at all to mention in prompt? isn't available tools injected by opencode anyway?

That is true, making the changes right now. Also do I go ahead and run the cartography skill, to update the codemaps?

alvinunreal:
@greptileai

PR #261, opened 2026-04-10
feat: add serve-mode orchestrator prompt injection + fix CJS build for GitHub installs

### LINKED ISSUE #260: Plugin agents not accessible via OpenCode serve API
## Problem

When using `opencode serve` (headless HTTP server mode), plugin-provided agents (orchestrator, oracle, librarian, etc.) are **not accessible** via the serve API — they don't appear in `GET /agent` and sending `agent: "orchestrator"` in `POST /session/:id/message` produces no response.

## Root Cause

The issue lies in OpenCode's Effect layer initialization order:

1. `GET /agent` calls `Agent.list()` which uses `Agent.defaultLayer` — this layer does **not** include `Plugin.defaultLayer`
2. The agent state (agents map) is built **once** from raw config before plugin config hooks run
3. The frozen agents map lacks plugin-provided agents (orchestrator, etc.) and their system prompts
4. Even in session context where Plugin is included, Effect initializes Plugin and Agent concurrently — there's no guaranteed ordering, so the config hook may not run before the agents map is built

## Also Found

A pre-existing bug prevents the plugin from loading in serve mode entirely due to CJS/ESM interop failures (`jsdom`, `turndown`, `vscode-jsonrpc`) in OpenCode's embedded Bun runtime. The `--packages external` build flag leaves these as runtime imports that Bun's strict ESM resolver rejects.

## Fix (PR #N)

Two-part fix:

**1. Serve-mode system prompt injection** (`src/index.ts`, `src/agents/orchestrator.ts`):
- Track `sessionID → agentName` via `chat.message` hook
- Inject `ORCHESTRATOR_PROMPT` at LLM call time via `experimental.chat.system.transform` hook — bypasses the frozen agents map entirely

With a minimal entry in `opencode.json` (`"orchestrator": {"model": "openai/gpt-5.4"}`), the agent becomes visible in `GET /agent` and callable via the serve API with full orchestrator behavior.

**2. Build fix** (`package.json`):
- Add `prepare` script using bundled build (`--external` flags instead of `--packages external`) so the plugin loads correctly when installed from GitHub

## Testing

Verified both scenarios pass with the fix:
- `GET /agent` first → session message: orchestrator responds with full persona
- Session message first (no `/agent`): same result

Without the fix, the orchestrator prompt is never injected in either scenario.

atomlink-ye:
Fixes #260

## Summary

Plugin-provided agents (orchestrator, etc.) are not accessible via `opencode serve` API due to two issues:

1. **Agent state timing**: The agents map is built once from raw config before plugin config hooks run, and `GET /agent` never triggers Plugin initialization. The orchestrator prompt is therefore never injected.

2. **Plugin load failure**: The `--packages external` build leaves CJS imports (`jsdom`, `turndown`, `vscode-jsonrpc`) as runtime imports that fail in OpenCode's embedded Bun runtime, preventing the plugin from loading in serve mode entirely.

### `src/agents/orchestrator.ts`
- Export `ORCHESTRATOR_PROMPT` constant (previously unexported)

### `src/index.ts`
- Add `sessionAgentMap: Map<string, string>` to track session → agent
- Add `chat.message` hook to populate the map
- Add `experimental.chat.system.transform` hook: when a session uses the `orchestrator` agent, inject the full `ORCHESTRATOR_PROMPT` at LLM call time — bypasses the frozen agents map entirely

### `package.json`
- Add `prepare` script with bundled build (`--external @ast-grep/napi --external @opencode-ai/plugin --external @opencode-ai/sdk`) for GitHub installations
- Original `build` script is **unchanged**

## How it works

Add a minimal entry to `opencode.json` so the agent is visible in `GET /agent`:
```json
{
  "agent": {
    "orchestrator": { "model": "openai/gpt-5.4" }
  }
}
```

The `experimental.chat.system.transform` hook then injects the full orchestrator system prompt at LLM call time, regardless of whether `GET /agent` was called first.

## Test plan

- [x] `GET /agent` returns orchestrator (from raw config entry)
- [x] `POST /session/:id/message` with `agent: "orchestrator"` responds with full orchestrator persona and specialist knowledge (@explorer, @librarian, @oracle, @fixer stats)
- [x] Works when `GET /agent` is called before any session (edge case that bypasses Plugin initialization)
- [x] Works when session message is sent first
- [x] Plugin loads successfully from GitHub install via `prepare` script

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   ],
   "scripts": {
     "build": "bun build src/index.ts --outdir dist --target bun --format esm --packages external && bun build src/cli/index.ts --outdir dist/cli --target bun --format esm --packages external && tsc --emitDeclarationOnly && bun run generate-schema",
+    "prepare": "bun build src/index.ts --outdir dist --target bun --format esm --external @ast-grep/napi --external @opencode-ai/plugin --external @opencode-ai/sdk",
     "contributors:add": "all-contributors add",
     "contributors:check": "all-contributors check",
     "contributors:generate": "all-contributors generate",
--- a/src/agents/orchestrator.ts
+++ b/src/agents/orchestrator.ts
@@ -23,7 +23,7 @@ export function resolvePrompt(
   return base;
 }
 
-const ORCHESTRATOR_PROMPT = `<Role>
+export const ORCHESTRATOR_PROMPT = `<Role>
 You are an AI coding orchestrator that optimizes for quality, speed, cost, and reliability by delegating to specialists when it provides net efficiency gains.
 </Role>
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -164,6 +164,9 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
     config.fallback?.enabled !== false && Object.keys(runtimeChains).length > 0,
   );
 
+  // Track session → agent mapping for serve-mode system prompt injection
+  const sessionAgentMap = new Map<string, string>();
+
   // Initialize todo-continuation hook (opt-in auto-continue for incomplete todos)
   const todoContinuationHook = createTodoContinuationHook(ctx, {
     maxContinuations: config.todoContinuation?.maxContinuations ?? 5,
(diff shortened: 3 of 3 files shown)


alvinunreal:
thanks @atomlink-ye

PR #270, opened 2026-04-12
fix(hooks): harden todo auto-continuation scheduling

dhaern:
## Summary

Harden `todo-continuation` so auto-continue is routed to the right orchestrator session and the countdown notification cannot cancel its own scheduled continuation.

This keeps the existing auto-continue behavior and prompt style, but fixes the scheduling/routing path:

- tracks orchestrator sessions from `chat.message`
- uses the resolved agent from the message payload when OpenCode falls back to the default agent
- supports multiple orchestrator sessions without a single-session lockout
- preserves the legacy first-idle fallback when no agent-bearing `chat.message` is available
- treats `session.status` with `idle` the same as `session.idle`
- prevents the `noReply` countdown notification from cancelling the continuation timer through its own `busy` transition
- keeps subagent busy/delete events from cancelling orchestrator continuations

## Motivation

`todo-continuation` relied on a single session ID inferred from idle ordering. That is fragile in real OpenCode sessions with subagents or multiple sessions, because idle ordering is not a reliable source of agent identity.

OpenCode also emits the resolved agent on the created user message. Using that signal avoids treating subagent sessions or stale sessions as the orchestrator.

There was also a scheduling race around the countdown notification. The hook sends a `noReply` notification before the continuation prompt; that notification can briefly move the session to `busy`. The hook now distinguishes that self-generated notification from external activity so it does not cancel the pending continuation timer.

## Scope

This change is intentionally limited to `todo-continuation`.

It does **not**:

- change the auto-continue prompt text
- change `autoEnableThreshold`
- change `maxContinuations`
- change todo semantics
- change models or agent routing

### Orchestrator session tracking

Added explicit `chat.message` routing into the hook.

The hook now tracks orchestrator sessions in a `Set` instead of relying on one global first-idle session ID.

When OpenCode omits `input.agent` because it resolved the default agent internally, the plugin now uses `output.message.agent` from the hook payload.

The existing first-idle fallback remains available only when no agent-bearing `chat.message` has been seen, preserving compatibility with older event flows.

### Idle event compatibility

The hook now handles both:

- `session.idle`
- `session.status` with `status.type === "idle"`

This matches OpenCode’s current event flow while preserving the existing legacy event support.

### Countdown notification guard

Added a small `isNotifying` guard around the `noReply` countdown notification.

While the notification prompt is in flight, a `busy` transition from that same notification no longer cancels the pending continuation timer. Abort/error handling remains unchanged.

## Safety model

The hook still keeps the same guardrails:

- only orchestrator sessions can auto-continue
- subagent busy/delete events do not cancel orchestrator timers
- user abort/error events suppress continuation
- pending timers are cancelled when auto-continue is disabled
- max consecutive continuation limits still apply
- incomplete todos are still checked before continuing
- questions still stop auto-continuation

## Validation

Validated with:

- `bun test src/hooks/todo-continuation/index.test.ts`
- `bun run typecheck`
- `bun run check:ci`

The new tests cover:

- orchestrator registration via `chat.message`
- multiple orchestrator sessions without first-idle lockout
- `chat.message` without an agent preserving the legacy fallback
- subagent `chat.message` preventing accidental first-idle orchestrator registration
- `session.status idle` triggering like `session.idle`
- countdown notification `busy` events not cancelling the pending continuation timer

## Notes

This is a scheduling/routing fix only.

It avoids changing the model-facing auto-continue instruction so the behavior stays compatible with existing users while making the hook less fragile in multi-session and subagent-heavy workflows.

### DIFF
--- a/src/hooks/todo-continuation/index.test.ts
+++ b/src/hooks/todo-continuation/index.test.ts
@@ -1897,6 +1897,153 @@ describe('createTodoContinuationHook', () => {
     });
   });
 
+  describe('session routing and notification cancellation', () => {
+    function createPendingCtx() {
+      return createMockContext({
+        todoResult: {
+          data: [
+            { id: '1', content: 'todo1', status: 'pending', priority: 'high' },
+          ],
+        },
+        messagesResult: {
+          data: [
+            {
+              info: { role: 'assistant' },
+              parts: [{ type: 'text', text: 'Work in progress' }],
+            },
+          ],
+        },
+      });
+    }
+
+    test('chat.message registers orchestrator sessions without first-idle lockout', async () => {
+      const ctx = createPendingCtx();
+      const hook = createTodoContinuationHook(ctx, {
+        cooldownMs: 50,
+      });
+      await hook.tool.auto_continue.execute({ enabled: true });
+
+      hook.handleChatMessage({ sessionID: 'sub1', agent: 'fixer' });
+      hook.handleChatMessage({ sessionID: 'main1', agent: 'orchestrator' });
+      hook.handleChatMessage({ sessionID: 'main2', agent: 'orchestrator' });
+
+      await hook.handleEvent({
+        event: {
+          type: 'session.idle',
+          properties: { sessionID: 'sub1' },
+        },
+      });
+      await delay(60);
+      expect(hasContinuation(ctx.client.session.prompt)).toBe(false);
+
+      await hook.handleEvent({
+        event: {
+          type: 'session.idle',
+          properties: { sessionID: 'main2' },
+        },
+      });
+      await delay(60);
+
+      expect(hasContinuation(ctx.client.session.prompt)).toBe(true);
+      expect(contCall(ctx.client.session.prompt)[0].path.id).toBe('main2');
+    });
+
+    test('chat.message without agent does not block legacy first-idle fallback', async () => {
+      const ctx = createPendingCtx();
+      const hook = createTodoContinuationHook(ctx, {
+        cooldownMs: 50,
+        autoEnable: true,
+        autoEnableThreshold: 1,
+      });
+
+      hook.handleChatMessage({ sessionID: 'main1' });
+      await hook.handleEvent({
+        event: {
+          type: 'session.idle',
+          properties: { sessionID: 'main1' },
+        },
+      });
+      await delay(60);
+
+      expect(hasContinuation(ctx.client.session.prompt)).toBe(true);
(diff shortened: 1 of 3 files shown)


dhaern:
@greptile-apps[bot] thanks, both findings were valid and are fixed in `90ba1ed`.

Changes:

- Added `pendingTimerSessionId` so `session.deleted` only cancels the pending timer when the deleted session owns that timer.
- Changed orchestrator deletion cleanup so removing one orchestrator session no longer resets shared continuation state for other registered orchestrator sessions.
- Reset `sawChatMessage` when the registered orchestrator session set becomes empty, restoring the legacy first-idle fallback for later sessions.
- Added regression tests for both cases:
  - deleting another orchestrator does not cancel the active session timer
  - deleting all orchestrators restores legacy first-idle fallback

Validation rerun:

- `bun test src/hooks/todo-continuation/index.test.ts`
- `bun run typecheck`
- `bun run check:ci`

dhaern:
@greptile-apps[bot] fixed in `13ae32c`.

The `busy` handler now applies the same `!state.isNotifying` guard to the consecutive-continuation reset path, so the countdown notification cannot reset `consecutiveContinuations` and bypass `maxContinuations`.

Added a regression test that runs two notification/continuation cycles with `maxContinuations: 2`, emits `busy` during the countdown notification, and verifies the third idle does not schedule another continuation.

Validation rerun:

- `bun test src/hooks/todo-continuation/index.test.ts`
- `bun run typecheck`
- `bun run check:ci`

dhaern:
@greptile-apps[bot] addressed in `5525efc`.

I replaced the single global `isNotifying` boolean with session-scoped notification tracking:

- `notifyingSessionIds` tracks in-flight countdown notifications per session.
- `notificationBusyUntilBySession` keeps a short 250ms grace window after notification completion, covering HTTP/SSE ordering where a `busy` event arrives just after the noReply response resolves.
- `clearNotificationState(sessionID)` removes both in-flight and grace-window state when the continuation timer fires or the session is deleted.

This keeps the existing countdown notification and prompt behavior, but removes the boolean-vs-stale-callback fragility Greptile called out.

I also added coverage for:

- a late countdown `busy` event still not cancelling the continuation timer
- countdown `busy` events not resetting the max-continuation counter
- real non-notification `busy` after the grace window still cancelling a pending timer

Validation rerun:

- `bun test src/hooks/todo-continuation/index.test.ts`
- `bun run typecheck`
- `bun run check:ci`

Issue #280, opened 2026-04-12
Stale incomplete todos can survive phase changes and delegation in orchestrator sessions


dhaern:
## Summary
The orchestrator can leave **stale incomplete todos** behind while work in the same run has already moved on.
This is not limited to delegation, and it is not limited to continuation-related features. It can also happen in normal orchestrator-only flows.
The result is that the todo list can stop matching the actual current phase of the work.
## Observed behavior
A common stale state looks like this:
- the first todo remains `in_progress`
- later todos remain `pending`
- the orchestrator has already moved on mentally
- the run continues through more tools / reasoning / integration / response
- but the todo list was never updated to match that progress
In other words, incomplete todos can survive even though they no longer represent the current active step.
## Why this matters
Once the todo list becomes stale:
- later orchestration decisions can be made from an outdated task state
- progress becomes harder to reason about
- users see an incorrect task state
- continuation-related behavior may amplify the bad state, even if it is not the root cause
## What investigation suggests so far
This does not look like a problem caused only by slim hooks.
From inspecting OpenCode core, the todo lifecycle appears to be mostly model-driven:
- `todowrite` replaces the stored todo list with whatever the model sends
- the stored state is persisted as-is
- there does not appear to be an automatic reconciliation/finalization layer in core
- there does not appear to be a built-in mechanism that forces the active todo to be advanced/closed correctly when the real work phase changes
That means stale todo state can survive if the model fails to update it in time.
Slim may expose or amplify the problem in some flows, but the underlying stale-todo condition appears broader than slim itself.
## Suspected areas
Likely relevant areas include:
### OpenCode core
- `todowrite` behavior and todo persistence model
- lack of automatic todo reconciliation/finalization during an active run
- reliance on prompting/model behavior rather than enforced todo-state transitions
### Slim / orchestration layer
- orchestrator flow / prompt behavior around todo hygiene
- delegated result return paths (`task`, `background_output`)
- intra-turn reminders or nudges
- any missing reconciliation/finalization step while a run is still active
## Typical situations where it seems more likely
- long / inflated sessions
- phase changes after research or integration
- delegated work returning to the parent session
- runs where the model keeps working but does not refresh the todo list in time
## Expected behavior
If the current step changed or finished, the todo list should be kept in sync during the same run.
The orchestrator should not keep carrying stale `in_progress` / `pending` items that no longer reflect the actual current work state.
## Scope
[...]

PR #283, opened 2026-04-13
improve intra-turn todo hygiene for stale incomplete todos

dhaern:
## Summary

- add an intra-turn todo hygiene reminder for orchestrator sessions
- reduce stale `in_progress` / `pending` todo state during active runs
- strengthen the closing case where the last active todo may be left `in_progress`

## Changes

- add `src/hooks/todo-continuation/todo-hygiene.ts`
- wire todo hygiene into `todo-continuation`
- inject a silent ephemeral system reminder during the same run
- reset the hygiene cycle on `todowrite`
- re-arm a stronger finishing reminder when only one active todo remains
- keep the reminder best-effort and fail-closed
- add/update tests for routing, reset/re-arm behavior, finishing behavior, and failure handling

## Why

Open orchestrator runs can leave stale incomplete todos behind even when the real work has already moved on.

This PR adds a small intra-turn reminder layer to help the model keep the todo list aligned with the current work state, especially around:
- phase changes
- delegated result returns
- final response closing

It is implemented as an expansion of `todo-continuation`, so it remains compatible with the existing continuation flow when unfinished todos still legitimately remain after a turn.

## Validation

- `bun test src/hooks/todo-continuation/index.test.ts src/hooks/todo-continuation/todo-hygiene.test.ts`
- `bun run build`

## Related

- addresses #280

### DIFF
--- a/src/hooks/todo-continuation/index.test.ts
+++ b/src/hooks/todo-continuation/index.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, mock, test } from 'bun:test';
 import { SLIM_INTERNAL_INITIATOR_MARKER } from '../../utils';
 import { createTodoContinuationHook } from './index';
+import {
+  TODO_FINAL_ACTIVE_REMINDER,
+  TODO_HYGIENE_REMINDER,
+} from './todo-hygiene';
 
 describe('createTodoContinuationHook', () => {
   function createMockContext(overrides?: {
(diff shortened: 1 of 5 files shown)


dhaern:
@greptile-apps I addressed the two review findings and pushed the follow-up commit (`6dd73fb`). Could you please re-analyze the PR?

PR #289, opened 2026-04-13
fix(auto-update): refresh host-managed package cache

huilang021x:
## Summary
- resolve the active install root from the running plugin package so auto-update targets OpenCode's `~/.cache/opencode/packages/<specifier>` wrapper cache instead of the legacy global cache root
- rewrite the wrapper package dependency, clear the active install root's package artifacts, and run `bun install` in that same directory before showing the restart toast
- add regression tests for packages-cache installs, no-context handling, and the no-legacy-fallback path
## Validation
- bun test src/hooks/auto-update-checker
- bun run typecheck
- bun run build


alvinunreal:
Thanks @huilang021x

PR #299, opened 2026-04-14
fix: replace Bun-only APIs for Node.js (Electron desktop) compatibility

### LINKED ISSUE #298: [Bug] v0.9.12 fails to load
After upgrading to 0.9.12, omo-slim doesn't successfully load.

I notice the plugin config of opencode is "oh-my-opencode-slim" now, no longer a static file path. I suspect there's a bug behind this.

whackur:
## Summary

- Replace all Bun-specific APIs (`import { spawn } from 'bun'`, `Bun.write()`, `Bun.spawn()`) with cross-runtime alternatives (`node:child_process`, `node:fs/promises`)
- Add `src/utils/compat.ts` with `crossSpawn()` and `crossWrite()` utilities
- Change build target from `--target bun` to `--target node` to avoid Bun-specific `__require` shim

## Problem

The OpenCode desktop app (Electron) runs the server **in-process using Node.js**, not Bun. When the plugin is loaded via `import()`, the bundled `from 'bun'` imports fail immediately because the `bun` module doesn't exist in Node.js. Additionally, `--target bun` produces a `__require` helper that is Bun-specific and throws `__require is not a function` in Node.js.

The plugin works fine in CLI mode because the CLI uses a Bun binary.

## Changes

| File | Change |
|------|--------|
| `src/utils/compat.ts` | **New** — `crossSpawn()`, `crossWrite()` using `node:child_process` + `node:fs/promises` |
| `src/tools/ast-grep/cli.ts` | `spawn` from `bun` → `crossSpawn` |
| `src/tools/ast-grep/downloader.ts` | `Bun.write()` → `crossWrite()` |
| `src/tools/lsp/client.ts` | `spawn` from `bun` → `node:child_process`, removed redundant stream wrappers |
| `src/multiplexer/tmux/index.ts` | `spawn` from `bun` → `crossSpawn` |
| `src/multiplexer/zellij/index.ts` | `spawn` from `bun` → `crossSpawn` |
| `src/utils/zip-extractor.ts` | `spawn`/`spawnSync` from `bun` → `node:child_process` |
| `src/cli/system.ts` | `Bun.spawn()` → `crossSpawn` |
| `src/hooks/auto-update-checker/index.ts` | `Bun.spawn()` → `crossSpawn` |
| `package.json` | build target `--target bun` → `--target node` |

## Why this is safe

`node:child_process` and `node:fs/promises` are **fully supported in both Bun and Node.js**. CLI mode (Bun) continues to work exactly as before — verified with all 532 tests passing.

## Test plan

- [x] `bun run typecheck` — passes
- [x] `bun test` — 532 tests pass, 0 failures
- [x] `bun run build` — 248 modules bundled successfully
- [x] Built `dist/index.js` contains zero `from "bun"` imports
- [x] Tested in OpenCode desktop app (Electron) — plugin loads successfully

Fixes #298

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "LICENSE"
   ],
   "scripts": {
-    "build": "bun build src/index.ts --outdir dist --target bun --format esm && bun build src/cli/index.ts --outdir dist/cli --target bun --format esm && tsc --emitDeclarationOnly && bun run generate-schema",
+    "build": "bun build src/index.ts --outdir dist --target node --format esm && bun build src/cli/index.ts --outdir dist/cli --target node --format esm && tsc --emitDeclarationOnly && bun run generate-schema",
     "contributors:add": "all-contributors add",
     "contributors:check": "all-contributors check",
     "contributors:generate": "all-contributors generate",
--- a/src/cli/system.ts
+++ b/src/cli/system.ts
@@ -1,4 +1,5 @@
 import { statSync } from 'node:fs';
+import { crossSpawn } from '../utils/compat';
 
 let cachedOpenCodePath: string | null = null;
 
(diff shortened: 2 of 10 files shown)


alvinunreal:
@greptile

alvinunreal:
Thanks @whackur

PR #300, opened 2026-04-14
improve round-based todo hygiene and request boundary handling

dhaern:
Advances https://github.com/alvinunreal/oh-my-opencode-slim/pull/283.

## Summary
- refines todo hygiene from a simple intra-turn nudge into a round-based flow that injects at most one reminder per model round
- keeps hygiene scoped to the current request after `todowrite`, instead of trying to reconcile stale todos at the start of unrelated requests
- hardens request-boundary detection and failure handling so reminders are less noisy and less fragile in real conversations

## Why this is better than #283

PR #283 moved the project in the right direction by adding intra-turn todo hygiene, but in practice it was still too easy to either under-fire or fire in the wrong moments.

This follow-up improves that in a few important ways:

### 1. Better request scoping
The previous direction was still vulnerable to cross-request contamination.

This version keeps the hygiene flow tied to the current request and only activates it after `todowrite` inside that request. That makes it much less likely to:
- drag old `in_progress` state into a new unrelated user request
- nudge the model at the start of simple requests that should not be touching old todos
- bias the model toward reconciling stale state before it has actually done new work

### 2. One reminder per reasoning round
Instead of re-arming reminders based on raw tool count or ad hoc heuristics, this version treats a model round as:

1. the model calls one or more tools
2. tools finish
3. before the next model turn, inject at most one reminder

That makes the behavior much cleaner in exploratory and batch-heavy requests, where the model may call many tools in one round.

The result is:
- less reminder spam
- better fit for parallel tool usage
- more predictable behavior during longer reasoning chains

### 3. Better prioritization of reminder intent
The hygiene state now accumulates reasons during a round and picks the strongest one once:

- `final_active`
- `delegation_resume`
- `general`

This is better than a flat reminder because it distinguishes between:
- a normal "update your todo state" case
- a delegated result just returning via `background_output`
- the stricter "do not leave the last active todo in_progress if you are finishing now" case

That makes the reminder more specific without increasing prompt noise.

### 4. More robust around failures
This version also fixes a couple of real failure modes that could still hurt #283-style behavior:

- a late `getTodoState()` failure no longer clears a reminder that was already armed for the same round
- non-injectable sessions are fully cleared instead of being left internally active
- request-boundary tracking is more stable because it uses message ids when available instead of depending on message array position

That last point is especially important because the previous boundary logic could be fragile if the same message appeared at a different index in a later transform pass.

### 5. Safer fallback behavior
When message metadata is incomplete, this version behaves more defensively:
- it avoids relying only on text equality when a message id exists
- it improves the fallback path when ids are missing
- it clears stale known-orchestrator boundary state when a new user message cannot be mapped to a specific session

That is a better failure mode than silently letting stale request state survive.

## Implementation notes
Main changes are in:
- `src/hooks/todo-continuation/todo-hygiene.ts`
- `src/hooks/todo-continuation/index.ts`

Tests were updated in:
- `src/hooks/todo-continuation/todo-hygiene.test.ts`
- `src/hooks/todo-continuation/index.test.ts`

## Validation
- `bun test src/hooks/todo-continuation/index.test.ts src/hooks/todo-continuation/todo-hygiene.test.ts`
- `bun run build`

### DIFF
--- a/src/hooks/todo-continuation/index.test.ts
+++ b/src/hooks/todo-continuation/index.test.ts
@@ -67,6 +67,23 @@ describe('createTodoContinuationHook', () => {
     return call;
   }
 
+  function userMessages(
+    text: string,
+    sessionID = 'main1',
+    agent?: string,
+    parts?: Array<{ type: string; text?: string }>,
+    id?: string,
+  ) {
+    return {
+      messages: [
+        {
+          info: { id, role: 'user', agent, sessionID },
+          parts: parts ?? [{ type: 'text', text }],
+        },
+      ],
+    };
+  }
+
   describe('tool toggle', () => {
     test('calling auto_continue execute with { enabled: true } sets state', async () => {
       const ctx = createMockContext();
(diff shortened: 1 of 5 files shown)


dhaern:
Addressed the three review threads in the latest push (https://github.com/alvinunreal/oh-my-opencode-slim/commit/6e1369c4b27dfb2ab809ccebe6cdcb8dabea4a89):

- removed the redundant direct re-check in inferSessionID
- deduplicated the external-user-message logic into a shared helper
- added a final_active short-circuit to avoid extra round lookups once the strongest reason is already armed

@greptile-apps please reprocess the updated diff.

PR #302, opened 2026-04-14
fix(agents): use council master model for councillor agent registration

dkovacevic15:
## Summary

- Fix councillor agent always using hardcoded `openai/gpt-5.4-mini` default model instead of the configured council master model
- Councillor was missing from the special-case check in `getModelForAgent()` that `council` and `council-master` already had

## Changes

- Added `'councillor'` to the model resolution condition in `src/agents/index.ts` so the agent definition uses `config.council.master.model` instead of falling through to `DEFAULT_MODELS.councillor`

This caused `ProviderModelNotFoundError` when the hardcoded default provider wasn't available, and used the wrong model even when it was. The council manager correctly passes per-councillor models at prompt time, but OpenCode needs a valid registered model to create the session in the first place.


alvinunreal:
Thanks @dkovacevic15

PR #307, opened 2026-04-15
feat: add disabled_agents config + Observer agent (visual/multimodal analysis)

jwcrystal:
## Why a standalone `@observer` agent?

> **Designer vs Multimodal — the model conflict:** Each agent has one model in OpenCode. A user's strongest reasoning model (for design decisions) may not support vision, while a vision-capable model may not be the best for reasoning. `@multimodal` lets users configure two different models for two different tasks without sacrificing Designer's capability.

### `@multimodal` agent (disabled by default)
- Visual/multimodal analysis for images, screenshots, PDFs, and diagrams
- Disabled by default — enable with `"disabled_agents": []` and configure a vision-capable model
- Processes files in an isolated context, returns only structured text to the orchestrator
- **Inline image interception**: when a user pastes an image, the hook strips image bytes from the orchestrator (preventing API errors on non-vision models), saves to `.opencode/images/`, and nudges delegation to `@multimodal`

### `disabled_agents` config
- Mirrors existing `disabled_mcps` pattern — general mechanism for disabling any agent
- Orchestrator prompt, background tools, and delegation rules all respect the config
- `PROTECTED_AGENTS` (orchestrator, councillor, council-master) cannot be disabled

### Bug fix
- Foreground model resolution: agents not in user's `opencodeConfig.agent` now auto-create entries with the correct model (previously fell back to orchestrator's model)

## Files changed

`src/agents/multimodal.ts` (new), `src/agents/orchestrator.ts`, `src/agents/index.ts`, `src/index.ts`, `src/config/constants.ts`, `src/config/schema.ts`, `src/config/agent-mcps.ts`, `src/background/background-manager.ts`, `src/tools/background.ts`, `README.md`, tests

## Test plan

- [x] `bun test` — 855 pass, 0 fail
- [x] `disabled_agents: []` → `@multimodal` available with correct model
- [x] Omit `disabled_agents` → multimodal not in agent list, image bytes stripped
- [x] Paste image → auto-delegated to `@multimodal`
- [ ] Serve-mode prompt excludes disabled agents (manual)

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,15 +342,15 @@ If any agent fails to respond, check your provider authentication and config fil
   </tr>
 </table>
 
-### 07. Multimodal: The Silent Witness
+### 07. Observer: The Silent Witness
 
 > [!NOTE]
-> **Why a separate agent?** Not all models support vision. Your strongest coding model (e.g. for design decisions) may not be able to read images, while a vision-capable model may not be the best for reasoning. Multimodal solves this by having its **own model** — configure a vision-capable model for it while keeping Designer on your strongest reasoning model. Disabled by default; enable via `disabled_agents: []` in config.
+> **Why a separate agent?** Not all models support vision. Your strongest coding model (e.g. for design decisions) may not be able to read images, while a vision-capable model may not be the best for reasoning. Observer solves this by having its **own model** — configure a vision-capable model for it while keeping Designer on your strongest reasoning model. Disabled by default; enable via `disabled_agents: []` in config.
 
 <table>
   <tr>
     <td width="240" valign="top">
-      <b>Multimodal</b><br>
+      <b>Observer</b><br>
       <i>Visual & binary analysis</i>
     </td>
     <td>
@@ -365,7 +365,7 @@ If any agent fails to respond, check your provider authentication and config fil
   </tr>
   <tr>
     <td colspan="2">
-      <b>Prompt:</b> <a href="src/agents/multimodal.ts"><code>multimodal.ts</code></a>
+      <b>Prompt:</b> <a href="src/agents/observer.ts"><code>observer.ts</code></a>
     </td>
   </tr>
   <tr>
--- a/README.md
+++ b/README.md
@@ -342,6 +342,39 @@ If any agent fails to respond, check your provider authentication and config fil
   </tr>
 </table>
 
+### 07. Multimodal: The Silent Witness
+
+> [!NOTE]
+> **Why a separate agent?** Not all models support vision. Your strongest coding model (e.g. for design decisions) may not be able to read images, while a vision-capable model may not be the best for reasoning. Multimodal solves this by having its **own model** — configure a vision-capable model for it while keeping Designer on your strongest reasoning model. Disabled by default; enable via `disabled_agents: []` in config.
+
+<table>
+  <tr>
+    <td width="240" valign="top">
+      <b>Multimodal</b><br>
+      <i>Visual & binary analysis</i>
+    </td>
+    <td>
+
+**Read-only visual analysis** — interprets images, screenshots, PDFs, and diagrams. Returns structured observations to the orchestrator without loading raw file bytes into the main context window.
+
+- Images, screenshots, diagrams → `read` tool (native image support)
+- PDFs and binary documents → `read` tool (text + structure extraction)
+- **Disabled by default** — enable with `"disabled_agents": []` and configure a vision-capable model
+
+    </td>
+  </tr>
+  <tr>
+    <td colspan="2">
+      <b>Prompt:</b> <a href="src/agents/multimodal.ts"><code>multimodal.ts</code></a>
+    </td>
+  </tr>
+  <tr>
+    <td colspan="2">
+      <b>Default Model:</b> <code>openai/gpt-5.4-mini</code> — <i>configure a vision-capable model to enable</i>
+    </td>
+  </tr>
+</table>
+
 ---
 
 ## 📚 Documentation
--- a/oh-my-opencode-slim.schema.json
+++ b/oh-my-opencode-slim.schema.json
@@ -340,7 +340,7 @@
       }
     },
     "disabled_agents": {
-      "description": "Agent names to disable completely. Disabled agents are not instantiated and cannot be delegated to. Orchestrator and council internal agents (councillor, council-master) cannot be disabled. By default, 'multimodal' is disabled. Remove it from this list and configure a vision-capable model to enable.",
+      "description": "Agent names to disable completely. Disabled agents are not instantiated and cannot be delegated to. Orchestrator and council internal agents (councillor, council-master) cannot be disabled. By default, 'observer' is disabled. Remove it from this list and configure a vision-capable model to enable.",
       "type": "array",
       "items": {
         "type": "string"
--- a/oh-my-opencode-slim.schema.json
+++ b/oh-my-opencode-slim.schema.json
@@ -340,7 +340,7 @@
       }
     },
     "disabled_agents": {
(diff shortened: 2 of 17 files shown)


alvinunreal:
I'd not go with a new optional agent, maybe simple we can add orchestrator_multimodal flag or something which sets special prompt for orchestrator to use designer for multimodal if it can't;

jwcrystal:
Thanks for the review! You're right that for a slim fork, adding a new agent increases complexity. Let me explain the original intent and propose a compromise.

## Why I added looker

The core use case: **not all models support multimodal input**. When a text-only model receives a screenshot/PDF, it either garbles the output or wastes context on binary data. A dedicated read-only visual agent with a separately configurable vision model solves this cleanly.

But I hear you — `looker` is ultimately a single-purpose agent, and `designer` already has visual capabilities in its prompt.

## What I'd like to keep: `disabled_agents`

The `disabled_agents` mechanism is independent of `looker`. It provides:

1. **Clean agent list** — users running text-only models can disable agents they'll never use
2. **Full coverage filtering** — agent descriptions, validation routing, parallel examples, background tool descriptions, and delegation validation all respect the config
3. **Extensibility** — future experimental agents (e.g., a code-search agent using embeddings) can be disabled by default without code changes

This is the same pattern as the existing `disabled_mcps` — a user-facing toggle, not a per-agent special case.

## Proposed compromise

1. **Keep `disabled_agents`** — it's a general mechanism, not just for looker
2. **Remove `looker` agent** — per your preference
3. **Remove `DEFAULT_DISABLED_AGENTS`** — since there's no agent to default-disable
4. **Add an `orchestrator_multimodal` flag** — when `false`, the orchestrator prompt instructs delegating visual analysis to `@designer` instead of reading images directly

This gives us:
- The `disabled_agents` infrastructure for future use
- No new agent
- Visual analysis routed through `designer` when the orchestrator model can't handle it

Would this direction work for you? I can refactor the PR accordingly.

jwcrystal:
Follow-up on the model separation issue — this is the key reason I designed looker as a separate agent rather than routing through designer:

## The model conflict

Designer and visual analysis have **different model requirements**:

| Task | Model need |
|------|-----------|
| UI/UX design decisions | Strong reasoning (code generation, aesthetic judgment) |
| Visual analysis | Vision capability (image/PDF reading, OCR) |

Designer has **one model** in OpenCode's architecture — it's set at agent config level, fixed for all invocations. There's no per-task model switching.

So if a user's strongest model (e.g. `glm-5.1`) doesn't support vision, they face a dilemma:
- Configure designer with `glm-5.1` → good design work, can't read images
- Configure designer with `gemma4:31b` (vision) → can read images, weaker design reasoning

## Why looker solves this

Looker lets the user configure **two different models** for two different tasks:
```
designer: { model: "glm-5.1" }     // strong reasoning for design
looker:   { model: "gemma4:31b" }  // vision-capable for image analysis
```

## Why it's not heavy

- Looker is a single-purpose agent: read file → return structured text
- `disabled_agents: ["looker"]` is the default — users who don't need it never see it in their agent list or orchestrator prompt
- The `disabled_agents` mechanism itself is valuable independent of looker (same pattern as `disabled_mcps`, useful for future experimental agents)

That said, if you'd prefer a simpler approach, I'm happy to go with just `disabled_agents` without looker and add visual routing later when there's clearer demand. Let me know what you think.

alvinunreal:
ok I see yes, than designer seems like sacrificed agent to get vision;

Than we can go your way, maybe call "multimodal" and in readme we explain why it exists as a standalone agent

ReqX:
Really like the idea, used a simple hook to a outside "opencode run" while testing glm-5v-turbo, but this is a splendid approach.

Naming wise, I am not convinced @alvinunreal :) It feels it breaks the naming convention. Every other agent is a single descriptive role noun: explorer, librarian, oracle, designer, fixer. Maybe **observer**? Fits the pattern perfectly (person-who-observes), describes exactly what it does, and mirrors the "read-only analysis" nature of the agent.

On the code:

The `disabled_agents` mechanism is clean. `PROTECTED_AGENTS`, the config schema, and the reuse of `getDisabledAgents()` across all call sites after the last fix look good.

The image interception hook solves a real problem and I like the approach of saving to `.opencode/images/` within the workspace. A few things I'd flag that remain after the latest fixes:

1. **Sync file I/O on every `messages.transform` call.** The hook fires before every LLM API call and runs `mkdirSync`, `readdirSync`, `statSync`, `unlinkSync`, `writeFileSync`. The cleanup loop in particular iterates all files in `.opencode/images/` on every single invocation. In practice this is bounded (small directory, few images) so not a blocker, but worth considering debouncing the cleanup or caching the mkdir/gitignore check.

2. **Full message history re-processed every call.** `for (const msg of typedOutput.messages)` iterates all historical user messages on every transform, re-stripping and re-saving images that were already handled in previous turns. The SHA-based naming means it just overwrites the same file with no accumulation bug, but it's redundant I/O that grows with conversation length.

3. **~100 lines of hook logic inline in `src/index.ts`.** The image interception logic (isImagePart, decodeDataUrl, extFromMime, save, cleanup, strip) might benefit from its own module like `src/hooks/image-hook.ts`. Might be my flavour, can be done if it grows as well and I am far from a "perfect" coder;)

4. **Silent empty `catch {}` blocks.** The mkdir/gitignore block and cleanup loop swallow errors with bare `catch {}`. The image save got proper logging, but the others should too. If `.opencode/images/` can't be created, debugging why images silently don't get saved would be painful.

None of these are merge-blockers in my view, the core mechanism is honestly solid.

@greptile-apps[bot] can you take a look at this specifically, maybe I am seeing it wrong.

Really love it and big contribution, I was missing this personally <3

alvinunreal:
observer works yes, sorry for extra rename

jwcrystal:
@ReqX  Thanks for the detailed review 🙏 Addressing each point:

`Observer` rename — Agreed. observer fits the naming convention perfectly. Will rename. (Just want an easy-to-understanding noun at first, HAHA)

1. Sync file I/O on every transform 

Fair point. The cleanup is the main concern — The cleanup loop already deletes files older than 1 hour, but it runs `readdirSync` + `statSync` on every messages.transform call regardless. Most invocations find nothing to delete. Will add a timestamp-based debounce (10 min interval) so it only runs periodically instead of every API call. — the 1-hour age threshold stays unchanged.

2. Full message history re-processed every call 

This is inherent to the messages.transform API — the hook receives the full conversation history with no way to distinguish new vs old messages. The SHA-based naming means the same image overwrites the same file, so no accumulation. A hash-based cache could skip the write, but you'd still need to decode + hash to check — marginal savings. Planning to keep as-is.

3. ~100 lines of hook logic inline in `src/index.ts` 

Agreed. Will extract to `src/hooks/image-hook.ts` — was always the plan, good time to do it.

4. Silent empty catch {} blocks 

This issue was worked based on MODEL before, then disabling log mechanism: **mkdir fails silently → writeFileSync fails → that catch has a log → the model sees the error → report back to the user**.

People might overlook this error detail when they encounter it from this agent, as the **only difference is in the quality of the error.**  e.g.,

- No log: "failed to save image: ENOENT: no such file or directory" — the user does not know why the directory does not exist
- log: "failed to create .opencode/images/: EACCES: permission denied" → then the write fails - the root cause is clear

`mkdirSync/gitignore` block: will add log(), considering the error quality what people want to know — it's the diagnostic chain that tells you why image saving silently failed (e.g. EACCES on mkdir vs ENOENT on write). 
`Cleanup` block: will keep silent — failing to delete old files doesn't affect functionality, logging would just add noise.

@alvinunreal Will rename multimodal → observer. Thanks for confirming!

PR #309, opened 2026-04-16
feat: add per-session log rotation with auto-cleanup

jwcrystal:
## Summary

- Replace single flat log file with per-session log files (`oh-my-opencode-slim.YYYYMMDDTHHMMSS.log`)
- Fix Windows-incompatible path (`process.env.HOME` → `os.homedir()`)
- Auto-cleanup log files older than 7 days on session init
- Add `OPENCODE_LOG_DIR` env var for test isolation

## Changes

| File | Change |
|------|--------|
| `src/utils/logger.ts` | `initLogger(sessionId)`, `resetLogger()`, `getLogDir()`, cleanup logic, `os.homedir()` |
| `src/utils/index.ts` | Barrel export new functions |
| `src/index.ts` | Call `initLogger()` at plugin startup |
| `src/utils/logger.test.ts` | 15 tests with temp dir isolation |

## Design decisions

- **`log()` signature unchanged** — all 163+ call sites work without modification
- **Uninitialized = silent no-op** — `log()` before `initLogger()` returns immediately instead of writing to a stale file. Removes the module-level `mkdirSync` side effect at import time.
- **No log levels** — YAGNI for now, can be added later without breaking the API
- **No config schema** — retention (7 days) and path are hardcoded defaults

## Test plan

- [x] 15 new tests covering: init, per-session files, cleanup (old/recent/mixed-age), no-op before init, circular data, env var override
- [x] All tests isolated via `OPENCODE_LOG_DIR` temp directory
- [x] `bun run typecheck` passes
- [x] `bun test` — 599/601 pass (2 pre-existing failures in `auto-update-checker/cache`)
- [ ] Manual test: launch OpenCode, verify `~/.local/share/opencode/oh-my-opencode-slim.*.log` created per session
- [ ] Manual test: verify old files cleaned up after 7 days

Related: #307

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -34,10 +34,7 @@ import {
 import { initLogger, log } from './utils/logger';
 
 const OhMyOpenCodeLite: Plugin = async (ctx) => {
-  const sessionId = new Date()
-    .toISOString()
-    .replace(/[-:]/g, '')
-    .slice(0, 15);
+  const sessionId = new Date().toISOString().replace(/[-:]/g, '').slice(0, 15);
   initLogger(sessionId);
   const config = loadPluginConfig(ctx.directory);
   const agentDefs = createAgents(config);
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,9 +31,14 @@ import {
   lsp_rename,
   setUserLspConfig,
 } from './tools';
-import { log } from './utils/logger';
+import { initLogger, log } from './utils/logger';
 
 const OhMyOpenCodeLite: Plugin = async (ctx) => {
+  const sessionId = new Date()
+    .toISOString()
+    .replace(/[-:]/g, '')
+    .slice(0, 15);
+  initLogger(sessionId);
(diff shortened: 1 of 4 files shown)


jwcrystal:
@greptile-apps thanks for the review. Check again please

jwcrystal:
@greptile-apps Valid point on serve-mode session isolation. The current timestamp-based init covers the normal single-session case where one OpenCode process = one chat. For serve mode (one process, multiple chat sessions), logs would indeed be mixed.

This is a known limitation scoped out of this initial PR. The follow-up plan is to switch `initLogger(sessionID)` inside the `chat.message` hook when a new OpenCode `sessionID` is detected — but that requires careful handling of concurrent writes and session lifecycle tracking, which deserves its own PR.

Current naming: `oh-my-opencode-slim.20260416T143052.log` (timestamp at plugin startup).

alvinunreal:
Thanks @jwcrystal

PR #322, opened 2026-04-16
fix: persist background task results to disk to survive plugin reinitialization

tom-dyar:
## Problem

`background_output(task_id)` always returns **"Task not found"** when called after context compaction fires in the parent session.

**Root cause**: `BackgroundTaskManager` stores tasks in a plain `new Map()`. When context compaction occurs, the plugin reinitializes and creates a fresh empty Map — all completed task results are lost. `getResult()` then returns `null` → "Task not found".

This is 100% reproducible:
1. Launch a background task
2. Wait long enough for context compaction to fire (or run `/compact` manually)
3. Receive completion notification
4. Call `background_output(task_id)` → "Task not found"

## Fix

Two small changes to `background-manager.ts`:

**`completeTask()`** — after marking the task complete, writes a JSON file to `~/.local/share/opencode/bg-tasks/<task-id>.json` via `persistTask()`.

**`getResult()`** — if the task is not in the in-memory Map (manager was recreated), falls back to `loadPersistedTask()` which reads from disk, then re-registers in memory for fast subsequent lookups.

Both I/O operations are wrapped in `try/catch` — file system errors log a warning and degrade gracefully.

## Implementation notes

- Uses `getLogDir()` from `utils/logger` so state is co-located under the same base directory (`~/.local/share/opencode/`)
- `mkdirSync({ recursive: true })` is idempotent — no pre-check needed
- No new dependencies

## Tests

Three new tests added to `background-manager.test.ts` under `getResult > disk persistence`:
- Completed task is retrievable by a **new** manager instance after reinitialization
- Failed task is also recoverable
- Returns `null` for a task that never completed (no disk file)

All 49 tests pass.


alvinunreal:
@copilot resolve the merge conflicts in this pull request

alvinunreal:
Thanks @tom-dyar  - conflict with main otherwise good

Issue #290, opened 2026-04-13
[Bug] Copilot-backed subagents like oracle can still trigger ask-question prompts instead of returning control to orchestrator


tlerbao:
## Summary
When using `oh-my-opencode-slim` with GitHub Copilot models, `subagent` replies such as `oracle` and sometimes `explorer` can still trigger `ask question` interactive follow-up prompts.

I verified this **without any custom local plugin loaded**. After removing my own `copilot-interactive-question` plugin, the behavior still reproduced.

## Why this seems wrong
My expectation is:
- `primary/main thread + github-copilot` => interactive follow-up can happen
- `subagent/background task + github-copilot` => should finish normally and hand results back to the orchestrator/main thread

Current behavior makes background/subagent execution inconsistent. In practice, the main thread sometimes shows:

- `-> Asked 1 question`
- and the subagent path stops for interactive confirmation instead of simply returning control to the orchestrator

This feels different from the normal background-task flow, where subagents investigate and then hand results back to the main process.

## Repro
Using an `oh-my-opencode-slim` preset where subagents are configured like this:
- `explorer` -> `github-copilot/claude-sonnet-4.6`
- `oracle` -> `github-copilot/claude-opus-4.6`
- `orchestrator` -> non-Copilot primary model

1. Start a normal orchestrator session.
2. Let orchestrator delegate to Copilot-backed subagents.
3. Observe the result.

### Observed behavior
- Parallel `explorer` subagents may complete without asking.
- A later `oracle` subagent may still trigger `ask question`.
- So the behavior is not only undesirable, but also inconsistent.

## Expected behavior
For Copilot-backed subagents/background tasks:
- no interactive `ask question` follow-up should be triggered at subagent completion
- the subagent should complete and hand its result back to the orchestrator/main thread
- any interactive confirmation should happen only in the primary/main-thread conversation if that is the intended policy

## Question
Is this the intended behavior in `oh-my-opencode-slim`?

If yes, can you explain the intended interaction model for Copilot-backed subagents such as `oracle`?

If not, should subagent/background-task replies explicitly bypass `ask question` so they behave like normal background investigation and return control to the orchestrator?

Taudris:
There's no way this is intended behavior. It's incredibly annoying.

[...]

PR #335, opened 2026-04-19
feat(background): non-blocking question relay for background subagents

ReqX:
Ref #290

## Summary

Background subagents (oracle, explorer, fixer, etc.) can call OpenCode's built-in `question` tool, which blocks the session waiting for user input that never comes. This causes stuck subagents and confusing UX.

This PR:
1. **Denies the `question` tool** for all background task sessions via `calculateToolPermissions()`
2. **Adds `ask_orchestrator`** - a non-blocking question relay tool that records questions on the task and returns immediately
3. **Surfaces relayed questions** in `background_output` so the orchestrator can review subagent assumptions

## How it works

```
Subagent hits ambiguity
  → calls ask_orchestrator("Should I use X or Y?")
  → tool records question on task.questions[]
  → returns "Continue with [ASSUMED: ...] markers"
  → subagent picks best guess, marks [ASSUMED: using X], keeps working

Orchestrator reads background_output
  → sees result + "Questions relayed from subagent" section
  → evaluates assumptions, asks user only if needed
```

## Commits

| # | Commit | Scope |
|---|--------|-------|
| 1 | `feat(background): non-blocking question relay` | Core code + tests (~135 prod lines) |
| 2 | `feat(agents): add ask_orchestrator guidance` | Subagent + orchestrator prompts |
| 3 | `docs(tools): add ask_orchestrator documentation` | docs/tools.md |
| 4 | `chore: biome formatting` | Import order, line wrapping |

## Testing

- 1031 tests pass (17 new)
- Types clean
- Backward compatible (persisted tasks load with `questions: []` fallback)

## Design decisions

- **Non-blocking by design.** The tool explicitly tells the model it will NOT receive an answer. Trades correctness for liveness — subagent guesses instead of blocking. Mitigated by `[ASSUMED: ...]` markers so orchestrator can evaluate.
- **Typed return union.** `addQuestion` returns `'recorded' | 'not-found' | 'terminal' | 'cap-reached'`, each case gets an accurate message for the subagent.
- **`question: false` dual enforcement.** OpenCode filters via both `user.tools` check and session permission conversion. Either path independently blocks the tool.
- **Questions on BackgroundTask, not a separate store.** Simple array field, persisted with the task, cleaned up with the task.
- **Cap at 50 questions per task.** Prevents unbounded accumulation from runaway agents.

## Risks

- Model compliance with `[ASSUMED: ...]` markers is soft — relies on prompt guidance in tool description and agent prompts

## Optional future work / options

- **Re-delegation**: orchestrator asks user, re-launches task with answer. Only if data shows meaningful questions.
- **Structured `[ASSUMED: ...]` extraction** — post-process subagent output to flag assumed sections for orchestrator review.
- **Config toggle** ,  `background.question_relay: false` to opt out (~25 lines, guards in 3 places). Not included now (YAGNI - add when needed).

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -15,8 +15,11 @@ Launch agents asynchronously and collect results later. This is how the Orchestr
 | Tool | Description |
 |------|-------------|
 | `background_task` | Launch an agent in a new session. `sync=true` blocks until complete; `sync=false` returns a task ID immediately |
-| `background_output` | Fetch the result of a background task by ID |
+| `background_output` | Fetch the result of a background task by ID. Surfaces questions relayed from subagents |
 | `background_cancel` | Abort a running background task |
+| `ask_orchestrator` | Non-blocking question relay for background subagents. Records a question for orchestrator review without waiting for an answer. Subagents state their assumption via `[ASSUMED: ...]` markers and continue working |
+
+Background subagents have the built-in `question` tool disabled (`question: false`) to prevent sessions from blocking on user input that never arrives. Instead, subagents use `ask_orchestrator` to relay questions to the orchestrator, who evaluates them when retrieving results via `background_output`.
 
 Background tasks integrate with [Multiplexer Integration](multiplexer-integration.md) — when multiplexer support is enabled, each background task spawns a pane so you can watch it live.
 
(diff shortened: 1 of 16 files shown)


ReqX:
Thanks @greptile-apps[bot] — both P2s fixed in 1730ff3:

1. **Stale test name** → renamed to "returns terminal for completed task"
2. **Temp-dir isolation** → wrapped addQuestion persistence test in its own describe block with `OPENCODE_LOG_DIR` + mkdtemp/cleanup, matching the existing disk persistence suite pattern.

Ready for re-review.

PR #339, opened 2026-04-19
fix: ignore non-string plugin entries during install detection

Daltonganger:
## Summary
- guard plugin entry handling in the installer config flow so mixed `plugin` arrays no longer crash on `startsWith`
- normalize `config.plugin` to string entries before install detection and plugin rewrite logic runs
- keeps the change intentionally minimal and scoped to the update/install path the bug reproduces in

## Verification
- `bun run typecheck`
- `bun test src/cli/config-io.test.ts`
- manual repro with a temp OpenCode config containing `legacy-plugin`, an object plugin entry, and `oh-my-opencode-slim@0.7.0`

### DIFF
--- a/src/cli/config-io.test.ts
+++ b/src/cli/config-io.test.ts
@@ -203,6 +203,31 @@ describe('config-io', () => {
     expect(saved.plugin).toEqual(['other', packageRoot]);
   });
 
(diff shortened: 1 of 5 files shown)


Daltonganger:
Changed what greptile found, so please do @greptile again!

PR #343, opened 2026-04-19
feat: add init-time health check and defensive logging

### LINKED ISSUE #310: V0.9.13 not working
i currently use previous version very good but it automatic update to v0.9.13 and it go to break OmO slim the orchestrator and other not work just build and plan agent now

ReqX:
## Summary

Adds defensive init-time validation to prevent the silent "plugin loaded but nothing works" failure mode reported in #310.

## Problem

When the plugin fails during initialization (e.g. jsdom import failure due to npm dependency resolution issues in opencode v1.4.0+, or the opencode v1.4.8-v1.4.9 deadlock), opencode shows the plugin as "loaded" in `/status` but zero agents, tools, or MCPs are actually registered. Users see "not working" with no actionable error message.

## Changes

- **`appLog()` helper** - defensive wrapper around `ctx.client.app.log()` with stderr fallback. Catches deadlocks on opencode v1.4.8-v1.4.9 where `client.app.log()` during init triggers a middleware cycle.
- **Init try/catch** - wraps the entire plugin initialization body. On failure: logs to file, attempts visible `appLog()` error, then re-throws so opencode knows it failed.
- **Post-init health check** - counts registered agents/tools/MCPs against minimum thresholds (`minAgents: 5, minTools: 5, minMcps: 1`). If suspiciously low, emits a warning with actionable guidance pointing to #310.
- **`probeJSDOM()`** - fire-and-forget async check that dynamically imports jsdom and instantiates a trivial DOM. Catches dependency resolution issues early instead of waiting for the first `webfetch` call. Non-blocking; does not delay plugin init.

## Context

Issue #310 has multiple root causes:
1. ~~jsdom hardcoded path~~ (fixed in v0.9.14 via #313)
2. ~~opencode v1.4.8-v1.4.9 deadlock~~ (fixed in opencode v1.4.10)
3. **opencode v1.4.0 npm dependency switch** (still open - [anomalyco/opencode#21625](https://github.com/anomalyco/opencode/issues/21625))

This PR doesn't fix cause #3 (that's upstream), but it makes the failure visible and actionable instead of silent.

## Testing

- `bun run typecheck` - clean
- `bun run check` - no new errors
- `bun test` - 1032/1032 pass

Closes #310 (partial - jsdom path fix already shipped; this adds diagnostics for remaining silent failures)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -36,156 +36,300 @@ import {
 import { resolveRuntimeAgentName, rewriteDisplayNameMentions } from './utils';
 import { initLogger, log } from './utils/logger';
 
+/**
+ * Best-effort log to opencode's app logger.
+ * Wrapped in try/catch to avoid deadlocking on opencode v1.4.8–v1.4.9
+ * where client.app.log() during init triggers a middleware cycle.
+ */
+async function appLog(
+  ctx: Parameters<Plugin>[0],
+  level: 'error' | 'warn' | 'info',
+  message: string,
+): Promise<void> {
+  try {
+    await ctx.client.app.log({
+      body: { service: 'oh-my-opencode-slim', level, message },
+    });
+  } catch {
+    // client.app.log may deadlock or be unavailable; stderr is the
+    // fallback
+    const prefix =
+      level === 'error' ? 'ERROR' : level === 'warn' ? 'WARN' : 'INFO';
+    console.error(`[oh-my-opencode-slim] ${prefix}: ${message}`);
+  }
+}
+
+/** Minimum expected registrations for a healthy plugin load. */
+const HEALTH_CHECK = {
+  minAgents: 5,
+  minTools: 5,
+  minMcps: 1,
+} as const;
+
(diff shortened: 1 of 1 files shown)


ReqX:
**@greptile-apps[bot] ready for re-review.** All three findings addressed:

1. **`toolCount` now includes `todoContinuationHook.tool`** - moved the count calculation to after `todoContinuationHook` is initialized and added `Object.keys(todoContinuationHook.tool).length` to the sum. Logged count is now accurate.

2. **MCP threshold adapts to user config** - when `disabled_mcps` is configured, the MCP threshold drops to 0 so deliberate MCP disabling doesn't trigger a false-positive warning. Users who disable all MCPs won't see the "suspiciously low" message.

3. **Added `.catch(() => {})` to fire-and-forget `appLog`** - the `probeJSDOM().then()` callback now catches the dangling promise to prevent unhandled rejection warnings if the context is torn down before the probe resolves.

Issue #345, opened 2026-04-20
The observer agent cannot obtain images.


zuuky:
The conditional statement in `image-hook.ts -> processImageAttachments -> cleanupOldImages` has an issue. The directory storing images doesn't have `lastCleanupByDir` stored after its initial creation, causing the deletion logic to be executed immediately after the directory is created. This results in the observer not obtaining the correct path when retrieving images.

PR #346, opened 2026-04-20
fix: #345

zuuky:
## Summary

Fixes #123

## Changes

The directory for storing images in session ID is given a default timeout upon initial creation to avoid the 10-minute timer being ineffective. The `if (now - lastCleanup < CLEANUP_INTERVAL) return;` statement will then directly trigger the deletion process after the first creation.


alvinunreal:
Thanks @zuuky

PR #352, opened 2026-04-20
rename cartography to codemap

alvinunreal:
Explorer can't edit files.

Renaming to codemap, rewriting in node since python3 not available in windows and using fixer


alvinunreal:
@greptile

alvinunreal:
@greptileai

PR #365, opened 2026-04-21
feat(agents): support custom subagent routing prompts

alvinunreal:
Allow custom agents to define prompt and orchestratorPrompt blocks while tightening validation, display-name handling, and runtime resolution. Also limit Bun test discovery to the plugin src tree so root test runs don’t recurse into the nested opencode checkout.


alvinunreal:
@greptileai

PR #381, opened 2026-04-22
fix: mutate system array in-place for Qwen compatibility

### LINKED ISSUE #315: Qwen3.5 series models fail with 'System message must be at the beginning' error in v0.9.14
## Issue

Image

After updating from v0.9.2 to v0.9.14, Qwen3.5 series models (e.g., Qwen3.6 plus, Qwen3.5 27B) throw the error:

## Reproduction Steps

1. Use oh-my-opencode-slim v0.9.14
2. Configure a Qwen3.5 series model 
3. Run opencode

## Expected Behavior

System prompt should be placed at the beginning of the message array.

## Actual Behavior

System prompt appears to be placed incorrectly, causing the error.

## Workaround

Removing the plugin resolves the issue immediately.

## Suggestion

Please check if the system prompt is being inserted at the correct position in the message array for Qwen3.5 models.

ReqX:
Fixes #315 - for real this time I hope.

## Problem

PR #336 attempted to fix multiple system messages breaking Qwen models (via VLLM/DashScope) by collapsing `output.system` to a single element. However, the fix was a no-op:

```ts
// PR #336 — creates a new array OpenCode never sees
output.system = joined ? [joined] : [];
```

OpenCode (`session/llm.ts:114-117`) passes `{ system }` to the hook, where `system` is a local `const` referencing the original array. After the hook returns, OpenCode reads from that local variable, not from `output.system`. The reassignment went to a detached object.

## Fix

Mutate the array in-place so OpenCode's reference sees the change:

```ts
const joined = output.system.join('\n\n');
output.system.length = 0;
if (joined) {
  output.system.push(joined);
}
```

## Tradeoff: prompt caching

OpenCode preserves a 2-part `[header, rest]` system message structure for prompt caching (Anthropic, OpenRouter, Bedrock, etc.). Collapsing to 1 message means the entire system prompt becomes the cache key; any injected reminder (todo hygiene, file tool nudge) invalidates the cache for the whole block.

Estimated impact: ~$0.17 over a 20-turn conversation (based on ~3K token header at Sonnet pricing). Not zero, but negligible compared to "Qwen users can't use the plugin at all."

If this becomes a real concern, the correct fix belongs in OpenCode itself; per-provider system message collapsing is not a plugin's job.


ReqX:
Test added in 3c484f2. Validates the in-place mutation contract with 5 cases including a regression guard that proves the old reassignment approach was broken. @greptile-apps[bot] mind taking another look?

ReqX:
All Greptile feedback addressed:

- **P2 (regression test)**: Added `src/utils/system-collapse.test.ts` with 5 tests covering multi-element, single-element, empty array, and a regression guard proving reassignment doesn't work.
- **P1 (tests exercised local copy)**: Extracted `collapseSystemInPlace` to `src/utils/system-collapse.ts`. Both the hook and tests import the same function. If the implementation regresses, the tests catch it.

mumuyizhou:
still multiple `system` :
image

which resulted in Qwen( Qwen-V3.6-35B-A3B & Qwen-V3.5-397B-A17B
) returns error. 
version: 
OMO-Slim 1.0.1 (dev)
Running in local development mode.

ReqX:
> still multiple `system` :
> image
> 
> which resulted in Qwen( Qwen-V3.6-35B-A3B & Qwen-V3.5-397B-A17B
> ) returns error. 
> version: 
> OMO-Slim 1.0.1 (dev)
> Running in local development mode.



Incredible:) 

Would you be able to send your agent to debugg this? If it can find which hook or part we miss, I can gladly take it from there. 
It's hard without a possibility to test honestly.

mumuyizhou:
> > still multiple `system` :
> > image
> > which resulted in Qwen( Qwen-V3.6-35B-A3B & Qwen-V3.5-397B-A17B
> > ) returns error.
> > version:
> > OMO-Slim 1.0.1 (dev)
> > Running in local development mode.
> 
> Incredible:)
> 
> Would you be able to send your agent to debugg this? If it can find which hook or part we miss, I can gladly take it from there. It's hard without a possibility to test honestly.

Sorry , I checked release note, v1.0.1 does not include your commit yet.  Qwen works fine with OMOS that I compiled with latest branch main in local ✌️✌️✌️ . Many thanks to you.

ReqX:
> still multiple `system` :
> image
> 
> which resulted in Qwen( Qwen-V3.6-35B-A3B & Qwen-V3.5-397B-A17B
> ) returns error. 
> version: 
> OMO-Slim 1.0.1 (dev)
> Running in local development mode.

Just super happy, this one was starting to haunt me! Thx for the check and extra effort!

PR #390, opened 2026-04-23
feat: add resumable specialist session manager

alvinunreal:
Related https://github.com/alvinunreal/oh-my-opencode-slim/issues/387

Issue #387, opened 2026-04-23
[Feature] Utilize session resumption (task_id) to preserve subagent context across multiple delegations


alvinreal:
## Summary

OpenCode's `task` tool supports **session resumption via `task_id`**, allowing the orchestrator to call the same subagent session again with full context preservation. Currently, oh-my-opencode-slim spawns fresh subagent sessions for each delegation, losing valuable context and conversation history.

## The Opportunity

When a subagent finishes and reports back, the orchestrator currently treats subsequent related work as a brand-new task. Instead, we should leverage `task_id` to resume the same session when:
- The follow-up work is directly related to the previous task
- Context from the previous session would improve quality/speed
- The same specialist agent is being called again

## How It Works in OpenCode

From the `task` tool implementation:
```typescript
const taskID = params.task_id
const session = taskID
  ? yield* sessions.get(SessionID.make(taskID)).pipe(Effect.catchCause(() => Effect.succeed(undefined)))
  : undefined
const nextSession = session ?? (yield* sessions.create({...}))
```

When `task_id` is provided, the tool looks up the existing session. If found, it continues that session with all prior context (files read, conversation history, todos) intact.

## Benefits

1. **Token efficiency** - Avoid re-sending full context summaries
2. **Better continuity** - Subagent remembers what it already explored/did
3. **Faster iterations** - No warm-up period for the agent
4. **Deeper work** - Enables multi-step collaboration with the same specialist

### Where to Store task_id

The orchestrator needs to track active subagent sessions. Options:
- **Memory/state**: Track in orchestrator's working memory (simplest)
- **Structured output**: Have subagents return `task_id` in a structured format
- **Session registry**: Maintain a lightweight map of agent_type → active_session_id

### Decision Logic

When should the orchestrator resume vs. spawn fresh?

| Scenario | Action |
|----------|--------|
| Follow-up to same task | Resume with `task_id` |
| New unrelated task | Fresh session |
| Same agent, different task | Fresh session (or user-configurable) |
| Explicit user request to continue | Resume |

### Prompt Updates Needed

1. **Orchestrator prompt**: Add guidance on when/how to use `task_id` for resumption
2. **Agent prompts**: Ensure agents return `task_id` clearly when they expect follow-up
3. **Documentation**: Explain the feature to users

## Proposed Design

```typescript
// In orchestrator delegation logic
const delegate = async (task, context) => {
  const existingSession = findRecentSessionForAgent(task.agentType, context)
  
[...]

PR #392, opened 2026-04-23
feat: surface interrupted subagent sessions for recovery

SweetSophia:
## Summary
- Reworks the original session recovery idea on top of the current `task-session-manager` instead of adding a separate tracker/hook.
- Adds a recoverable interruption detector for task outputs with parseable `task_id` values, including empty `<task_result>`, provider errors, 429/rate-limit, timeout, quota and resource-exhaustion signals.
- Appends an informational `[task partial state available]` note to interrupted task tool output so the orchestrator can decide whether to resume the existing subagent session.
- Keeps the existing `<resumable_sessions>` prompt injection and alias/session tracking as the single source of truth.

## What changed from the old PR shape
- Removed the duplicate `TaskSessionTracker` approach.
- Removed the separate `delegate-task-resume` hook.
- Reused the existing `task-session-manager` and `parseTaskIdFromTaskOutput()` path.
- Left `AGENTS.md` and unrelated formatting/test changes out of this PR.

## Safety / compatibility
- Recovery notes are only added when the task output includes a parseable `task_id`.
- Argument/validation errors and missing-session errors do not get recovery notes.
- Existing recovery markers are not duplicated.
- The note is informational only; it does not force auto-resume.

## Validation
- `bun test`
- `bun run typecheck`
- `bun run lint`
- `git diff --check`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,258 +4,123 @@ This document provides guidelines for AI agents operating in this repository.
 
 ## Project Overview
 
-**oh-my-opencode-slim** - A lightweight agent orchestration plugin for OpenCode, a slimmed-down fork of oh-my-opencode. Built with TypeScript, Bun, and Biome.
+**oh-my-opencode-slim** — An OpenCode plugin that adds specialist-agent orchestration. Built with TypeScript, Bun, and Biome. Not a standalone app; it registers agents, tools, hooks, and MCPs into the OpenCode runtime.
 
 ## Commands
 
 | Command | Description |
 |---------|-------------|
-| `bun run build` | Build TypeScript to `dist/` (both index.ts and cli/index.ts) |
-| `bun run typecheck` | Run TypeScript type checking without emitting |
-| `bun test` | Run all tests with Bun |
-| `bun run lint` | Run Biome linter on entire codebase |
-| `bun run format` | Format entire codebase with Biome |
-| `bun run check` | Run Biome check with auto-fix (lint + format + organize imports) |
-| `bun run check:ci` | Run Biome check without auto-fix (CI mode) |
-| `bun run dev` | Build and run with OpenCode |
+| `bun run build` | Build plugin + CLI bundles to `dist/`, emit declarations, regenerate config schema |
+| `bun run typecheck` | `tsc --noEmit` — type check only |
+| `bun test` | Run all tests (root is `./src` via `bunfig.toml`) |
+| `bun run check` | Biome check with auto-fix (lint + format + organize imports) |
+| `bun run check:ci` | Biome check without auto-fix (CI gate) |
+| `bun run dev` | Build then launch OpenCode |
+| `bun run generate-schema` | Regenerate `oh-my-opencode-slim.schema.json` from Zod schema |
 
-**Running a single test:** Use Bun's test filtering with the `-t` flag:
-```bash
-bun test -t "test-name-pattern"
-```
+**Single test:** `bun test -t "test-name-pattern"`
+**Single file:** `bun test src/config/loader.test.ts`
 
-## Code Style
+### Build internals
 
-### General Rules
-- **Formatter/Linter:** Biome (configured in `biome.json`)
-- **Line width:** 80 characters
-- **Indentation:** 2 spaces
-- **Line endings:** LF (Unix)
-- **Quotes:** Single quotes in JavaScript/TypeScript
-- **Trailing commas:** Always enabled
+The build is a 4-step pipeline (`bun run build`):
 
-### TypeScript Guidelines
-- **Strict mode:** Enabled in `tsconfig.json`
-- **No explicit `any`:** Generates a linter warning (disabled for test files)
-- **Module resolution:** `bundler` strategy
-- **Declarations:** Generate `.d.ts` files in `dist/`
+1. `build:plugin` — Bun-bundle `src/index.ts` → `dist/index.js` (ESM, Node target)
+2. `build:cli` — Bun-bundle `src/cli/index.ts` → `dist/cli/index.js`
+3. `tsc --emitDeclarationOnly` — Generate `.d.ts` files into `dist/`
+4. `generate-schema` — Produce JSON Schema from the Zod config schema
 
-### Imports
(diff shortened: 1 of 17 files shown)


dhaern:
> ## Summary
> When a delegated subagent task (e.g. `@fixer`) hits a provider error or is interrupted mid-run, it may return an empty `<task_result>` even though the underlying session contains useful partial state. Previously, the orchestrator would treat this as a dead task and spawn a fresh session, losing all context.
> 
> This PR adds the **recovery note mechanism** requested in issue #387:
> 
> * When a `task` tool result is empty/interrupted, a `[task partial state available]` note is appended containing the `task_id`
> * The note is **informational** — it tells the orchestrator a session exists that can be resumed, without auto-resuming
> * The orchestrator's prompt now includes a **Session Resumption** section explaining `task_id` usage
> 
> ## What was built
> ### New files
> File	Purpose
> `src/utils/task-session-tracker.ts`	Tracks active subagent sessions by `task_id` → metadata
> `src/hooks/delegate-task-resume/detector.ts`	Detects empty/interrupted task results; parses `task_id` from XML output
> `src/hooks/delegate-task-resume/guidance.ts`	Builds the `[task partial state available]` recovery note
> `src/hooks/delegate-task-resume/hook.ts`	`tool.execute.after` hook wiring
> `src/hooks/delegate-task-resume/index.test.ts`	20 tests (detector, guidance, hook integration)
> `src/utils/task-session-tracker.test.ts`	10 tests
> ### Modified files
> File	Change
> `src/agents/orchestrator.ts`	Added `## 7. Session Resumption` section to prompt — explains `task_id`, when to resume vs. fresh
> `src/hooks/index.ts`	Export `createDelegateTaskResumeHook`
> `src/index.ts`	Instantiate `TaskSessionTracker` + hook; wire into `tool.execute.after`, `session.created`, `session.deleted`
> `src/interview/interview.test.ts`	Biome format fix (existing)
> `src/utils/system-collapse.test.ts`	Biome import sort fix (existing)
> ### Key design decisions
> 1. **Informational, not automatic** — the recovery note is injected into the task output so the LLM decides whether to resume; no auto-resume
> 2. **`task_id` parsed from XML output** — the `task` tool returns `<task_id>ses_...</task_id>` in its output; we parse this rather than using `sessionID` which refers to the orchestrator's session (not the subagent's)
> 3. **Distinguishes from `delegate-task-retry`** — that hook handles parameter validation errors; this hook handles provider/interrupt errors
> 4. **Agent-from-tracker** — when the tracker has an entry, the note includes `@fixer` etc. so the orchestrator knows which specialist session it was
> 
> ## Verification
> * **951 tests** pass (32 new tests added)
> * **Typecheck**: `tsc --noEmit` clean
> * **Lint/format**: `check:ci` zero errors, zero warnings
> * **Build**: `bun run build` completes successfully
> 
> ## Related
> Fixes the core mechanism for [#387](https://github.com/alvinunreal/oh-my-opencode-slim/issues/387) (session resumption).

I noticed a couple of things that may be worth checking before merging:

## 1. `task_id` parser may not match OpenCode’s real task output

In this PR, `parseTaskId()` expects XML-style output:

```ts
const match = output.match(/<task_id>([^<]+)<\/task_id>/);
```

But the OpenCode task tool output I see in practice is line-based:

task_id: ses_xxx (for resuming to continue this task if needed)

<task_result>
...
</task_result>

So for the main interrupted-task case:

task_id: ses_xxx (for resuming to continue this task if needed)

<task_result>
</task_result>

parseTaskId() would return undefined, and the recovery note would not be appended.

PR #390 added parseTaskIdFromTaskOutput() in src/utils/task.ts, which parses this real format:

/^task_id:\s*([^\s()]+)(?:\s*\(.*)?$/

Maybe this PR should reuse that helper, or support both formats:

task_id: ses_xxx
<task_id>ses_xxx</task_id>

## 2. This overlaps with the new session manager from #390

#390 already introduced:

- src/hooks/task-session-manager/
- src/utils/session-manager.ts
- src/utils/task.ts

This PR adds a separate:

- src/hooks/delegate-task-resume/
- src/utils/task-session-tracker.ts

The concepts are related but currently split across two systems:

- #390 remembers resumable sessions and aliases
- #392 tracks interrupted/completed status and appends recovery guidance

That can work, but it may become harder to reason about over time. It might be cleaner to either:

- integrate this recovery note into the task-session-manager system, or
- at least reuse the same task-id parser/helper and make the boundary between the two systems explicit.

## 3. The recovery note says the session “may contain partial state”, but does not verify it

The current hook infers recoverability from the task output being empty/interrupted. It does not inspect:

ctx.client.session.messages({ path: { id: taskId } })

That is probably fine for a first implementation, but it means the note is heuristic. In my local test, the interrupted subagent session really did contain partial state and resuming the same task_id worked, but in other cases the session may be empty.

Maybe the wording should stay conservative (may contain partial state) unless the hook actually checks session messages.

## 4. Suggested minimal changes

I think the core idea is good. The minimum I would change before merge is:

1. Reuse or align with parseTaskIdFromTaskOutput() from #390.
2. Add a test for the actual OpenCode task output format:

task_id: ses_child1 (for resuming to continue this task if needed)

<task_result>
</task_result>

3. Keep the note informational, as it already is.
4. Consider whether this should live under task-session-manager or remain separate with clearly shared helpers.

Overall: good direction, but the parser mismatch may prevent the PR from covering the real interrupted-task case.

@greptile-apps

SweetSophia:
https://github.com/alvinunreal/oh-my-opencode-slim/pull/392#issuecomment-4305739079
It has been addressed with the last 2 commits

dhaern:
I rebased/reworked this PR on top of the current `master` and removed the duplicated architecture from the original version.

The PR now integrates the useful recovery behavior directly into the existing `task-session-manager`:

- no separate `TaskSessionTracker`
- no separate `delegate-task-resume` hook
- no unrelated `AGENTS.md` or formatting changes
- recovery notes are only appended for task outputs with a parseable `task_id` and recoverable interruption signals
- validation/argument errors and missing-session errors are intentionally ignored

Validated with:

- `bun test`
- `bun run typecheck`
- `bun run lint`
- `git diff --check`

PR #393, opened 2026-04-23
fix: prevent duplicate tmux pane spawns

alvinunreal:
## Summary
- add an in-flight session spawn guard so create/busy events cannot open two tmux panes for one child session
- apply the same protection to busy-triggered respawns while a pane spawn is already pending
- add regression tests covering duplicate create events and create-plus-busy race conditions


alvinunreal:
@greptile review

PR #394, opened 2026-04-23
feat: add Enter shortcut for interview questions

alvinunreal:
## Summary
- add plain Enter handling in the interview UI so the current selected answer can advance to the next question or submit on the last question
- keep Enter untouched inside text inputs and textareas, and guard against repeat/double-submit behavior while answers are posting
- cover the new keyboard flow in interview page tests and fix an import-order issue required for `bun run check:ci`

## Validation
- `bun run check:ci`
- `bun run typecheck`
- `bun test`

### DIFF
--- a/src/interview/interview.test.ts
+++ b/src/interview/interview.test.ts
@@ -1724,40 +1724,6 @@ describe('renderInterviewPage', () => {
     );
   });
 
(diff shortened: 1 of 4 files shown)


alvinunreal:
@greptile

alvinunreal:
@greptileai

PR #402, opened 2026-04-24
fix: improve plugin-side performance and reduce runtime overhead

dhaern:
## Summary

This PR reduces avoidable plugin-side overhead in hot paths used by agent sessions, subagents, transforms, logging, and optional multiplexer integrations.

The goal is to keep behavior equivalent while making common paths cheaper and avoiding work that cannot be useful in the current runtime context.

### Non-blocking logger writes

`log()` no longer writes with `appendFileSync()` on every call. Writes are now queued with async `appendFile()` calls, preserving log order while avoiding event-loop blocking during frequent hooks and subagent-heavy sessions.

Safety / tradeoff:
- plugin behavior is unchanged
- log order is preserved through a write chain
- if the process is killed abruptly, the very latest queued diagnostic log entry may be lost
- tests use `flushLoggerForTesting()` to keep log assertions deterministic

### Faster system prompt collapse

`collapseSystemInPlace()` now has fast paths for empty and single-entry arrays.

This avoids unnecessary `join()` work in the common case while preserving the previous behavior, including cleanup of a single empty system entry.

### Precompiled display-name mention rewriting

Display-name mention rewriting is now prepared once during plugin initialization with `createDisplayNameMentionRewriter(config)`.

Previously, each message transform rebuilt runtime agent names and regular expressions. Now the transform applies a precompiled rewriter function.

Safety:
- `@displayName` still rewrites to the internal agent name
- internal agent names still work
- custom agents remain supported
- the legacy `rewriteDisplayNameMentions(config, text)` helper is preserved
- display-name changes still require reloading the plugin, matching the existing config-loading behavior

### Precomputed skill permission rules per agent

The available-skills filter now caches only permission rules per agent.

It intentionally does not cache the final `<available_skills>` block. If OpenCode injects a different block later, Slim still filters the current block correctly on every request.

Safety:
- explicit `skills` config remains honored
- `skills: []` still means no skills
- wildcard allow/deny behavior is unchanged
- orchestrator defaults are preserved
- tests cover that the final skills block text is not cached

### Image hook fast path

The image hook now scans for actual image parts before creating `.opencode/images`, writing `.gitignore`, or doing heavier filesystem work.

If no images are present:
- no image directory is created
- no `.gitignore` is written
- if `.opencode/images` already exists, the existing debounced cleanup still runs

If images are present, behavior remains the same.

### Multiplexer-aware council launch delays

Council councillor launch staggering now only applies when a real multiplexer session is active.

The stagger exists to avoid tmux/zellij pane creation collisions. Outside tmux/zellij it only adds latency, so it is skipped when no multiplexer panes can be created.

Safety:
- `multiplexer.type = "none"` is unchanged
- `auto` outside tmux/zellij remains disabled
- explicit tmux/zellij config outside an actual session no longer pays multiplexer-only delays
- inside tmux/zellij, the stagger remains

### Safer Copilot chat header caching tests

The Copilot internal-initiator cache continues to cache positive marker detections only. It does not cache negative lookups, avoiding edge cases where a transient or incomplete message read could cache `false` before a marker is available.

Additional tests cover:
- positive cache behavior
- transient lookup failures
- cache key separation by `sessionID:messageID`

## Compatibility

This PR does not touch OpenCode core or the TUI renderer.

It does not remove features, truncate output, restrict tools, change default permissions, or alter agent behavior. The changes either preserve current semantics or skip work that is only useful in unavailable runtime contexts, such as multiplexer pane delays outside a real tmux/zellij session.

## Validation

Validated with:

```bash
bun run typecheck
bun test
bun run build
bunx biome check <modified files>
git diff --check
```

Full test result:

```text
948 pass
0 fail
```

## Diff size

```text
Total added lines: 309
Total deleted lines: 47
Net change: +262

Test added lines: 178
Test deleted lines: 11
Net test change: +167

Non-test added lines: 131
Non-test deleted lines: 36
Net non-test change: +95
```

### DIFF
--- a/src/council/council-manager.ts
+++ b/src/council/council-manager.ts
@@ -350,8 +350,9 @@ export class CouncilManager {
       // Parallel execution (default): run all councillors concurrently
       const promises = entries.map(([name, config], index) =>
         (async () => {
-          // Stagger launches to avoid tmux split-window collisions
-          if (index > 0) {
+          // Stagger launches only when multiplexer panes can be created.
+          // Outside tmux/zellij this delay only adds latency with no benefit.
+          if (this.tmuxEnabled && index > 0) {
             await new Promise((r) =>
               setTimeout(r, index * COUNCILLOR_STAGGER_MS),
             );
--- a/src/hooks/chat-headers.test.ts
+++ b/src/hooks/chat-headers.test.ts
@@ -23,13 +23,15 @@ function createMockContext(parts: unknown[] = []) {
 
 function createInput(
   overrides?: Partial<{
+    sessionID: string;
     providerID: string;
     npm: string;
     messageID: string;
   }>,
 ) {
+  const sessionID = overrides?.sessionID ?? 'session-1';
   return {
-    sessionID: 'session-1',
+    sessionID,
     agent: 'orchestrator',
     model: {
       id: 'github-copilot/claude',
@@ -80,7 +82,7 @@ function createInput(
     },
     message: {
       id: overrides?.messageID ?? 'message-1',
-      sessionID: 'session-1',
+      sessionID,
       role: 'user' as const,
       time: { created: Date.now() },
       agent: 'orchestrator',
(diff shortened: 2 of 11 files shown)


SweetSophia (approved):
## Peer Review: PR #402 — fix: improve plugin-side performance and reduce runtime overhead

Reviewed the full diff (11 files, +309/-47). Seven distinct optimizations, all well-contained.

### Each Change Reviewed

1. **Logger async writes** — Promise-chain  preserves order while removing event-loop blocking. Trade-off (last log entry may be lost on abrupt kill) is explicitly documented. Tests updated with  for determinism. ✅
2. **System collapse fast paths** — Early returns for empty/single-entry arrays avoid  overhead in the common case. New test for single empty-string cleanup. ✅
3. **Precompiled display-name rewriter** — Regex patterns computed once at plugin init. Legacy  helper preserved for backward compat. ✅
4. **Precomputed skill permission rules** —  Map caches per-agent rules. Test confirms final skills block text is NOT cached (different block content per call still works correctly). ✅
5. **Image hook fast path** — Scans for image parts before creating . If no images, no directory/GitIgnore creation. Behavior unchanged when images are present. ✅
6. **Multiplexer-aware launch stagger** — Added  guard so stagger only fires when panes can actually be created. Matches documented safety guarantees. ✅
7. **Copilot transient cache handling** — Three new tests cover: positive cache hit, transient failure retry, and sessionID:messageID cache key separation. ✅

### Compatibility ✅
No features removed. No breaking API changes. No default permission changes. All changes either preserve semantics or skip work that is only useful in unavailable runtime contexts.

### Validation ✅
- PR is MERGEABLE ✅
- PR description: typecheck, test, build, biome check, git diff-check all pass (948 pass / 0 fail)
- Validation commands include Bundled 169 modules in 68ms

  index.js  1.0 MB  (entry point)

Bundled 17 modules in 16ms

  index.js  41.92 KB  (entry point)

✅ Schema written to /home/niya/github/oh-my-opencode-slim/oh-my-opencode-slim.schema.json and Checked 188 files in 136ms. No fixes applied. in addition to standard checks

### ⚠️ Large PR Flag
This PR touches 11 files (+309/-47), exceeding the 10-file threshold. However, each change is isolated to its own module, covered by targeted tests, and follows existing codebase conventions. The scope is wide but not deep. Leader acknowledgment recommended but not blocking.

**Decision: APPROVE.** All changes are sound, well-tested, and compatible. Ready to merge.

SweetSophia (approved):
## Peer Review: PR #402 — fix: improve plugin-side performance and reduce runtime overhead

Reviewed the full diff (11 files, +309/-47). Seven distinct optimizations, all well-contained.

### Each Change Reviewed

1. **Logger async writes** — Promise-chain appendFile preserves order while removing event-loop blocking. Trade-off (last log entry may be lost on abrupt kill) is explicitly documented. Tests updated with flushLoggerForTesting() for determinism. ✅
2. **System collapse fast paths** — Early returns for empty/single-entry arrays avoid join() overhead in the common case. New test for single empty-string cleanup. ✅
3. **Precompiled display-name rewriter** — Regex patterns computed once at plugin init. Legacy rewriteDisplayNameMentions(config, text) helper preserved for backward compat. ✅
4. **Precomputed skill permission rules** — permissionRulesByAgent Map caches per-agent rules. Test confirms final skills block text is NOT cached (different block content per call still works correctly). ✅
5. **Image hook fast path** — Scans for image parts before creating .opencode/images/. If no images, no directory/GitIgnore creation. Behavior unchanged when images are present. ✅
6. **Multiplexer-aware launch stagger** — Added this.tmuxEnabled guard so stagger only fires when panes can actually be created. Matches documented safety guarantees. ✅
7. **Copilot transient cache handling** — Three new tests cover: positive cache hit, transient failure retry, and sessionID:messageID cache key separation. ✅

### Compatibility ✅
No features removed. No breaking API changes. No default permission changes. All changes either preserve semantics or skip work that is only useful in unavailable runtime contexts.

### Validation ✅
- PR is MERGEABLE ✅
- PR description: typecheck, test, build, biome check, git diff-check all pass (948 pass / 0 fail)
- Validation commands include bun run build and bunx biome check in addition to standard checks

### ⚠️ Large PR Flag
This PR touches 11 files (+309/-47), exceeding the 10-file threshold. However, each change is isolated to its own module, covered by targeted tests, and follows existing codebase conventions. The scope is wide but not deep. Leader acknowledgment recommended but not blocking.

**Decision: APPROVE.** All changes are sound, well-tested, and compatible. Ready to merge.

alvinunreal:
Thanks @dhaern

PR #406, opened 2026-04-24
fix: improve apply_patch stale patch rescue performance

dhaern:
## Summary

This PR improves the plugin-side `apply_patch` rescue layer for stale patches while preserving OpenCode’s native `apply_patch` execution and rendering flow.

The hook still runs before native execution and only rewrites patch text when it can produce a safer, canonical patch. Native OpenCode remains responsible for applying and rendering the final patch.

The changes are limited to measured production paths: parsing, tolerant matching, stale patch rescue, and safety guardrails.

Highlights:

- Faster prefix/suffix rescue for stale chunks.
- Faster LCS rescue with bounded candidate search.
- Faster anchored insertion resolution.
- Safer tolerant matching for unicode and trailing-whitespace differences.
- Better preservation of patch boundaries and change contexts.
- Cleaner rewrite result shape focused only on production behavior.
- Additional guardrail tests for ambiguity, indentation safety, tolerant anchors, and parser edge cases.

## Why

Stale patches can happen when a patch is generated against slightly outdated file contents. The existing rescue layer already handled many of these cases safely, but large files or repeated candidate windows could make rescue noticeably slower.

This PR improves the measured hot paths while keeping the safety model conservative:

- exact patches still pass through unchanged;
- recoverable stale patches are canonicalized more efficiently;
- ambiguous matches still fail before native execution;
- indentation-sensitive cases remain protected;
- native `apply_patch` behavior stays intact.

## Performance

Measured from real hook execution during interactive `apply_patch` usage.

### Before

| Scenario | Hook time |
|---|---:|
| prefix/suffix rescue | ~805 ms |
| LCS rescue | ~625 ms |
| anchored insertion | ~305 ms |

### After

| Scenario | Hook time |
|---|---:|
| prefix/suffix rescue | ~419–454 ms |
| LCS rescue | ~268–294 ms |
| anchored insertion | ~109–116 ms |

Approximate improvement:

| Scenario | Improvement |
|---|---:|
| prefix/suffix rescue | ~44–48% faster |
| LCS rescue | ~53–57% faster |
| anchored insertion | ~62–64% faster |

A representative mixed-code patch through the real hook path completed in ~17 ms and passed through unchanged when no rescue was needed.

## Implementation Notes

This focuses on the hot paths that showed measurable impact:

- Add File parsing avoids repeated string concatenation.
- Matching reuses prepared normalized targets instead of repeatedly recomputing them.
- Prefix/suffix rescue avoids repeated full scans for common one-line anchors.
- LCS rescue uses lower memory, pre-normalized inputs, cheap pruning, and bounded candidate search.
- Anchor resolution performs a single tolerant scan while preserving ambiguity checks.

### Matching and rescue

The matching layer now avoids repeated work in the highest-impact paths:

- precomputed tolerant comparison targets;
- optimized one-line prefix/suffix lookup;
- binary search over ordered right-side hits;
- O(m) LCS memory usage;
- pre-normalized LCS inputs;
- cheap shared-line upper bound before scoring;
- border-anchored LCS candidate collection;
- bounded LCS search.

Automatic tolerant rescue remains limited to safe comparisons:

- exact;
- unicode-normalized exact;
- trim-end;
- unicode + trim-end.

Full-trim matching remains excluded from automatic rescue to avoid unsafe indentation changes.

### Parser behavior

Patch parsing now better distinguishes real patch boundaries from patch body content.

This allows lines such as:

```text
 *** End Patch
```

to remain valid hunk content, while real patch boundaries are still recognized reliably.

Change contexts also preserve meaningful trailing spaces, which improves canonicalization for trim-end anchored insertions.

### Rewrite flow

The rewrite result now stays focused on production behavior:

- `patchText`
- `changed`

Diagnostic-only metadata was removed from the runtime path.

## Safety

The hook remains fail-closed before native execution.

Protected cases include:

- ambiguous prefix/suffix rescue;
- ambiguous insertion anchors;
- missing insertion anchors;
- unsafe indentation-only rescue;
- malformed patch bodies;
- path/worktree validation errors.

The PR improves successful rescue cases without expanding the class of patches that can be silently rewritten.

## Tests

Added and updated coverage includes:

- patch boundary parsing when `*** End Patch` appears as hunk content;
- trailing-space change context preservation;
- unicode + trim-end prefix/suffix rescue;
- one-line prefix/suffix ambiguity detection;
- right-side hit ordering for prefix/suffix rescue;
- LCS candidate pruning and bounded scoring behavior;
- trim-end anchored insertion canonicalization;
- tolerant anchor ambiguity detection;
- removal of runtime diagnostic metadata expectations.

Validation:

```text
focused apply-patch tests: 130 pass, 0 fail
typecheck: pass
biome check: pass
full test suite: 964 pass, 0 fail
build: pass
git diff --check: pass
```

Additional real `apply_patch` smoke coverage was also verified through the interactive tool path:

```text
prefix/suffix stale rescue: pass
trim-end anchored insertion: pass
LCS stale rescue: pass
ambiguous anchor fail-closed behavior: pass
```

## Diff Size

```text
9 files changed, 382 insertions(+), 182 deletions(-)
net: +200
```

Production code only:

```text
+272 / -172
net: +100
```

Tests:

```text
+110 / -10
net: +100
```

### DIFF
--- a/src/hooks/apply-patch/codec.test.ts
+++ b/src/hooks/apply-patch/codec.test.ts
@@ -82,6 +82,31 @@ PATCH`);
     ]);
   });
 
+  test('parsePatchStrict preserves End Patch text when it is hunk context', () => {
+    const markerPadding = '  ';
+    const parsed = parsePatchStrict(`*** Begin Patch${markerPadding}
+*** Update File: sample.txt
+@@ marker
+ *** End Patch
+ keep
+*** End Patch${markerPadding}`);
+
+    expect(parsed.hunks).toEqual([
+      {
+        type: 'update',
+        path: 'sample.txt',
+        chunks: [
+          {
+            old_lines: ['*** End Patch', 'keep'],
+            new_lines: ['*** End Patch', 'keep'],
+            change_context: 'marker',
+            is_end_of_file: undefined,
+          },
+        ],
+      },
+    ]);
+  });
+
   test('parsePatchStrict fails on garbage inside @@', () => {
     expect(() =>
       parsePatchStrict(`*** Begin Patch
--- a/src/hooks/apply-patch/codec.ts
+++ b/src/hooks/apply-patch/codec.ts
@@ -68,6 +68,19 @@ function unexpectedPatchLine(context: string, line: string): never {
   );
 }
 
+function parseChangeContext(line: string): string | undefined {
+  const context = line.slice(2);
+  if (context.length === 0) {
+    return undefined;
+  }
+
+  return context.startsWith(' ') ? context.slice(1) || undefined : context;
+}
+
+function isPatchBoundary(line: string, marker: string): boolean {
+  return line.trimEnd() === marker;
+}
+
 function parseChunks(lines: string[], index: number, mode: ParseMode) {
   const chunks: PatchChunk[] = [];
   let at = index;
@@ -81,7 +94,7 @@ function parseChunks(lines: string[], index: number, mode: ParseMode) {
       continue;
     }
 
-    const context = lines[at].slice(2).trim() || undefined;
+    const context = parseChangeContext(lines[at]);
     at += 1;
 
     const old_lines: string[] = [];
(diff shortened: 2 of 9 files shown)


dhaern:
Thanks @greptile-apps for the review.

Addressed the two notes in `5bc93aa`:

- Removed the unreachable post-loop ambiguity guard in `rescueByPrefixSuffix`.
- Added an inline comment explaining why the one-line prefix/suffix fast path compares at the broadest safe automatic level while still excluding full-trim rescue.

Validation after the update:

```text
focused apply-patch tests: 130 pass, 0 fail
biome matching.ts: pass
git diff --check: pass
```

CI is passing on the updated branch as well.

Issue #415, opened 2026-04-27
[Bug]: Multiple hooks inject into system messages in cache-busting ways


alvinreal:
## Summary

Three hooks in the plugin inject content into system/user messages in ways that **bust LLM provider cache**, causing unnecessary token usage and increased latency. The issue is that modifications at the **head** of the conversation (system prompt or prepending to user messages) invalidate the entire cached prefix.

### 1. Phase Reminder Hook (`src/hooks/phase-reminder/index.ts`)
- **Hook**: `experimental.chat.messages.transform`
- **Problem**: Prepends phase reminder to the **last user message's text part** on every request
- **Impact**: Every request has different user message content → cache miss
- **Code**: Lines 63-64 prepend `${PHASE_REMINDER}\n\n---\n\n${originalText}`

### 2. Todo Hygiene Hook (`src/hooks/todo-continuation/todo-hygiene.ts`)
- **Hook**: `experimental.chat.system.transform`
- **Problem**: Injects dynamic reminders based on **runtime todo state** (which changes constantly)
- **Impact**: System prompt varies between requests → cache miss
- **Code**: Different reminders for `general` vs `final_active` states (lines 107-135)

### 3. Post-File-Tool Nudge Hook (`src/hooks/post-file-tool-nudge/index.ts`)
- **Hook**: `experimental.chat.system.transform`
- **Problem**: Uses **stateful pending set** to conditionally inject after file reads/writes
- **Impact**: Non-deterministic system message content based on prior tool execution
- **Code**: Lines 34-37 only inject if `pendingSessionIds.delete(input.sessionID)` returns true

## Why This Breaks Caching

LLM providers (OpenAI, Anthropic, etc.) cache responses based on a **prefix hash** of:
- Model ID
- Message content (system + conversation history)
- Temperature and other params

When hooks inject **varying content** at the **head** (beginning) of the conversation:
1. The prefix hash changes
2. Cache lookup fails
3. Full re-computation required
4. **Result**: Wasted tokens (up to 300K+ in long sessions) and increased latency

## Reference Implementation

The [cortexkit/opencode-magic-context](https://github.com/cortexkit/opencode-magic-context) plugin handles this correctly by:
1. **Appending nudges to the tail** (last user message), not head
2. Using **stable system prompts** that don't change between requests
3. Tracking nudge placement with `nudgePlacementStore` to avoid duplicates

## Recommended Fixes

| Hook | Current | Recommended |
[...]

PR #416, opened 2026-04-27
fix(todo-hygiene): keep dynamic reminders cache-friendly

dhaern:
## Summary

- Move `todo-hygiene` dynamic reminders out of `experimental.chat.system.transform`.
- Inject the reminder ephemerally at the tail of the latest user message via `experimental.chat.messages.transform`.
- Preserve the existing dynamic intra-turn behavior instead of replacing it with static guidance.
- Add per-session debounce and defensive fail-open handling so `todo-hygiene` does not repeatedly inspect todos or block message dispatch.

## Context

Related to #415.

Issue #415 correctly points out that dynamic text injected through `experimental.chat.system.transform` can hurt prompt caching because it changes the system prompt.

However, `todo-hygiene` needs to be treated differently from static reminders: its behavior depends on runtime todo state during the assistant turn.

The current behavior is intentionally dynamic:

1. `todowrite` starts or resets the hygiene cycle.
2. Later tool activity inspects the current todo state.
3. The hook arms a one-shot reminder when needed.
4. If only one active todo remains, the stronger `final_active` reminder wins.
5. The next model call in the same request receives the reminder.

Moving this to a static system prompt would lose that stateful behavior.

## What changed

Instead of appending dynamic todo reminders to the system prompt, this PR injects them ephemerally into the outgoing message payload:

```diff
- output.system.push(dynamicTodoReminder)
+ append dynamicTodoReminder to the latest user message in messages.transform
```

The injected reminder is wrapped as:

```xml
<instruction name="todo_hygiene">
...
</instruction>
```

The wrapper is stripped/deduplicated during message transforms so repeated transforms do not accumulate stale reminders or affect request-boundary detection.

## Cache behavior

This keeps the system prompt stable while still allowing `todo-hygiene` to provide runtime guidance.

The reminder remains ephemeral:

- it is not persisted with `session.prompt`;
- it is not stored as normal chat history;
- it is only added to the outbound model payload when there is pending todo-hygiene state.

## Performance and safety

This PR also adds:

- a per-session debounce window to avoid repeated todo-state inspections while a reminder is already pending;
- early `shouldInject` gating so non-orchestrator/subagent sessions do not arm todo-hygiene state;
- fail-open handling in `messages.transform`;
- a short timeout when consuming pending reminders, so todo-hygiene cannot block prompt dispatch if todo-state lookup is slow or unavailable;
- no synthetic text parts for attachment-only payloads.

## Validation

```sh
bun test src/hooks/todo-continuation
bun run typecheck
```

### DIFF
--- a/src/hooks/todo-continuation/index.test.ts
+++ b/src/hooks/todo-continuation/index.test.ts
@@ -84,6 +84,14 @@ describe('createTodoContinuationHook', () => {
     };
   }
 
+  function allMessageText(output: {
+    messages: Array<{ parts: Array<{ text?: string }> }>;
+  }): string {
+    return output.messages
+      .flatMap((message) => message.parts.map((part) => part.text ?? ''))
+      .join('\n');
+  }
+
   describe('tool toggle', () => {
     test('calling auto_continue execute with { enabled: true } sets state', async () => {
       const ctx = createMockContext();
@@ -115,15 +123,13 @@ describe('createTodoContinuationHook', () => {
         },
       });
       const hook = createTodoContinuationHook(ctx);
-      const system = { system: ['base'] };
+      const output = userMessages('continue previous work', 'sub1', 'explorer');
 
-      await hook.handleMessagesTransform(
-        userMessages('continue previous work', 'sub1', 'explorer'),
-      );
+      await hook.handleMessagesTransform(output);
       await hook.handleToolExecuteAfter({ tool: 'task', sessionID: 'sub1' });
-      await hook.handleChatSystemTransform({ sessionID: 'sub1' }, system);
+      await hook.handleMessagesTransform(output);
 
-      expect(system.system.join('\n')).not.toContain(TODO_HYGIENE_REMINDER);
+      expect(allMessageText(output)).not.toContain(TODO_HYGIENE_REMINDER);
     });
 
     test('does not inject anything at request start before todowrite', async () => {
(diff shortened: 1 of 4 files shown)


dhaern:
@greptile-apps Thanks for flagging this. I pushed a follow-up fix in 20d0af0 that removes the delayed-consumption race instead of trying to cancel the timed-out promise.

What changed:

- `messages.transform` no longer performs any async todo lookup or timeout-wrapped consumption.
- `consumePendingReminder()` is now synchronous/in-memory: it only consumes already-armed pending state.
- Todo-state inspection remains in `tool.execute.after`, where the reminder is armed.
- This keeps the anti-blocking behavior for prompt dispatch while avoiding the abandoned-promise path that could delete pending state after a timeout.

Added/updated tests:

- `messages transform injects pending reminder without todo lookup`
- `consuming a pending reminder does not inspect todos`

Validation:

```sh
bun test src/hooks/todo-continuation
bun run typecheck
```

Both pass.

dhaern:
@greptile-apps Addressed the cosmetic P2 log note in b5a2292: the log now says `Consumed todo hygiene reminder` instead of implying injection already happened. Keeping the consumption-time validation trade-off intentionally to keep `messages.transform` sync/in-memory.

alvinunreal:
thanks @dhaern - included those changes in my PR

PR #419, opened 2026-04-27
Fix cache-busting reminder injections

### LINKED ISSUE #415: [Bug]: Multiple hooks inject into system messages in cache-busting ways

alvinunreal:
## Summary
- Closes #415.
- Keeps the orchestrator phase reminder active on every orchestrator turn, but appends it at the final latest user-message tail instead of prepending it or putting it in system.
- Stops dynamic todo hygiene, post-file-tool, and resumable-session state from mutating the cached system prompt.
- Preserves runtime nudges at cache-safe tails: post-file reminders stay in file-tool output, todo hygiene reminders are request-local message-transform instructions, and resumable session context is appended to the latest orchestrator user message before the phase reminder.

## What changed
- `phase-reminder` now appends `<internal_reminder>...</internal_reminder>` after the latest orchestrator user text. It runs after resumable-session injection, skips non-orchestrator and internal notification turns, and avoids duplicate reminders.
- `post-file-tool-nudge` appends the reminder to Read/Write tool output instead of setting pending state for system injection. This nudge is intentionally retained because post-read/write drift is still a useful trigger.
- `todo-continuation` follows the PR #416 transport model: tool activity arms pending reminder state, and `messages.transform` reads that state into a trailing `<instruction name="todo_hygiene">...</instruction>` on the latest orchestrator user message. This keeps todo reminders out of persisted tool output while preserving dynamic intra-turn behavior.
- Todo reminder reads are non-destructive so compaction transforms over cloned history cannot consume/drop a pending live reminder; new request boundaries and session cleanup still clear pending state.
- `task-session-manager` injects `<resumable_sessions>` into request-local message context instead of `experimental.chat.system.transform`.
- Removed the dynamic system-transform hooks/no-op compatibility shims, so the plugin no longer wires unused cache-sensitive reminder paths.

## Why this fixes the issue
OpenCode/provider prompt caching depends on stable cached prefix content. Before this change, runtime state such as file-tool activity, todo state, resumable sessions, and per-request reminders could alter system or prepend content ahead of the user's turn. This PR keeps dynamic reminders out of system and places request-local guidance at the tail/tool-output context, so reminders remain visible without invalidating the stable system prompt.

For todo hygiene specifically, OpenCode persists mutated tool outputs into tool history, so the reminder is no longer appended to `output.output`. It is armed in memory from `tool.execute.after` and injected only into outbound model payloads from `messages.transform`; compaction-safe non-destructive reads prevent cloned compaction transforms from dropping pending reminders.

## Verification
- `bun test src/hooks/phase-reminder/index.test.ts src/hooks/post-file-tool-nudge/index.test.ts src/hooks/todo-continuation/index.test.ts src/hooks/todo-continuation/todo-hygiene.test.ts src/hooks/task-session-manager/index.test.ts`
- `bunx biome check src/hooks/todo-continuation/index.ts src/hooks/todo-continuation/index.test.ts src/hooks/todo-continuation/todo-hygiene.ts src/hooks/todo-continuation/todo-hygiene.test.ts`
- `bun run typecheck`
- `bun run build`
- Manual relaunch/log verification showed system hashes staying stable across plain messages, Read-tool nudges, and todo hygiene activity.

### DIFF
--- a/src/agents/orchestrator.ts
+++ b/src/agents/orchestrator.ts
@@ -1,4 +1,5 @@
 import type { AgentConfig } from '@opencode-ai/sdk/v2';
+import { PHASE_REMINDER_TEXT } from '../config/constants';
 
 export interface AgentDefinition {
   name: string;
@@ -211,6 +212,10 @@ ${enabledValidationRouting}
 - Verify solution meets requirements
 
 </Workflow>
+
+<WorkflowReminder>
+${PHASE_REMINDER_TEXT}
+</WorkflowReminder>
 
 <Communication>
 
--- a/src/agents/orchestrator.ts
+++ b/src/agents/orchestrator.ts
@@ -1,5 +1,4 @@
 import type { AgentConfig } from '@opencode-ai/sdk/v2';
-import { PHASE_REMINDER_TEXT } from '../config/constants';
 
 export interface AgentDefinition {
   name: string;
@@ -212,10 +211,6 @@ ${enabledValidationRouting}
 - Verify solution meets requirements
 
 </Workflow>
-
-<WorkflowReminder>
-${PHASE_REMINDER_TEXT}
-</WorkflowReminder>
 
 <Communication>
 
--- a/src/hooks/phase-reminder/index.test.ts
+++ b/src/hooks/phase-reminder/index.test.ts
@@ -1,9 +1,8 @@
 import { describe, expect, test } from 'bun:test';
-import { SLIM_INTERNAL_INITIATOR_MARKER } from '../../utils';
 import { createPhaseReminderHook, PHASE_REMINDER } from './index';
 
 describe('createPhaseReminderHook', () => {
-  test('prepends reminder for orchestrator sessions', async () => {
+  test('does not mutate orchestrator messages', async () => {
     const hook = createPhaseReminderHook();
     const output = {
       messages: [
@@ -16,9 +15,8 @@ describe('createPhaseReminderHook', () => {
 
     await hook['experimental.chat.messages.transform']({}, output);
 
-    expect(output.messages[0].parts[0].text).toBe(
-      `${PHASE_REMINDER}\n\n---\n\nhello`,
-    );
+    expect(output.messages[0].parts[0].text).toBe('hello');
+    expect(output.messages[0].parts[0].text).not.toContain(PHASE_REMINDER);
   });
 
   test('skips non-orchestrator sessions', async () => {
@@ -37,27 +35,22 @@ describe('createPhaseReminderHook', () => {
     expect(output.messages[0].parts[0].text).toBe('hello');
   });
 
-  test('skips internal notification turns', async () => {
+  test('does not mutate internal notification turns', async () => {
     const hook = createPhaseReminderHook();
+    const text =
+      '[Background task "x" completed]\n<!-- slim-internal-initiator -->';
     const output = {
       messages: [
         {
           info: { role: 'user' },
-          parts: [
-            {
-              type: 'text',
-              text: `[Background task "x" completed]\n${SLIM_INTERNAL_INITIATOR_MARKER}`,
-            },
-          ],
+          parts: [{ type: 'text', text }],
         },
       ],
     };
 
     await hook['experimental.chat.messages.transform']({}, output);
 
-    expect(output.messages[0].parts[0].text).toContain(
-      SLIM_INTERNAL_INITIATOR_MARKER,
-    );
+    expect(output.messages[0].parts[0].text).toBe(text);
     expect(output.messages[0].parts[0].text).not.toContain(PHASE_REMINDER);
   });
 });
--- a/src/hooks/phase-reminder/index.test.ts
+++ b/src/hooks/phase-reminder/index.test.ts
@@ -1,8 +1,9 @@
 import { describe, expect, test } from 'bun:test';
+import { SLIM_INTERNAL_INITIATOR_MARKER } from '../../utils';
 import { createPhaseReminderHook, PHASE_REMINDER } from './index';
 
 describe('createPhaseReminderHook', () => {
-  test('does not mutate orchestrator messages', async () => {
+  test('appends reminder for orchestrator sessions', async () => {
     const hook = createPhaseReminderHook();
     const output = {
       messages: [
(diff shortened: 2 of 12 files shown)


dhaern:
> ## Summary
> * Closes [[Bug]: Multiple hooks inject into system messages in cache-busting ways #415](https://github.com/alvinunreal/oh-my-opencode-slim/issues/415).
> * Keeps the orchestrator phase reminder active on every orchestrator turn, but appends it at the latest user-message tail instead of prepending it or putting it in system.
> * Stops dynamic todo hygiene, post-file-tool, and resumable-session state from mutating the cached system prompt.
> * Preserves runtime nudges by appending them at the tail: file/todo reminders go into tool output as `<internal_reminder>`, while resumable session context is appended to the latest orchestrator user message.
> 
> ## What changed
> * `phase-reminder` now appends `<reminder>...</reminder>` after the latest orchestrator user text. It skips non-orchestrator and internal notification turns and avoids duplicate reminders.
> * `post-file-tool-nudge` appends the reminder to Read/Write tool output instead of setting pending state for system injection.
> * `todo-continuation` injects hygiene/final-active reminders into the triggering tool output.
> * `task-session-manager` injects `<resumable_sessions>` into request-local message context instead of `experimental.chat.system.transform`.
> * Removed the dynamic system-transform hooks/no-op compatibility shims, so the plugin no longer wires unused cache-sensitive reminder paths.
> 
> ## Why this fixes the issue
> OpenCode/provider prompt caching depends on stable cached prefix content. Before this change, runtime state such as file-tool activity, todo state, resumable sessions, and per-request reminders could alter system or prepend content ahead of the user's turn. This PR keeps dynamic reminders out of system and places request-local guidance at the tail/tool-output context, so reminders remain visible without invalidating the stable system prompt.
> 
> ## Verification
> * `bun test src/hooks/phase-reminder/index.test.ts src/hooks/post-file-tool-nudge/index.test.ts src/hooks/todo-continuation/index.test.ts src/hooks/todo-continuation/todo-hygiene.test.ts src/hooks/task-session-manager/index.test.ts`
> * `bunx biome check src/index.ts src/hooks/post-file-tool-nudge/index.ts src/hooks/post-file-tool-nudge/index.test.ts src/hooks/todo-continuation/index.ts src/hooks/todo-continuation/index.test.ts src/hooks/todo-continuation/todo-hygiene.ts src/hooks/todo-continuation/todo-hygiene.test.ts src/hooks/task-session-manager/index.ts src/hooks/task-session-manager/index.test.ts`
> * `bun run typecheck`
> * `bun run build`
> * Manual relaunch/log verification showed system hashes staying stable across plain messages, Read-tool nudges, and todo hygiene activity.

I like the broader direction here: moving runtime reminders out of `system.transform` seems like the right fix for #415.

One thing I’m unsure about is the `todo-hygiene` transport. In #416, the reminder stayed pending in memory after `tool.execute.after` and was consumed synchronously in `messages.transform`, so it remained request-local and separate from tool output content.

This PR instead appends the reminder directly to `output.output`.

Do you think that could risk persisting the reminder as part of the tool result/history, or losing it when `output.output` is not a string? If not, I’m fine with this direction — I just wanted to double-check that the dynamic todo behavior from #416 is preserved.

alvinunreal:
@greptileai

alvinunreal:
Thanks for calling this out — you were right to question the todo-hygiene transport.

I checked PR #416 and the OpenCode source paths instead of guessing. `tool.execute.after` receives the real mutable tool output, and OpenCode persists `output.output` into the tool part state, so appending todo-hygiene reminders to `output.output` would make them durable tool-result history rather than request-local guidance.

I updated this PR to use the #416-style model:

- `tool.execute.after` only arms pending todo-hygiene state in memory.
- `messages.transform` reads that pending state synchronously and appends it to the latest orchestrator user message as `<instruction name="todo_hygiene">...</instruction>`.
- Todo reminders no longer mutate or persist into tool output.
- The `shouldInject` guard now also applies before `todowrite`/RESET final-active handling.
- Pending reminder reads are non-destructive, so compaction transforms over cloned history cannot consume/drop the live reminder.
- Added regression coverage for the compaction-like cloned transform case.

Validation is passing for the cache-related hook tests, changed-file Biome checks, typecheck, and build.

alvinunreal:
@greptileai

PR #420, opened 2026-04-27
fix(preset): persist runtime preset across plugin re-inits

ReqX:
## Summary

- Fix runtime preset switching (`/preset`) persisting across plugin re-init cycles
- Add module-level state that survives `dispose()`/re-init within the same process
- Add automatic reset of removed agents to config-file baseline when switching presets
- Add support for overriding `variant`, `temperature`, and `options` fields in config() hook re-inits (previously only `model` was handled)

## Problem

The `/preset` command calls `client.config.update()` which triggers `Instance.dispose()`. The plugin's `config()` hook re-runs with stale `modelArrayMap` from the initial load, reverting model changes back to config-file defaults. Runtime preset switches were unreliable after any config update.

## Solution

Three-layer defense:

1. **Module-level singleton** (`config/runtime-preset.ts`) stores the active runtime preset name, surviving `dispose()`/re-init cycles within the same process
2. **Config() hook override** patches stale model resolution with runtime preset values (including `variant`, `temperature`, `options`) after normal resolution runs, and resets agents from the previous preset to config-file baseline
3. **Init block re-merge** (defensive) re-merges runtime preset into `config.agents` in case OpenCode ever fully re-runs the plugin function

## Known limitations

- **Fallback chains not updated on preset switch**: The `ForegroundFallbackManager` and `runtimeChains` are built once during init and not rebuilt when the config() hook re-runs. If a preset changes models and the primary model gets rate-limited mid-conversation, runtime failover will try the previous preset's fallback models instead of the new preset's. This is a pre-existing limitation.

## Test plan

- [x] 10 new tests: runtime preset state management, stale state handling, rollback on error, factory sync
- [x] All 974 tests pass (964 existing + 10 new)
- [x] `bun run typecheck` clean
- [x] Backwards compatible: all new code gated on `getActiveRuntimePreset()` which returns `null` until `/preset` is called

## Files changed

| File | Change |
|------|--------|
| `src/config/runtime-preset.ts` | NEW: module-level singleton state with previous-preset tracking |
| `src/config/runtime-preset.test.ts` | NEW: 6 tests for state management |
| `src/config/loader.ts` | Export `deepMerge` for init block re-merge |
| `src/config/index.ts` | Barrel export update |
| `src/index.ts` | Init block re-merge, config() hook override (`model`/`variant`/`temperature`/`options`) + previous-preset reset |
| `src/tools/preset-manager.ts` | Alias resolution, reset diffs, error rollback, state sync |
| `src/tools/preset-manager.test.ts` | 4 new tests for stale state, rollback, factory sync |
| `docs/preset-switching.md` | Updated persistence semantics and reset behavior |

### DIFF
--- a/docs/preset-switching.md
+++ b/docs/preset-switching.md
@@ -13,7 +13,9 @@ Switch agent model presets at runtime without restarting OpenCode using the `/pr
 
 1. Define named presets in `oh-my-opencode-slim.jsonc` under the `presets` field
 2. Run `/preset <name>` to switch. The plugin calls the OpenCode SDK's `config.update()` method, which triggers a server-side cache invalidation
-3. The next LLM call uses the new models and settings
+3. Agents covered by the new preset get the preset's values
+4. Agents that were in the *previous* preset but are *not* in the new one are reset to their config-file baseline values
+5. The next LLM call uses the new models and settings
 
 ## Example Configuration
 
@@ -61,9 +63,9 @@ There are two ways to activate a preset:
 | Method | How | Persists? |
 |--------|-----|-----------|
 | Config file | Set `"preset": "cheap"` in `oh-my-opencode-slim.jsonc` | Yes, across restarts |
-| `/preset` command | Run `/preset cheap` during a session | No, reverts on restart |
+| `/preset` command | Run `/preset cheap` during a session | Across re-inits, not restarts |
 
-On restart, the plugin's `config()` hook re-applies the preset from the config file, overwriting any runtime switch. To make a runtime switch permanent, update the `"preset"` field in your config file.
+Runtime preset switches persist across plugin re-inits (triggered by config changes, etc.) within the same process, but revert on process restart. On restart, the plugin applies the preset from the config file. To make a runtime switch permanent, update the `"preset"` field in your config file.
 
 ## Example Output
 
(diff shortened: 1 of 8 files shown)


ReqX:
@greptile-apps

Thanks for the review. Both findings addressed:

**P2: Missing `await` on async function call** (preset-manager.test.ts:755): Fixed, added `async` to the test callback and `await` on the `handleCommandExecuteBefore` call.

**P2: Empty-preset guard bypassed when only reset updates exist** (preset-manager.ts:169-177): Fixed, guard now checks `agentUpdates` directly instead of `allUpdates`. An empty preset always shows the empty-preset message, regardless of pending resets from a previous preset.

Ready for re-review.

ReqX:
@greptile-apps — Thanks for the re-review.

**P1 (stale scalar fields not cleared):** Fixed in e904619.

The root cause: `mapOverrideToAgentConfig(baseline)` only emits fields the baseline defines, so `temperature`/`variant`/`options` from the previous preset leak when the baseline omits them.

The fix is in the config() hook reset block: instead of checking `'field' in baseline`, it now checks `'field' in prevOverride`. This uses the previous preset's keys to drive deletion. If the old preset set `temperature: 0.3` and the baseline doesn't define temperature, it still gets cleared.

The `preset-manager.ts` `resetUpdates` path intentionally keeps the baseline-driven approach (with a new comment explaining why). The config() hook is the authoritative cleanup point since it runs on every `client.config.update()` call.

Worth noting: before this PR, no reset logic existed at all. Stale fields always leaked across preset switches. This introduces cleanup that's now complete rather than partial.

Rdy for yet another review pass.

PR #442, opened 2026-05-04
fix: debounce tmux layout updates

dhaern:
## Summary
- Debounce tmux layout rebalancing after pane spawn/close bursts to avoid repeated `select-layout` / `set-window-option` churn.
- Keep explicit `applyLayout()` calls immediate while cancelling any pending debounced layout update.
- Check tmux layout command exit codes and log stderr instead of reporting failed layouts as applied.

## Notes
- `main-pane-width` / `main-pane-height` percentages were verified against tmux 3.6a, so this PR does not convert percentages to cells.
- This keeps pane creation/close behavior unchanged; only the cosmetic layout rebalance is coalesced.

## Validation
- `bun test src/multiplexer/tmux/index.test.ts src/multiplexer/factory.test.ts src/multiplexer/session-manager.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun test`
- `bun run lint`

### DIFF
--- a/src/multiplexer/tmux/index.test.ts
+++ b/src/multiplexer/tmux/index.test.ts
@@ -1,0 +1,148 @@
+import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test';
+
+type SpawnResult = {
+  exited: Promise<number>;
+  stdout: () => Promise<string>;
+  stderr: () => Promise<string>;
+  kill: () => boolean;
+  exitCode: number | null;
+  proc: never;
+};
+
+const logMock = mock(() => {});
+const crossSpawnMock = mock((_command: string[]) => createSpawnResult());
+
+mock.module('../../utils/logger', () => ({
+  log: logMock,
+}));
+
+mock.module('../../utils/compat', () => ({
+  crossSpawn: crossSpawnMock,
+}));
+
+let importCounter = 0;
+
+function createSpawnResult(
+  exitCode = 0,
+  stdout = '',
+  stderr = '',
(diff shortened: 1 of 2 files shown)


dhaern:
@greptile-apps Thanks for the review. All three findings addressed in the latest push:

1. **Tight timing margin** — increased test wait from 170ms to 300ms (2× debounce) to avoid flakiness under CI load.

2. **In-flight `applyLayoutNow` concurrency** — added a `layoutGeneration` counter. Both `applyLayout` and `scheduleLayout` increment it; stale debounced callbacks are discarded when the generation no longer matches, preventing concurrent layout runs.

3. **`args` log missing tmux binary** — `runTmux` now logs `[tmux, ...args]` so the full command is reproducible from logs.

dhaern:
@greptile-apps Good catches, both fixed in the latest push:

1. **Third test `wait(170)`** — bumped to `wait(300)` for consistency with the coalesce test.

2. **Dead post-await generation guard** — removed the stale `if (this.layoutGeneration !== gen) return;` and comment from `applyLayout`. The real protection is in `scheduleLayout` where the gen check gates the deferred `applyLayoutNow` call. The direct `applyLayout` path already cancels the timer at the top, so the post-await guard was indeed unreachable.

dhaern:
@greptile-apps Fair point. Replaced `const gen = ++this.layoutGeneration; ... void gen;` with a plain `this.layoutGeneration++;` — the side-effect is what matters, no binding needed.

PR #452, opened 2026-05-08
fix: sync alias agents to core config so delegation resolves correctly

dbc-hbin:
When opencode core registers a built-in agent that shares a name with a plugin alias (e.g. core's native "explore" and the plugin's "explorer"), the two coexist as separate agents. Delegation to the alias picks the core's built-in (which has no configured model) and falls back to the parent model.

Now, after merging plugin agents into opencodeConfig.agent, we synchronize each alias to its target: copy the target's config (including model/variant/temperature) and mark the alias hidden. This ensures:

- @explore resolves to the same model as @explorer
- Alias entries are hidden from the TUI sidebar
- User-configured alias models are respected (not overwritten)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -632,22 +632,38 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
       // target agents (e.g. "explorer") AFTER all model mutations so the
       // alias always reflects the latest target model/variant/temperature.
       // This also overrides any core built-in agents that share alias names.
+      // Fields are deleted from the alias when absent from the target so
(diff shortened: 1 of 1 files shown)


alvinunreal:
Thanks @hanbinnoh  - during installation we disable builtin agents.
Can you check if it's applied for your config?

```
    "agent": {
      "explore": {
        "disable": true
      },
      "general": {
        "disable": true
      }
    },
    ```

dbc-hbin:
OpenCode detected the issue and modified the config, but I thought there might be others like me who suddenly ended up using expensive models for explorer tasks and wasting tokens, so I updated the code.

PR #457, opened 2026-05-11
Add boomerang-style fork workers

alvinunreal:
## Summary
- Adds native Fork support with `/fork` and a `fork_session` tool that runs a child orchestrator worker and returns a `<fork_summary>` to the original session.
- Passes compact parent context, decisions, constraints, and file references into the forked orchestrator so it can do the requested work without re-discovering the thread.
- Adds restricted `read_session`, workspace-safe file context loading, nested-fork guards, depth checks, and fork descendant tracking.
- Documents the workflow in `docs/fork.md`, links it from README, and references it from `docs/tools.md`.

## Behavior
- `/fork <goal>` tells the main orchestrator to call `fork_session` with the user's requested worker goal and clearly relevant files.
- `fork_session` creates a real child session with `parentID` set to the caller, runs it as `orchestrator`, allows normal subagent delegation, extracts the result, then returns the summary to the parent session.
- Fork workers and their descendants cannot start another fork worker.
- `read_session` is only available from fork worker sessions and can only read that worker's source session.
- File context is constrained to the workspace realpath, skips binary files, blocks symlink escape, and caps large outputs.

## Review
- @oracle reviewed the implementation. The `/fork` command intentionally uses the concise user-facing name requested for this feature.

## Validation
- `bun run check:ci`
- `bun run typecheck`
- `bun test`
- `bun run build`


alvinunreal:
@greptileai

alvinunreal:
@greptileai

Issue #474, opened 2026-05-15
Should Orchestrator be the main coding agent?


alvinunreal:

alvinunreal:
So idea is we can change Orchestrator to just delegator, opencode added experimental background agents - Orchestrator can be purely for planning, delegating parallel tasks and managing the workflow.

Optionally enabling current executioner flow.

Vulcankta:
Now that experimental background agents are in OpenCode, a delegator-style Orchestrator makes sense.
Just an idea — following the existing `/auto-continue` pattern, a simple `/orch-mode` toggle (off by default = executioner, on = delegator) would let early adopters test it without disrupting current users while the upstream API stabilizes.

alvinunreal:
Beta is released: https://github.com/alvinunreal/oh-my-opencode-slim#v2-background-orchestration-beta

alvinunreal:
`/deepwork` added in beta version - creates a plan file, and uses oracle for reviews, does phased implementations

mhenke:
> This was generated by AI during triage.

Closing as already covered. The orchestrator-as-delegator idea shipped: background orchestration is in (see `docs/v2-background-orchestration.md`) and `/deepwork` landed in the v2 beta. If you still want a dedicated `/orch-mode` toggle, that's a separate ask, but the core question here is resolved.

PR #476, opened 2026-05-15
[WIP] - Implement background subagent scheduler

alvinunreal:
Ideas for v2

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,6 +490,7 @@ Use this section as a map: start with installation, then jump to features, confi
 | Doc | What it covers |
 |-----|----------------|
 | **[Council](docs/council.md)** | Run multiple models in parallel and synthesize a single answer with `@council` |
+| **[V2 Background Orchestration](docs/v2-background-orchestration.md)** | Scheduler-first orchestrator model built around native background subagents |
 | **[Multiplexer Integration](docs/multiplexer-integration.md)** | Watch agents work live in Tmux or Zellij panes |
 | **[Session Management](docs/session-management.md)** | Reuse recent child-agent sessions with short aliases instead of starting over |
 | **[Session Goal](docs/session-goal.md)** | Pin a session objective with `/goal` so todos, delegation, and verification stay aligned |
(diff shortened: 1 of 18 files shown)


dhaern:
I like the overall direction of this PR. Moving slim toward a scheduler-first model fits background subagents much better than the old assumption that a `task` result always means delegated work is already finished.

The strongest part is the new job-board concept: tracking background work separately from resumable sessions, distinguishing running jobs from terminal results, and requiring the orchestrator to reconcile completed work before finalizing. That is the right abstraction for this model.

A few fragile points I would tighten before calling this stable:

- **Injected-completion dedupe on task relaunches**

  If the same task ID is relaunched/reset and later produces the same terminal state and result summary, the old processed-completion signature can cause the new completion to be ignored. That could leave the job board showing a relaunched task as still running even though a terminal result arrived.

- **Task result parsing should be stricter**

  Since parsed task output now feeds scheduler state, the parser should avoid accepting malformed or mismatched result/error blocks. This is small, but important because bad parsing here can create misleading job-board state.

- **Unused lifecycle state**

  The job state model includes `launched`, but the implementation appears to move jobs directly into `running`. If there is no actual lifecycle transition using `launched`, it makes the state model look more complex than it really is.

- **Idle-based reconciliation is a heuristic**

  Reconciliation through the parent returning to idle is a reasonable first approach. It ensures the orchestrator had a chance to see the terminal result, but it does not prove the result was actually considered. That is acceptable for an initial implementation, but the docs should avoid presenting it as a strict semantic guarantee.

- **Synthetic completion ingestion should stay conservative**

  Requiring synthetic message parts is the right protection. Since this depends on formatted text rather than a structured event, I would keep the accepted shape narrow and avoid broad matching that could accidentally ingest unrelated content.

- **Keep the scheduler model strong, but define the direct-work escape hatch clearly**

  The scheduler-first prompt is the right direction, especially for multi-step work like the issue #474 example. I would just keep the direct-work escape hatch explicit for trivial/mechanical actions where scheduling would add more overhead than value.

Overall, I think this is a strong foundation for V2. The architecture is good, the job-board abstraction is the right place for this state, and the launch/status/reconcile split is exactly the distinction slim needs. I would mainly tighten the parser and lifecycle edge cases so the board cannot drift into stale or misleading state.

PR #481, opened 2026-05-15
[BETA] - Version 2

alvinunreal:
Ideas for v2

v2.0 priorities:
- background agent manager, session reuse
- multitasking, non blocking orchestrator
- deepwork, planning with files
- prompt tunings
- oh-my-opencode-slim for self improving, self tuning

Before release

- [x] Fix session resuse, doesn't work now
- [x] remove task_status
- [x] clean prompt for background tasks - opencode include most already


alvinunreal:
@greptileai

alvinunreal:
@greptileai

dhaern:
This is noticeably stronger than #476. The overall V2 direction feels much clearer now, and the job-board model looks like a much better foundation for background orchestration.

The remaining points I would still keep explicit are mostly around edge-case semantics:

- **Unregistered synthetic completions**

  If a synthetic completion is seen for a task that is not currently registered in the background job board, it looks like that occurrence may be parsed again on later message transforms instead of being marked as already inspected.
  That does not seem to corrupt scheduler state, but it can create repeated unnecessary work and makes this path slightly noisier than it needs to be.

- **Meaning of `reconciled`**

  The current flow appears to treat a terminal job as reconciled once it was injected into an orchestrator turn and the parent session later returns to idle.

  That is practical for a beta, but semantically it means:

  > reconciled = the result was surfaced to the orchestrator during a completed turn

  not necessarily:

  > reconciled = the result was explicitly consumed or acted on

  That distinction is fine if intentional, but I would keep the wording and docs precise around it. Otherwise a terminal result could disappear from the board after an unrelated orchestrator response simply because the session returned to idle.

Overall, this is a much better WIP/beta foundation than the first PR. The architecture looks solid; I would mainly keep these two remaining edge cases clear before promoting the model beyond beta.

dhaern:
Im using and testing this locally on `2.0.0-beta.7` with `explorer` configured to use `gemini-3.5-flash-preview`. The BackgroundJobBoard changes in this PR look good, but I think there are still two separate lifecycle problems below the Job Board layer.

### 1. Pane timeout is still independent from background job state

In my run, two background explorer smoke tasks were launched and both tmux panes spawned correctly:

- `ses_1b96400cbffe2NeOXMAvD97wvv` / `Smoke explorer tools` / pane `%1`
- `ses_1b963eff9ffesDEBBA5f3FXzNU` / `Smoke explorer grep` / pane `%2`

The first pane closed after an idle event, which is expected if the task finished.

The second pane was killed by slim exactly around 10 minutes after creation:

```text
[multiplexer-session-manager] child session created, spawning pane {"sessionId":"ses_1b963eff9ffesDEBBA5f3FXzNU", ...}
...
[multiplexer-session-manager] closing session pane {"sessionId":"ses_1b963eff9ffesDEBBA5f3FXzNU","paneId":"%2","reason":"timeout"}
```

This maps directly to the hardcoded timeout in `src/multiplexer/session-manager.ts`:

```ts
const SESSION_TIMEOUT_MS = 10 * 60 * 1000;
```

And in `pollSessions()`:

```ts
const isTimedOut = now - tracked.createdAt > SESSION_TIMEOUT_MS;

if (isIdle || missingTooLong || isTimedOut) {
  sessionsToClose.push({
    sessionId,
    reason: isIdle ? "idle" : isTimedOut ? "timeout" : "missing",
  });
}
```

So the visible failure is not necessarily Gemini/streaming-related. The pane lifecycle is still governed by multiplexer heuristics (`session.idle`, `/session/status`, missing status, fixed timeout), not by the native background task lifecycle or the BackgroundJobBoard state.

That means a legitimate long-running background job, or a job whose status is not observed correctly, can have its pane killed even though the actual task may still be running or unreconciled.

### 2. `/session/status` polling can fail on partial or empty JSON

The same run also logged this:

```text
[multiplexer-session-manager] poll error {"error":"SyntaxError: Unexpected end of JSON input"}
```

That appears to come from `fetchSessionStatuses()`:

```ts
const response = await fetch(url, { signal: AbortSignal.timeout(2_000) });
...
return (await response.json()) as Record<string, { type: string }>;
```

So if `/session/status` returns an empty, truncated, or otherwise non-JSON body during a race condition — for example while sessions are being created/closed, the server is busy, or an attach/pane is racing startup — the poll loop throws and only logs the error.

This does not directly close the pane by itself because `pollSessions()` catches the error and returns, but it makes status polling unreliable. Combined with the fixed 10-minute pane timeout, it can leave the manager with stale or incomplete lifecycle information until the timeout closes the pane.

### Why this matters for V2

#481 improves deduplication and reconciliation at the BackgroundJobBoard level, but the multiplexer pane lifecycle is still separate from that state. The pane manager can still close panes based on local heuristics even when the background job has not been reconciled.

I would treat this as a follow-up rather than expanding this PR: pane lifetime should probably be aligned with the native background job state / `task_status` / BackgroundJobBoard, or at least the fixed timeout should be configurable and not be treated as the authoritative task lifecycle signal.

alvinunreal:
@greptileai

omar-khallaf:
Hi, @alvinunreal. I think without #515, beta can't be installed since the CLI always defaults to the `@latest` version.

alvinunreal:
@omar-mohamed-khallaf released new beta

stephanschielke:
_Just as a random motivational booster: I ditched all my harnesses for this beta version of omo-slim and don't look back or miss anything. Great work and IMHO all in the right direction. So thank you for the great work!_

xenstar:
so this is beta version right, not the stable version? I think it would be good if launching beta version, then it's in the pre-realease.

alvinunreal:
it's released, no more beta @xenstar

xenstar:
I think it's time for setting up a community on discord or telegram group.

alvinunreal:
@xenstar we have telegram

PR #484, opened 2026-05-16
detect 'Insufficient balance' as rate-limit error for fallback

zackslash:
## Summary
- Extend `isRateLimitError()` to match "Insufficient balance." errors returned by providers such as OpenCode Zen when an account is depleted
- Add missing `insufficient` keyword to the `session.status` retry handler's keyword list
- This ensures that `ForegroundFallbackManager` triggers fallback to the next model in the chain instead of silently failing

## Problem
1. The existing `RATE_LIMIT_PATTERNS` regex `/insufficient.?quota/i` does not match the error message `"Insufficient balance."` — only `"insufficient quota"`. Providers like OpenCode Zen return this exact message when a subscription runs out, causing the fallback system to never activate.
2. The `session.status` retry handler maintains its own keyword list that was missing `insufficient` entirely, so even if the regex were fixed, "Insufficient balance." arriving via that path would still not trigger fallback.

## Fix
- Changed `/insufficient.?quota/i` → `/insufficient.?(quota|balance)/i` in `RATE_LIMIT_PATTERNS`
- Added `msg.includes('insufficient')` to the `session.status` keyword list
- Added a test case confirming `"Insufficient balance."` is detected as a rate-limit error

## Testing
- Added test: `returns true for "Insufficient balance."` in `isRateLimitError` test suite

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -85,6 +85,10 @@ describe('isRateLimitError', () => {
     expect(isRateLimitError({ message: 'overloaded_error' })).toBe(true);
   });
 
+  test('returns true for "Insufficient balance."', () => {
+    expect(isRateLimitError({ message: 'Insufficient balance.' })).toBe(true);
+  });
+
   test('returns false for non-rate-limit error', () => {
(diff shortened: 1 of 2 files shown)


zackslash:
Addressed session.status handler too

PR #486, opened 2026-05-17
fix(job-board): distinguish resumed sessions from new launches in age label

stevekstevek:
## What changed

Added `lastLaunchedAt: number` field to `BackgroundJobRecord`.

- **New session** (first time a taskID is registered): `lastLaunchedAt = launchedAt = now`
- **Session reuse** (same taskID relaunched): `lastLaunchedAt = now`, `launchedAt` preserved from original launch

`formatJob` now uses `lastLaunchedAt` for age and compares it to `launchedAt` to select the right label:
- `running [just launched, Xs ago]` — new session, dispatched < 30s ago
- `running [resumed, Xs ago]` — existing session reused, resumed < 30s ago
- `running` — no annotation if > 30s (unchanged)

## Why

The v2 beta job board already annotates freshly dispatched tasks with `[just launched]`, which is a big help. But session reuse shows the same label — an orchestrator that dispatches `task(..., task_id=existing)` can't tell whether it just dispatched a new task or resumed a pre-existing one.

The `resumed` label closes that gap: after calling `task(task_id=X)` for an existing session, the board immediately shows `running [resumed, Xs ago]`, making the intent unambiguous.

## Tests

One new test: `'annotates resumed running jobs with resumed label in the prompt'`
One updated test: `'resets terminal state when an existing task id is relaunched'` (asserts `lastLaunchedAt`)

All 14 tests pass.

### DIFF
--- a/src/utils/background-job-board.test.ts
+++ b/src/utils/background-job-board.test.ts
@@ -165,6 +165,7 @@ describe('BackgroundJobBoard', () => {
       terminalUnreconciled: false,
       completedAt: undefined,
       resultSummary: undefined,
+      lastLaunchedAt: 300,
       updatedAt: 300,
     });
   });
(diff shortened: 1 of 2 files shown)


alvinunreal:
Thanks for the PR — the launched vs resumed distinction is useful for the V2 job board. I’m going to merge it and apply the small follow-up fixes noted in review (formatting, Math.floor at the 30s boundary, and the launchedAt invariant assertion).

PR #489, opened 2026-05-18
fix(preset): resolve /preset race condition and council default hint

### LINKED ISSUE #438: [Bug]: /preset not working (at runtime)

On my machine the `/preset` command doesn't work. The preset seems to switch, however, the agents are no longer available in the selection.

### Steps to reproduce

Normal `/preset <id>` invocation.

### oh-my-opencode.json

```jsonc
{
  "preset": "opencode-go",
  "presets": {
    "openai": {
      "orchestrator": {
        "model": "openai/gpt-5.5",
        "skills": ["*"],
        "mcps": ["*", "!context7"]
      },
      "oracle": {
        "model": "openai/gpt-5.5",
        "variant": "high",
        "skills": ["simplify"],
        "mcps": []
      },
      "librarian": {
        "model": "openai/gpt-5.4-mini",
        "variant": "low",
        "skills": [],
        "mcps": ["websearch", "context7", "grep_app"]
      },
      "explorer": {
        "model": "openai/gpt-5.4-mini",
        "variant": "low",
        "skills": [],
        "mcps": []
      },
      "designer": {
        "model": "openai/gpt-5.4-mini",
        "variant": "medium",
        "skills": ["agent-browser"],
        "mcps": []
      },
      "fixer": {
        "model": "openai/gpt-5.4-mini",
        "variant": "low",
        "skills": [],
        "mcps": []
      }
    },
    "opencode-go": {
      "orchestrator": {
        "model": "opencode-go/glm-5.1",
        "skills": ["*"],
        "mcps": ["*", "!context7"]
      },
      "oracle": {
        "variant": "max",
        "skills": ["simplify"],
        "mcps": []
      },
      "council": {
        "variant": "high",
        "skills": [],
        "mcps": []
      },
      "librarian": {
        "model": "opencode-go/minimax-m2.7",
        "skills": [],
        "mcps": ["websearch", "context7", "grep_app"]
      },
      "explorer": {
        "model": "opencode-go/minimax-m2.7",
        "skills": [],
        "mcps": []
      },
      "designer": {
        "model": "opencode-go/kimi-k2.6",
        "variant": "medium",
        "skills": ["agent-browser"],
        "mcps": []
      },
      "fixer": {
        "variant": "high",
        "skills": [],
        "mcps": []
      }
    }
  }
}
```

### OpenCode version

1.14.33

latest

### Operating system

macOS 26.4.1

Image

Image

### LINKED ISSUE #479: Council agent may force default preset despite council.default_preset
## Summary

When `council.default_preset` is configured to a non-`default` preset, the council manager appears designed to use it when no preset argument is supplied. However, the council agent prompt/tool description still tells the agent that the default is `"default"`, and in practice the council agent can end up running the `default` council preset instead of the configured `council.default_preset`.

## Environment

- Package: `oh-my-opencode-slim`
- Version observed: `1.1.1`
- OpenCode usage: local plugin wrapper importing from `../node_modules/oh-my-opencode-slim/dist/index.js`

## Config excerpt

```jsonc
{
  "council": {
    "default_preset": "relaxed",
    "presets": {
      "default": {
        "alpha": { "model": "nano-gpt/zai-org/glm-5.1" },
        "beta": { "model": "opencode-go/kimi-k2.6" },
        "gamma": { "model": "opencode-go/qwen3.6-plus" }
      },
      "relaxed": {
        "alpha": { "model": "nano-gpt/xiaomi/mimo-v2.5-pro" },
        "beta": { "model": "nano-gpt/deepseek/deepseek-v4-pro-cheaper:thinking" },
        "gamma": { "model": "opencode-go/qwen3.6-plus" },
        "delta": { "model": "nano-gpt/moonshotai/kimi-k2.6:thinking" }
      }
    },
    "councillor_retries": 3
  }
}
```

## Observed behavior

A council test used the `default` preset instead of `relaxed`:

```text
alpha = glm-5.1
beta  = kimi-k2.6
gamma = qwen3.6-plus
```

That matches the configured `council.presets.default`, not `council.presets.relaxed`.

## Expected behavior

If `council.default_preset` is set to `"relaxed"`, invoking the council without an explicit preset should run the `relaxed` preset:

```text
alpha = mimo-v2.5-pro
beta  = deepseek-v4-pro-cheaper:thinking
gamma = qwen3.6-plus
delta = kimi-k2.6:thinking
```

## Relevant implementation detail

The manager logic seems correct:

```js
const resolvedPreset = presetName ?? councilConfig.default_preset ?? "default";
```

But the council agent prompt says:

```text
Optionally specify a preset (default: "default")
```

And the tool schema says:

```js
preset: z.string().optional().describe('Council preset to use (default: "default"). Must match a preset in the council config.')
```

This can encourage the council agent to explicitly pass `preset: "default"`, which bypasses `council.default_preset` because `presetName` is then no longer undefined.

## Suggested fix

Update the council agent prompt and tool description to make the runtime behavior explicit, e.g.:

```text
Optionally specify a preset. If omitted, uses council.default_preset from config, falling back to "default".
```

And ideally instruct the council agent not to pass `preset: "default"` unless the user explicitly requested the default preset.

## Workaround

Duplicate the desired preset contents into `council.presets.default`, or explicitly tell the council agent to use the desired preset in the user prompt.

### LINKED ISSUE #488: [Bug]: `/preset` command causes 'Agent not found' error due to race condition on instance reload
## Description

Running `/preset <name>` fails with:

```
Agent not found: "orchestrator". Available agents: build, plan
```

## Root Cause Analysis

The `switchPreset()` function (in `src/tools/preset-manager.ts`) calls `ctx.client.config.update({ body: { agent: allUpdates } })` which:

1. Writes agent model overrides to `<workspace>/config.json` (via OpenCode's `configSvc.update` which writes to `InstanceState.directory + "/config.json"`)
2. Calls `markInstanceForDisposal()` which triggers an instance reload

On reload:
- OpenCode reads `config.json` which now contains agent entries like `orchestrator`, `oracle`, `explorer`, etc.
- These agents are plugin-defined and only exist after the plugin's `config` hook runs
- But the config is read **before** plugins register their agents
- The `activeRuntimePreset` in-memory variable (from `src/config/runtime-preset.ts`) resets to `null` on reload, so the `config` hook's preset application code at line 33662 doesn't run
- Result: OpenCode tries to use `orchestrator` as default agent, but only built-in agents (`build`, `plan`) are registered → error

## Why the preset is lost on reload

`activeRuntimePreset` is a module-level `var` initialized to `null`. When `switchPreset()` calls `setActiveRuntimePresetWithPrevious(presetName)`, it sets the in-memory value. But after `markInstanceForDisposal()` triggers a reload, the plugin reinitializes and `activeRuntimePreset` goes back to `null`.

The plugin init code (around line 33450) does check `getActiveRuntimePreset()` and would re-apply the preset if it were set, but since the value was lost on reload, it falls through.

## Current flow

```
/preset personal
  → switchPreset("personal")
    → setActiveRuntimePresetWithPrevious("personal")  // in-memory only
    → ctx.client.config.update({ body: { agent: allUpdates } })
      → writes <workspace>/config.json with { agent: { orchestrator: {...}, oracle: {...}, ... } }
      → markInstanceForDisposal()
        → instance reloads
          → activeRuntimePreset = null (module reinitialized)
          → reads config.json with plugin-agent entries
          → plugins not loaded yet → "Agent not found: orchestrator"
```

## Suggested Fix

Instead of using `config.update()` (which persists to disk and triggers a reload), the preset switch should:

1. **Persist the active preset name** to a file that survives reload — e.g., the existing TUI state file (`~/.local/share/opencode/storage/oh-my-opencode-slim/tui-state.json`) or the plugin's own config
2. **Read the persisted preset name** during plugin initialization (around line 33450), before `loadPluginConfig()` or right after
3. **Remove the `config.update()` call** from `switchPreset()` since the `config` hook already applies preset overrides when `activeRuntimePreset` is set
4. Instead, after persisting the preset name, call `markInstanceForDisposal()` directly (if available) or use a different reload mechanism that ensures the plugin's `config` hook picks up the persisted preset

Alternatively, the simplest fix might be to:
- Persist the preset name in the TUI state file
- On init, read it back and call `setActiveRuntimePreset()` before the existing code at line 33450 checks `getActiveRuntimePreset()`
- This way the existing reload-time preset application code works correctly

## Workaround

Change `"preset": "work"` to `"preset": "personal"` in `oh-my-opencode-slim.json`. This applies the preset at config load time (line 18787: `config.agents = deepMerge(preset, config.agents)`) without going through `config.update()`. However, the `/preset` command at runtime still produces the error.

Also, after hitting this bug, manually delete the stale `config.json` file that gets created in the workspace root directory.

## Environment

- oh-my-opencode-slim version: 1.1.1
- OpenCode version: 1.15.4
- OS: macOS

ReqX:
## Summary

Fixes three preset-related bugs in one PR:

### #488 / #438 — `/preset` race condition (same root cause)

`switchPreset()` pushed the success response to `output.parts` **after** calling `client.config.update()`, which triggers `Instance.dispose()` → plugin re-init. The in-flight response was lost because the instance was disposed before the output could be processed, causing "Agent not found" errors.

**Fix**: Reorder `switchPreset()` to emit the success response **before** calling `config.update()`. This ensures the user sees the confirmation before the disposal cycle runs. On `config.update()` failure, a follow-up warning is emitted and runtime preset state is rolled back.

### #479 — Council `default_preset` bypass

The council tool description and agent prompt hardcoded `(default: "default")`, which misled the LLM into explicitly passing `preset: "default"` instead of omitting the parameter. When omitted, the runtime correctly falls back to `council.default_preset` → `"default"`, but the explicit `"default"` bypassed the user's configured default.

**Fix**: Update description to indicate omitting uses `council.default_preset` from config.

## Changes

| File | Change |
|------|--------|
| `src/tools/preset-manager.ts` | Reorder `switchPreset()` — emit response before `config.update()` |
| `src/tools/preset-manager.test.ts` | Update test expectations for new error message wording |
| `src/tools/council.ts` | Remove misleading `(default: "default")` from tool param description |
| `src/agents/council.ts` | Remove misleading `(default: "default")` from agent prompt |

## Validation

- `bun run check:ci` - passes clean
- `bun test -t "preset"` - 76 pass, 0 regressions
- Pre-existing typecheck errors (missing SDK modules) unchanged

Closes #488, closes #438, closes #479

### DIFF
--- a/src/agents/council.ts
+++ b/src/agents/council.ts
@@ -17,7 +17,7 @@ orchestration system that runs consensus across multiple models.
 
 **Usage**:
 1. Call the \`council_session\` tool with the user's prompt
-2. Optionally specify a preset (default: "default")
+2. Optionally specify a preset (if omitted, uses council.default_preset from config)
 3. Receive the councillor responses formatted for synthesis
 4. Follow the Synthesis Process below
 5. Present the result to the user
--- a/src/tools/council.ts
+++ b/src/tools/council.ts
@@ -46,7 +46,7 @@ Returns the councillor responses with a summary footer.`,
         .string()
         .optional()
         .describe(
-          'Council preset to use (default: "default"). Must match a preset in the council config.',
+          'Council preset to use. If omitted, uses council.default_preset from config (falls back to "default"). Must match a preset in the council config.',
         ),
     },
     async execute(args, toolContext) {
--- a/src/tools/preset-manager.test.ts
+++ b/src/tools/preset-manager.test.ts
@@ -319,7 +319,9 @@ describe('createPresetManager', () => {
       );
 
       const text = getOutputText(output);
-      expect(text).toContain('Failed to switch preset');
+      expect(text).toContain(
+        'Warning: preset "cheap" may not have fully applied',
+      );
       expect(text).toContain('Server unavailable');
       expect(readTuiSnapshot().agentModels).toEqual({
         explorer: 'openai/gpt-5.4-mini',
@@ -794,7 +796,9 @@ describe('createPresetManager', () => {
 
       // Active preset should still be "cheap" after error
       expect(getActiveRuntimePreset()).toBe('cheap');
-      expect(getOutputText(output2)).toContain('Failed to switch preset');
+      expect(getOutputText(output2)).toContain(
+        'Warning: preset "expensive" may not have fully applied',
+      );
 
       // Cleanup
       setActiveRuntimePreset(null);
--- a/src/tools/preset-manager.ts
+++ b/src/tools/preset-manager.ts
@@ -186,6 +186,30 @@ export function createPresetManager(ctx: PluginInput, config: PluginConfig) {
     const previousPreset = activePreset;
     setActiveRuntimePresetWithPrevious(presetName);
 
+    // Build summary and emit response BEFORE config.update() to avoid
+    // race condition where instance disposal kills the in-flight response.
+    // See GitHub issues #488 and #438.
+    const summaryParts: string[] = [];
+    for (const [name, cfg] of Object.entries(agentUpdates)) {
+      const parts: string[] = [name];
+      if (cfg.model) parts.push(`model: ${cfg.model}`);
+      if (cfg.variant) parts.push(`variant: ${cfg.variant}`);
+      if (cfg.temperature !== undefined) parts.push(`temp: ${cfg.temperature}`);
+      if (cfg.options) parts.push('options: yes');
+      summaryParts.push(parts.join(' → '));
+    }
+    if (Object.keys(resetUpdates).length > 0) {
+      summaryParts.push(
+        `Reset to baseline: ${Object.keys(resetUpdates).join(', ')}`,
+      );
+    }
+
+    output.parts.push(
+      createInternalAgentTextPart(
+        `Switched to preset "${presetName}":\n${summaryParts.join('\n')}`,
+      ),
+    );
+
     try {
       await ctx.client.config.update({
         body: { agent: allUpdates },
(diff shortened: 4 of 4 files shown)


ReqX:
Closing - agent filed prematurely. Will re-submit after proper review.

PR #491, opened 2026-05-19
fix(multiplexer): skip missing cleanup for sessions never observed in status

andrewylies:
## Summary
- Track whether a multiplexer child session has appeared in `/session/status`.
- Skip `missing` cleanup for sessions that were never observed in `/session/status`.
- Keep the existing `missing` fallback for previously observed sessions, with idle/deleted/timeout cleanup unchanged.
- Add regression coverage for never-seen and previously-seen status cases.

Nested council child sessions can emit lifecycle events and spawn panes without appearing in `/session/status`. Before this fix, the polling fallback could close those panes as `missing`, then later `busy` events would respawn them and cause a pane flicker loop.

This is intentionally narrower than removing `missing` cleanup entirely: sessions that were previously observed in `/session/status` still use the existing fallback cleanup path.

## Validation
- `bun test src/multiplexer/session-manager.test.ts`
- `bun run typecheck`
- `bun run check:ci`
- `bun test`

### DIFF
--- a/src/multiplexer/session-manager.test.ts
+++ b/src/multiplexer/session-manager.test.ts
@@ -390,6 +390,61 @@ describe('MultiplexerSessionManager', () => {
       expect(mockMultiplexer.closePane).not.toHaveBeenCalled();
     });
 
+    test('does not close missing session that was never seen in status', async () => {
+      const ctx = createMockContext();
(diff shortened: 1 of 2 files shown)


alvinunreal:
Thanks @andrewylies

PR #496, opened 2026-05-22
Fix Node ESM plugin packaging

xinxingi:
## Summary

This fixes Node ESM packaging issues in the published plugin artifact.

### Changes
- switch runtime imports from `@opencode-ai/plugin/tool` to `@opencode-ai/plugin`
- externalize `@opencode-ai/plugin/*` and `@opencode-ai/sdk/*` in build scripts
- add a release verification guard to fail if `vscode-jsonrpc/node` leaks into built artifacts

## Why

The published npm package can fail to load in OpenCode desktop because the built artifact may retain an import path that resolves to `vscode-jsonrpc/node` through host-side dependencies. This change keeps those runtime imports external and adds a regression check.

## Validation

Validated locally by code inspection and branch-local commit creation.
Full build/release verification was not run in my environment because `bun` was not available in the repo environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,8 +47,8 @@
     "LICENSE"
   ],
   "scripts": {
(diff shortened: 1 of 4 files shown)


alvinunreal:
Thanks @xinxingi

PR #503, opened 2026-05-24
docs(spec): trace_regenerate strict heading format

boin:
## What

Document the strict heading format required by `trace_regenerate`:
- `## REQ-N:` / `## DES-N:` (level-2, trailing colon required)
- Multi-REQ anchors must be comma-separated on a single `Rationale anchor:` line

Headings like `## REQ-1 多平台插件库` (no colon) are silently ignored and produce an empty trace table.

## Why

User report: `trace_regenerate` ran successfully but produced a header-only table. Root cause is `HEADING_ID_RE = /^##\\s+(REQ-\\d+):/` — the colon is mandatory but undocumented at invoke site.

## Changes

- `src/skills/grill/SKILL.md` — new 'Heading format is strict' subsection
- `src/tools/trace/index.ts` — tool description now states the format constraint and the failure mode
- No parser change (strict format kept by design, per user preference)

Second commit is `bun run check` auto-fix (import order + formatting) — no behavior change.

## Verify

- `bun run check:ci` ✓
- `bun run typecheck` ✓
- `bun test src/tools/trace src/tools/spec` ✓ (24/24)
- Full suite: 1 pre-existing failure in `src/hooks/apply-patch/hook.test.ts` unrelated


boin:
Wrong target — meant to open against fork's own master, not upstream. Reopening on the fork.

PR #522, opened 2026-06-06
fix: respect OPENCODE_CONFIG_DIR during install

flyinghail:
## Summary
- Make install-time OpenCode config writes honor `OPENCODE_CONFIG_DIR`
- Add regression coverage for `opencode.json` / `opencode.jsonc` path resolution
- Document custom OpenCode config directorybehavior
 
## Problem
The installer modified the default OpenCode global config directory even when `OPENCODE_CONFIG_DIR` was set. Problems occur when attempting to isolate the Agent stack via `OPENCODE_CONFIG_DIR`, as plugin registration details, disabled default agents, and LSP settings are saved in the global configuration, impacting all isolated profile configurations.

## Fix
`getOpenCodeConfigPaths()` now reuses `getConfigDir()`, which already applies the intended precedence:
1. `OPENCODE_CONFIG_DIR`
2. `XDG_CONFIG_HOME/opencode`
3. `~/.config/opencode`

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ Then:
    opencode models --refresh
    ```
 3. **Open your plugin config** at `~/.config/opencode/oh-my-opencode-slim.json`
+   or `$OPENCODE_CONFIG_DIR/oh-my-opencode-slim.json` if you use a custom
+   OpenCode config directory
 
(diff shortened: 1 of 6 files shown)


alvinunreal:
thanks @flyinghail

PR #547, opened 2026-06-14
fix: keep companion positioned on niri resize

smatheusblu:
## Summary

Fixes desktop Companion positioning when the window is configured for a corner and then changes size, especially on niri.

What changed:

- recompute companion geometry when position, size, grid dimensions, screen bounds, or selected session changes;
- reload `companion.size`/`companion.position` from state only when the config tuple actually changes, so ordinary agent/session updates do not reset the right-click size picker;
- keep the cross-platform egui/winit `InnerSize` + `OuterPosition` path;
- add a narrow niri-only best-effort IPC fallback when `NIRI_SOCKET` is present;
- dynamically read niri logical output bounds from `niri msg --json outputs` (no monitor name or screen-size hardcoding);
- move only the current companion process window with stable app-id/title using `move-floating-window --id`, without focusing it;
- add Rust unit coverage for geometry, config tuple behavior, niri JSON parsing, output selection, delta computation, and command args.

## Why

On niri, the native Companion can be floating but still ignore the egui/winit `OuterPosition` request. This caused two visible bugs:

- `companion.position: "top-left"` could still leave the Companion at the previous bottom-right location;
- when multiple agents made the bottom-right Companion grow, niri kept the old top-left coordinate, so the larger window extended off-screen.

The app-side geometry fix is still useful for compositors that honor `OuterPosition`. The niri IPC fallback is a bounded compositor-specific workaround for niri's positioning behavior.

## niri behavior

The fallback is intentionally narrow and best-effort:

- gated by `NIRI_SOCKET`;
- uses `std::process::Command`, not a shell;
- runs off the egui update thread;
- uses bounded delayed retries to wait for async resize/position settling;
- no-ops silently if niri is unavailable, JSON changes, the matching window is missing, or the window is not floating;
- does not focus or activate any window.

## Validation

Automated checks:

- [x] `cd companion && cargo fmt --all --check && cargo check && cargo test && cargo build --release` — 31 Rust tests passed
- [x] `bun run check:ci`
- [x] `bun run typecheck`
- [x] `bun test` — one transient timeout in `system > isOpenCodeInstalled detects opencode in ~/.opencode/bin`
- [x] `bun test -t "isOpenCodeInstalled detects opencode in ~/.opencode/bin"` — targeted rerun passed
- [x] `bun run build`

Manual/runtime validation on niri with the PR-built companion binary:

- top-left config:
  - JSON: `/tmp/opencode/companion-top-left-window-v5.json`
  - screenshot: `/tmp/opencode/companion-top-left-v5.png`
  - observed `tile_pos_in_workspace_view: [10.0, 10.0]`
  - observed `window_size: [120, 120]`
- bottom-right with five active agents:
  - JSON: `/tmp/opencode/companion-bottom-right-window-v5.json`
  - screenshot: `/tmp/opencode/companion-bottom-right-v5.png`
  - observed `tile_pos_in_workspace_view: [2190.0, 830.0]`
  - observed `window_size: [360, 240]`
  - on the current 2560×1080 output, the window is fully visible with the expected 10px gap
- cleanup:
  - JSON: `/tmp/opencode/companion-after-kill-window-v5.json`
  - no companion window remained after process termination

### DIFF
--- a/companion/src/app.rs
+++ b/companion/src/app.rs
@@ -1,11 +1,16 @@
 use std::sync::mpsc::Receiver;
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc,
+};
 use std::time::Duration;
 
 use eframe::egui;
 
 use crate::gifs::Gifs;
+use crate::niri;
 use crate::screen::primary_size;
-use crate::state::{read_state, start_watcher, SessionInfo};
+use crate::state::{read_state, start_watcher, CompanionConfigState, SessionInfo};
 
 const DEFAULT_SIZE: f32 = 120.0;
 const GAP: f32 = 10.0;
@@ -21,6 +26,23 @@ const SIZE_KEY: &str = "companion_size";
 const MENU_OPEN_KEY: &str = "companion_menu_open";
 const MENU_POS_KEY: &str = "companion_menu_pos";
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct WindowGeometryKey {
+    session_id: String,
+    position: String,
+    size_px: u32,
+    cols: u32,
+    rows: u32,
+    screen_w: u32,
+    screen_h: u32,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct ConfigKey {
+    position: String,
+    size: String,
+}
+
 fn grid_cols(n: usize) -> usize {
     match n {
         0 | 1 => 1,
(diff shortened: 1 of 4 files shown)


smatheusblu:
Addressed the Greptile retry-loop concern in 47c54d3: after a successful niri move, the retry worker now returns instead of continuing, so it cannot apply the same stale delta again.

Also fixed the failing CI test. The failure was in task-session-manager context compaction, not companion runtime code: capped context ordering depended on read timestamps, so CI could select the later 5-line file instead of the larger 12-line file. Context files now sort deterministically by line count, then recency, then path while preserving the (+N more) cap behavior.

Validation on the pushed commit:
- cargo fmt --all --check && cargo check && cargo test && cargo build --release (companion): passed, 31 tests
- bun run lint && bun run typecheck && bun run build && bun test: passed, 1142 tests

alvinunreal:
thanks @smatheusblu

PR #552, opened 2026-06-15
ci: cross-compile macos-x64 companion on macos-latest runner

bobbyunknown:
### Problem
GitHub-hosted Intel macOS runners (`macos-13` / `macos-12`) have extremely restricted queue times or are deprecated, causing the `macos-x64` build job to get stuck indefinitely waiting for a runner.

### Solution
Since the newer standard GitHub `macos-latest` (Apple Silicon M1/arm64) runners are instantly available and natively support cross-compilation to Intel targets through the Xcode SDK, we can compile the `x86_64` binary on an arm64 host:
* Changed `runs-on` from `macos-13` to `macos-latest` for the `macos-x64` job.
* Maintained the cargo target argument `--target x86_64-apple-darwin`.

### Verification
Successfully tested on a fork: the runner starts immediately and generates a valid `Mach-O 64-bit executable x86_64` companion binary without SDK or linker errors.

    
    
Screenshot 2026-06-15 at 15-31-37
Screenshot 2026-06-15 at 15-32-13

### DIFF
--- a/.github/workflows/companion-release.yml
+++ b/.github/workflows/companion-release.yml
@@ -9,7 +9,7 @@ on:
         type: string
       targets:
         description: >-
-          Comma-separated targets to build: macos-arm64,linux-x64,linux-arm64,windows-x64
+          Comma-separated targets to build: macos-arm64,macos-x64,linux-x64,linux-arm64,windows-x64
         required: true
         default: macos-arm64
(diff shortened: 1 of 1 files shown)


alvinunreal:
thanks @bobbyunknown

PR #555, opened 2026-06-15
Smooth companion GIF loops

gustavocaiano:
## Summary
- Purely styling-only GIF asset change for the Companion animations.
- Rewrites the Companion GIF loops to ping-pong (play forward, then backward) so the video does not visibly cut halfway through the loop and feels smoother visually.
- Keeps the PR limited to `companion/gifs/*.gif`.

## Validation
- Verified each Companion GIF has 70 frames, `loop=0`, and forward + reverse-tail frame order.


alvinunreal:
thanks @gustavocaiano pretty good

PR #557, opened 2026-06-15
fix(smartfetch): retry session cleanup to prevent orphaned conversations

dhaern:
## Problem

The smartfetch secondary-model feature creates a temporary session (`title: "smartfetch-secondary"`) to invoke a cheap model on fetched content, then deletes it in a `finally` block:

```typescript
} finally {
  await client.session
    .delete({ path: { id: sessionId }, query: { directory } })
    .catch(() => undefined);  // ← swallows ALL errors silently
}
```

The `.catch(() => undefined)` silently discards every delete error. When the delete fails — most commonly during an OpenCode instance dispose/reload cycle — the session is left orphaned in the database and shows up as an empty conversation in the UI.

This is a **race condition**: if smartfetch creates and cleans up a session while OpenCode is reloading instances, the delete targets an instance that is being torn down and fails. The silent catch hides it completely — no log, no retry, no way to know.

## Fix

Replace the silent catch with `deleteSessionSafely()`:

- **3 retry attempts** with 500ms delay between each — the dispose/reload window is transient (milliseconds), so a retry almost always succeeds on the next attempt.
- **`console.warn`** if all retries fail, so the issue is visible in logs instead of silently leaking sessions.

```typescript
async function deleteSessionSafely(client, sessionId, directory) {
  for (let attempt = 1; attempt <= 3; attempt++) {
    try {
      await client.session.delete({ path: { id: sessionId }, query: { directory } });
      return;
    } catch (error) {
      if (attempt >= 3) {
        console.warn(`[smartfetch] Failed to clean up secondary session ${sessionId} ...`);
        return;
      }
      await new Promise(resolve => setTimeout(resolve, 500));
    }
  }
}
```

## Files changed

| File | Change |
|---|---|
| `src/tools/smartfetch/secondary-model.ts` | Replace `.catch(() => undefined)` with `deleteSessionSafely()` (retry + log) |
| `src/tools/smartfetch/secondary-model.test.ts` | 2 new tests: transient delete failure (retry succeeds), persistent failure (warn logged, no throw) |

## Validation

```
4 pass, 0 fail
- falls back when the first model returns empty text     ✓
- falls back when the first model throws                  ✓
- retries session delete on transient failure             ✓ (new)
- logs warning when all delete retries fail               ✓ (new)
```

### DIFF
--- a/src/tools/smartfetch/secondary-model.test.ts
+++ b/src/tools/smartfetch/secondary-model.test.ts
@@ -7,9 +7,13 @@ type PromptStep = {
   error?: Error;
 };
 
-function createMockClient(steps: PromptStep[]) {
+function createMockClient(steps: PromptStep[], deleteBehavior?: {
+  failTimes?: number;
+}) {
   let createCount = 0;
   let promptCount = 0;
+  let deleteCallCount = 0;
+  const failTimes = deleteBehavior?.failTimes ?? 0;
 
   return {
     session: {
@@ -25,7 +29,14 @@ function createMockClient(steps: PromptStep[]) {
           },
         };
       }),
-      delete: mock(async () => ({})),
+      delete: mock(async () => {
+        deleteCallCount++;
+        if (deleteCallCount <= failTimes) {
+          throw new Error('delete failed');
+        }
+        return {};
+      }),
+      _deleteCallCount: () => deleteCallCount,
     },
     tool: {
       ids: mock(async () => ({ data: ['read', 'bash'] })),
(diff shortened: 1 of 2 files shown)


alvinunreal:
cool tnx @dhaern

dhaern:
Thanks for the review, @greptile-apps.

Both points addressed in `6711849`:

1. **Real timer delays** — extracted `_testConfig.deleteRetryDelayMs` so tests set it to `0` instead of sleeping. Retry tests now run in ~12ms combined (was ~1.5s).
2. **Unused `_deleteCallCount`** — removed from the mock.

Also added a second commit (`4303afa`) since the first review: a **30s timeout** on `client.session.prompt()` via `Promise.race`. If the secondary model upstream hangs without closing the connection, the prompt no longer blocks indefinitely — it throws, falls back to the next model, and ultimately returns raw content if all models fail. New test covers the timeout fallback path.

5/5 tests passing, ~162ms total. Could you re-review?

PR #571, opened 2026-06-17
fix: installer model provider detection

alvinunreal:
This PR fixes a crash when preset model values are non-strings (such as arrays containing objects or strings). It updates detectCurrentConfig to safely extract model IDs.

Verification:
- ran `bun test src/cli/config-io.test.ts` (33 tests pass)
- ran `bun run typecheck` (`tsc --noEmit` pass)


alvinunreal:
@greptileai

PR #573, opened 2026-06-18
fix(multiplexer): stabilize pane lifecycle cleanup

### LINKED ISSUE #553: [Bug]: Deferred multiplexer idle close is not retried after background job completion

What happened:

Background subagent panes may remain open indefinitely if an idle close is deferred while the associated background job is still running.

A background task may emit an idle event while its background job state is still `running`. The multiplexer correctly skips closing the pane at that moment:

```text
background task launch registered state:"running"
pane spawned
session status idle received backgroundJobState:"running"
close skipped; background job running
session idle event received backgroundJobState:"running"
close skipped; background job running
background job status updated state:"completed"
```

Once the background job transitions to `completed`, no additional `idle` or `session.deleted` event may arrive. As a result, the deferred close is never retried and the pane may remain open.

Expected:

If a pane close is deferred because the background job is still running, the close should be retried once the job reaches a terminal state.

This should not be addressed by restoring missing-status cleanup. Instead, the deferred close should be retried when the associated background job transitions to a terminal state.

### Steps to reproduce

1. Enable the multiplexer with tmux.
2. Launch a background subagent task that completes normally.
3. Observe a case where the child session emits `idle` while the background job board still reports the task as `running`.
4. Check the logs for `close skipped; background job running`.
5. After the background job transitions to `completed`, observe that no additional `idle` or `session.deleted` event may arrive, so the pane close is not retried.

### oh-my-opencode.json

```jsonc
{
  "multiplexer": {
    "type": "tmux"
  },
  "subtask": {
    "timeoutMs": 600000
  }
}
```

### OpenCode version

1.17.3

2.0.2

### Operating system

macOS

```shell
[task-session-manager] background task launch registered {"state":"running"}
[multiplexer-session-manager] pane spawned
[multiplexer-session-manager] session status idle received {"backgroundJobState":"running"}
[multiplexer-session-manager] close skipped; background job running
[multiplexer-session-manager] session idle event received {"backgroundJobState":"running"}
[multiplexer-session-manager] close skipped; background job running
[task-session-manager] background job status updated {"state":"completed","terminalUnreconciled":true}
```

andrewylies:
## Summary

This fixes the multiplexer pane cleanup path related to #553 while preserving the safer direction from #549: absence from `/session/status` is no longer treated as a pane close signal.

- use `/session/status` for explicit states only
- close panes on explicit idle, not on missing status entries
- clear stale deferred closes when an explicit non-idle status is observed
- retry deferred idle closes when the `BackgroundJobBoard` reaches a terminal state

This avoids flicker from closing live panes that are temporarily missing from `/session/status`, and also addresses the case where an idle close was deferred because the background job was still marked as running.

Fixes #553.
Follow-up to #549/#569.

## Behavior

The v2 background run flow is preserved:

- running background panes stay open when idle events arrive
- idle closes skipped because a background job is still running are deferred
- terminal states (`completed`, `error`, `cancelled`) retry only previously deferred idle closes
- `markCancelled()` also completes deferred pane cleanup
- terminal updates without a pending deferred close are no-ops

This does not add tmux/Zellij backend changes, watchdogs, process probing, or council/background-specific heuristics.

## Scope

This PR intentionally avoids inferring pane liveness from missing `/session/status` entries. Cleanup now relies on explicit lifecycle signals: idle/deleted events or hook-driven background job terminal states.

If no explicit lifecycle signal or background terminal update arrives, this keeps the safer fail-open behavior rather than reintroducing heuristic cleanup. Future tmux/Zellij-specific lifecycle or liveness checks can address that edge case separately.

## Validation

- `bun test src/multiplexer/session-manager.test.ts src/utils/background-job-board.test.ts`
- `bun run typecheck`
- `bun test`
- `bunx biome check src/multiplexer/session-manager.ts src/multiplexer/session-manager.test.ts src/utils/background-job-board.ts src/utils/background-job-board.test.ts src/index.ts src/multiplexer/codemap.md`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -254,6 +254,9 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
       multiplexerConfig,
       backgroundJobBoard,
     );
+    backgroundJobBoard.setTerminalStateListener((taskID) => {
+      void multiplexerSessionManager.retryDeferredIdleClose(taskID);
+    });
 
     // Initialize auto-update checker hook
     autoUpdateChecker = createAutoUpdateCheckerHook(ctx, {
--- a/src/multiplexer/codemap.md
+++ b/src/multiplexer/codemap.md
@@ -54,11 +54,10 @@
     - known sessions (`knownSessions`),
     - in-flight spawns (`spawningSessions`).
   - `respawnIfKnown` handles busy sessions that reappear after being closed.
-  - Polling fallback (`pollSessions`) is enabled when event coverage is incomplete.
-    It handles:
-    - idle detection,
-    - missing status grace period,
-    - max session lifetime timeout.
+  - Polling fallback (`pollSessions`) handles explicit idle detection only.
+    Missing from `/session/status` is not a close signal.
+  - Deferred idle closes keep panes open while `BackgroundJobBoard` says the
+    task is running, then complete via the hook-driven terminal-state callback.
 
 - `index.ts`
   - Re-exports factory, manager, and implementations for external import.
--- a/src/multiplexer/session-manager.test.ts
+++ b/src/multiplexer/session-manager.test.ts
@@ -410,6 +410,267 @@ describe('MultiplexerSessionManager', () => {
       expect(mockMultiplexer.closePane).not.toHaveBeenCalled();
     });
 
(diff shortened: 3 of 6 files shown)


andrewylies:
@greptileai

PR #574, opened 2026-06-18
Show agent provider, model, and variant in TUI

fslse:
## Summary
- Show agent provider, model, and variant as separate rows in the OMO-Slim TUI sidebar
- Persist agent variants in the TUI snapshot
- Keep preset switching updates in sync with TUI variant display
- Make configured agent/provider/model mismatches easier to spot at a glance

## Purpose
This makes it easier to quickly catch configuration mistakes, such as an agent using the wrong provider, model, or reasoning variant.

## Tests
- bun test src/tui.test.ts
- bun test src/tui-state.test.ts
- bun test src/tools/preset-manager.test.ts
- bun run typecheck
- bun run build

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -611,6 +611,7 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
       }
 
       const tuiAgentModels: Record<string, string> = {};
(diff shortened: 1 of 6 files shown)


alvinunreal:
Thanks @fslse

PR #576, opened 2026-06-19
fix(multiplexer): clear deferred idle close on explicit non-idle status

andrewylies:
Follow-up to #573.

#573 defers idle pane cleanup while a matching background job is still running,
then retries the deferred close when the job reaches a terminal state.

One event-path edge case remained: polling cleared `deferredIdleCloses` for any
explicit non-idle status, but `session.status` events only cleared it for
`busy`. A non-busy status such as `retry` could leave a stale deferred close
until the background job became terminal.

This aligns the event path with polling:

- `idle` still closes the pane or defers the close
- any explicit non-idle status clears a stale deferred idle close
- only `busy` triggers pane respawn

Validation:

- `bun test src/multiplexer/session-manager.test.ts`
- `bunx biome check src/multiplexer/session-manager.ts src/multiplexer/session-manager.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/multiplexer/session-manager.test.ts
+++ b/src/multiplexer/session-manager.test.ts
@@ -662,6 +662,68 @@ describe('MultiplexerSessionManager', () => {
       expect(mockMultiplexer.closePane).not.toHaveBeenCalled();
     });
 
+    test('explicit non-idle status event clears stale deferred idle close', async () => {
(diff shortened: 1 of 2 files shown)


alvinunreal:
thanks @andrewylies

PR #579, opened 2026-06-20
fix: inherit preset primary model for missing subagent configs

### LINKED ISSUE #499: [Bug]: Missing built-in subagent config silently falls back to hardcoded openai/* models
## Summary

When the active preset does **not** define one of the built-in subagents (for example `observer` or `councillor`), oh-my-opencode-slim silently falls back to hardcoded `openai/*` model IDs instead of warning or failing clearly.

This makes it look like the active preset switched to OpenAI even when a different preset is active.

## Reproduction

1. Use a custom `oh-my-opencode-slim.jsonc` with an active preset such as `"preset": "relaxed"`.
2. Do **not** define `observer` in that preset.
3. Enable the observer agent (`"disabled_agents": []`) or trigger a flow that uses an omitted built-in subagent.
4. Ask the system to analyze an image via `@observer`.

## Observed behavior

The spawned subagent uses a hardcoded default such as:

- `openai/gpt-5.4-mini`

This can produce errors like:

- `ProviderModelNotFoundError`
- `Model not found: openai/gpt-5.4-mini`

Even though the user never configured `openai/gpt-5.4-mini` in the active preset.

## Expected behavior

If an enabled built-in subagent is missing from the active preset, the plugin should do one of the following:

1. emit a startup warning that the agent is unconfigured and a builtin default will be used, or
2. fail clearly when that agent is invoked, or
3. require explicit configuration for enabled built-in agents, or
4. keep provider-agnostic defaults instead of hardcoded `openai/*` IDs.

What should **not** happen is a silent fallback to provider-specific OpenAI defaults that makes debugging extremely confusing.

## Likely cause

In `dist/index.js` there is a hardcoded `DEFAULT_MODELS` object containing entries like:

- `observer: "openai/gpt-5.4-mini"`
- `council: "openai/gpt-5.4-mini"`
- `councillor: "openai/gpt-5.4-mini"`

The active preset is merged, but omitted built-in agents still fall back to those hardcoded defaults.

## Suggested fix

- Avoid provider-specific hardcoded runtime defaults for built-in subagents, or
- warn loudly when an enabled built-in agent is missing from the active preset, or
- require explicit config for enabled built-in subagents.

## Environment

- oh-my-opencode-slim: 1.1.1
- Reproduced with a custom project-local slim config using non-OpenAI provider/model IDs

linze0721:
## Summary
When a preset does not define a built-in subagent (observer, councillor, etc.), the system now inherits the active preset primary model instead of silently falling back to hardcoded openai/* defaults.

Closes #499

## Changes
- Added active-preset primary model resolution in src/agents/index.ts
- Updated built-in subagent and ACP wrapper fallback to use resolved preset model
- Added regression tests (94 pass, 0 fail)

### DIFF
--- a/src/agents/custom.test.ts
+++ b/src/agents/custom.test.ts
@@ -152,6 +152,34 @@ describe('custom-agent creation', () => {
     expect(orchestrator?.config.prompt).toContain('@claude-research');
   });
 
+  test('falls back to active preset primary model for ACP wrappers', () => {
+    const config: PluginConfig = {
+      preset: 'opencode-go',
+      presets: {
+        'opencode-go': {
+          orchestrator: { model: 'opencode-go/glm-5.1' },
+        },
+      },
+      agents: {
+        orchestrator: { model: 'opencode-go/glm-5.1' },
+      },
+      acpAgents: {
+        bridge: {
+          command: 'bridge-acp',
+          args: [],
+          env: {},
+          timeoutMs: 0,
+          permissionMode: 'ask',
+        },
+      },
+    };
+
+    const agents = createAgents(config);
+    const wrapper = agents.find((agent) => agent.name === 'bridge');
+
+    expect(wrapper?.config.model).toBe('opencode-go/glm-5.1');
+  });
+
   test('falls back to oracle model for ACP wrappers', () => {
     const defaults = {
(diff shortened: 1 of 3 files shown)


alvinunreal:
@greptileai

PR #588, opened 2026-06-22
Defer todo auto-continue while background agents are active (#587)

### LINKED ISSUE #587: [Feature]: Defer auto-continue while background agents are still active
## Problem

The todo auto-continue hook can wake the main/orchestrator session as soon as it becomes idle and incomplete todos remain. However, the session may be idle only because it is waiting for one or more background agents.

In that case, auto-continue can inject a new prompt unnecessarily, even though the background agent completion will later wake the main session with its result. This can cause duplicate work, race conditions, or the orchestrator continuing before the background result is available.

## Proposed behavior

Before scheduling or sending an auto-continue prompt, check whether the current parent/orchestrator session has any non-terminal background jobs.

If at least one background job is still active, queued, pending, or otherwise non-terminal:

- do not inject the auto-continue prompt;
- cancel any pending auto-continue timer;
- keep auto-continue enabled;
- do not increment the consecutive continuation counter.

This should be a deferral rather than a permanent disable. Once the background task completes, its completion event should wake the parent session naturally.

## Suggested implementation

Reuse the existing `BackgroundJobBoard`, which already groups jobs by `parentSessionID`.

Expose a predicate to the todo-continuation hook, for example:

```ts
hasActiveBackgroundTasks?: (sessionID: string) => boolean;

### Extra context

_No response_

kumar-shivang:
Fixes #587

## Problem

OpenCode's todo auto-continue wakes the main/orchestrator session as
soon as it becomes idle with incomplete todos. When the orchestrator
is idle only because it is waiting for one or more background agents,
the wake-up is unnecessary: the background completion will wake the
orchestrator naturally with its result, and continuing before then
causes duplicate work, race conditions, or the orchestrator advancing
past the result it was actually waiting for.

## Changes

- `BackgroundJobBoard.hasActiveBackgroundTasks(parentSessionID)`: a
  predicate that returns `true` when any job under the parent is in
  a non-terminal state — still `running`, or terminal but
  `terminalUnreconciled`. Reconciled jobs are not considered active;
  they have already been consumed by the orchestrator in a prior
  turn.

- `createTodoContinuationHook`: a new hook under
  `src/hooks/todo-continuation/` that
  - on `session.idle` (or `session.status` with `type === 'idle'`)
    for a managed parent, marks the session as 'deferred' if the
    board reports active background tasks;
  - on `session.status === 'busy'` or `session.deleted`, clears
    the deferral so the next idle cycle re-evaluates the board;
  - in `experimental.chat.messages.transform`, removes the last
    user message from the outbound stream when it is a synthetic
    wake-up (`TextPart.synthetic === true`) targeting a deferred
    session. Real user messages and internal notifications (those
    carrying the `SLIM_INTERNAL_INITIATOR_MARKER` sentinel) are
    preserved.

- `src/index.ts` wires the hook into both `event` and
  `experimental.chat.messages.transform`. It runs **before**
  `taskSessionManagerHook` in the message transform pipeline so
  the synthetic wake-up is removed before any other transformation.

## Behavior

Per the issue's proposed behavior:

- Auto-continue is not injected when at least one background job is
  non-terminal.
- Auto-continue remains enabled — only the specific wake-up is
  skipped.
- The deferral is cleared as soon as the wake-up is removed, the
  session becomes busy, or the session is deleted. The next idle
  cycle re-evaluates the `BackgroundJobBoard`.

## Tests

- 7 new tests for `BackgroundJobBoard.hasActiveBackgroundTasks`
  covering: no jobs, running job, terminal-unreconciled job,
  cancelled-unreconciled job, reconciled job, parent isolation, and
  multi-job mixed states.
- 12 new tests for `createTodoContinuationHook` covering: no-op
  when no idle, non-synthetic user message preservation, synthetic
  wake-up suppression, no-defer-when-no-active-tasks, busy-clears,
  deletion-clears, non-managed-session skip, internal-marker
  preservation, `session.status` idle variant, post-reconciliation
  re-evaluation, empty messages, and no-user-message handling.
- `bun run typecheck` passes.
- `bun run check:ci` reports no new issues from these changes
  (one pre-existing format nit in `src/config/constants.ts` is
  unrelated).
- Full `bun test`: 1211 pass, 1 pre-existing failure in
  `src/agents/index.test.ts` unrelated to this PR.

## Docs

- New `src/hooks/todo-continuation/codemap.md`.
- New "Auto-Continue Deferral" section in
  `docs/background-orchestration.md`.
- `src/hooks/codemap.md` updated to list the new hook.

### DIFF
--- a/docs/background-orchestration.md
+++ b/docs/background-orchestration.md
@@ -313,6 +313,28 @@ multiplexer panes attached while the parent orchestrator continues scheduling.
 
 ---
 
+## Auto-Continue Deferral
+
+OpenCode's built-in todo auto-continue wakes the orchestrator session as soon
+as it becomes idle with incomplete todos, by injecting a synthetic user
+message into the prompt pipeline.
+
+The `todo-continuation` hook (issue #587) defers that wake-up when the
+orchestrator is idle only because it is waiting for background work. On
+`session.idle`, the hook consults
+`BackgroundJobBoard.hasActiveBackgroundTasks(parentSessionID)`. If at least
+one job under the parent is in a non-terminal state (still `running`, or
+terminal but `terminalUnreconciled`), the session is marked deferred. The
+next synthetic wake-up targeting that session is then removed from the
+outbound message array before the request reaches the LLM, so the
+orchestrator is not woken unnecessarily. The deferral is cleared on
+`session.status` `busy`, on `session.deleted`, and immediately after the
+synthetic wake-up is removed; the next idle cycle re-evaluates the board.
+Auto-continue itself remains enabled — only the specific wake-up is
+skipped.
+
+---
+
 ## Startup Behavior
 
 The installer and docs configure background subagents as a requirement for the
--- a/src/hooks/codemap.md
+++ b/src/hooks/codemap.md
@@ -50,12 +50,12 @@ and managers for all hook-based runtime behaviors used by
 |---|---|---|
 | `tool.execute.before` | Pre-process tool inputs | `apply-patch`, `task-session-manager` |
 | `tool.execute.after` | Post-process tool outputs | `delegate-task-retry`, `json-error-recovery`, `post-file-tool-nudge`, `task-session-manager` |
-| `experimental.chat.messages.transform` | Rewrite outbound user content | `filter-available-skills`, `phase-reminder` |
+| `experimental.chat.messages.transform` | Rewrite outbound user content | `filter-available-skills`, `phase-reminder`, `todo-continuation` |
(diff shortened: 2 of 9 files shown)


alvinunreal:
@kumar-shivang thanks Kumar,

I've not faced this problem, this happen on latest opencode, and latest slim plugin?

kumar-shivang:
Yes it does, 
image
auto continue keeps pushing the orchestrator even if there is some background agent working. 
(The same also happens with opencode-pty, but that's understandable)

opencode version - 1.17.9
model - gpt-5.5

alvinunreal:
this is old v1 plugin - try running fresh installation; in v2 it's removed

Issue #560, opened 2026-06-16
Foreground task fallback succeeds but result is orphaned (orchestrator only sees "Task cancelled")


Danny02:
## Summary

Foreground `task` fallback succeeds but the result is **orphaned** — the orchestrator is told the task failed (`"Task cancelled"`) and never learns about the successful fallback re-prompt. This is specific to **foreground** (blocking) `task` calls; the `background: true` path reconciles correctly.

## Environment

- `oh-my-opencode-slim`: **2.0.3**
- opencode: **1.17.7**
- node: **v26.3.0**
- OS: **macOS 15.7.7 (24G720)**
- Preset: `hybrid` with per-role fallback chains; global `fallback.enabled: true`

## What happens

1. Orchestrator calls `task` (foreground/blocking) targeting a subagent.
2. The child subagent session hits a **rate limit**.
3. opencode resolves the parent's `task` tool call **immediately** as cancelled → parent sees `"Task cancelled"` and reports failure.
4. **Separately**, `ForegroundFallbackManager` reacts to the rate-limit event, aborts the child session, and re-prompts **the same session** with the next model in the chain via `promptAsync` (returns immediately). This fallback **succeeds**.
5. But the parent tool call already returned. The fallback re-prompt was never registered as a tracked `BackgroundJobBoard` job, so no synthetic completion part is injected back into the parent, `hasTerminalUnreconciled` never flips, and the orchestrator is **never notified**. The successful result is orphaned.

## Root cause

There are two independent fallback mechanisms:

- **Delegated/subagent fallback** — synchronous `try/catch` around the in-band `prompt()`. Correct: the retry resolves *before* the tool returns.
- **Foreground fallback** (`ForegroundFallbackManager`) — purely **event-driven**. On a rate-limit signal it aborts and re-prompts the same session asynchronously.

The reconciliation system (`BackgroundJobBoard` + `task-session-manager`) only learns a terminal outcome via **(a)** the synchronous `task` tool return value, or **(b)** a **synthetic completion part** injected by the runtime for *tracked background tasks* (`updateFromInjectedCompletion` → `reconcileInjectedTerminalJobs`).

A foreground `task` that rate-limits hits neither: the tool already returned (a is "cancelled"), and it was never on the background path (b never fires). So the foreground fallback path never feeds its result back to the parent.

## Affected code (bundled `dist/index.js`, v2.0.3)

[...]

Issue #530, opened 2026-06-09
[Bug]: Foreground fallback walks the entire chain on a single rate limit (races opencode core's native retry)


leducmaxime:
**Plugin version:** `2.0.0-beta.15`
**Feature:** `fallback` (foreground model fallback / `ForegroundFallbackManager`)
**Severity:** High — a single 429 on the primary model burns the entire fallback chain and ends in `fallback chain exhausted`, despite every fallback model answering correctly.

---

## Summary

When the primary model hits a rate limit (429), the foreground fallback correctly
switches to the next model in the chain, which answers successfully. But the
session then keeps falling back to model 3, model 4, … with a growing delay
between each switch, eventually logging `fallback chain exhausted` — even though
the prompt was already answered correctly by the second model.

Root cause: the plugin's fallback runs **concurrently with opencode core's own
native 429 retry/backoff** on the original model, and the plugin's de-duplication
is only a fixed **5-second time window**. opencode core's native retry uses
exponential backoff, so its later retries land *outside* that 5s window, each
re-emits a rate-limit event, and each one is treated by the plugin as a brand-new
rate limit that advances the chain again.

## Environment

- `oh-my-opencode-slim@2.0.0-beta.15`
- Multiple providers configured with per-agent fallback chains (4 models for the
  `orchestrator` agent).
- Primary model returns `429` (normal rate limiting).

Relevant config (`oh-my-opencode-slim.json`):

```jsonc
"fallback": {
  "enabled": true,
  "timeoutMs": 15000,
  "retryDelayMs": 500,
  "retry_on_empty": true,
  "chains": {
    "orchestrator": [
      "openai/gpt-5.5",
      "anthropic/claude-opus-4-8",
      "xai/grok-4.3",
      "deepseek/deepseek-v4-pro"
    ]
    // ...
  }
}
```

## Steps to reproduce

1. Configure a fallback chain of 3+ models for an agent.
2. Send any prompt (e.g. `hi`) while the primary model is rate-limited (429).
3. Observe the session fall back through model 2, 3, 4 with growing gaps and end
   in `fallback chain exhausted`, despite model 2 already answering.

## Expected behavior

On a 429, fall back **once** to the next working model that produces a successful
answer, then stop. A late native-retry 429 for the *same* user turn should not
advance the chain further.

## Actual behavior

The whole chain is walked on a single user turn, with each model answering
correctly, until the chain is exhausted.

[...]

Issue #526, opened 2026-06-08
[Bug]: Model fallback chain triggers cascading cancellations: all models in the array get spawned & cancelled after a single rate-limit error


leducmaxime:
### What happened, and what did you expect?

I'm using a preset where every agent has a fallback chain of models (an array of model IDs). When the first model in the chain returns a usage-limit / rate-limit error, the fallback doesn't switch cleanly to the next model — it cascades.

What happens:

The first model hits a real usage-limit error.
Fallback starts the next model in the array.
The same error seems to be re-triggered/re-captured after the next model has already started running, which cancels it and advances to the next model.
Steps 2–3 repeat down the entire array until the last model is reached — and it gets cancelled too.
After a delay, one or more of the previously cancelled models sometimes resume and continue the task, occasionally several at the same time.
The net result is a racy mess: multiple models get spawned, cancelled, and partially resumed for the same task. I can confirm that only the first model in the chain was actually hitting rate limits — every successor was cancelled by a re-captured/stale error, not by its own limit.

What I expected:

On a usage-limit error, cleanly stop the failing model, switch to the next model once, and let it run to completion (unless it errors).
The original model's error should never be re-applied to a successor that has already started.
There should never be multiple models running the same task concurrently.

### Steps to reproduce

1. Configure an agent with a fallback array of 3–4 models (see config below).
2. Drive the first model into a usage-limit / rate-limit error.
3. Observe the fallback cascade: each successive model starts, then gets cancelled, down to the last model.
4. Observe that after a delay one or more of the cancelled models resume the task (sometimes concurrently).

### oh-my-opencode.json

```jsonc
{
  "$schema": "https://unpkg.com/oh-my-opencode-slim@latest/oh-my-opencode-slim.schema.json",
  "preset": "perfect",
  "presets": {
    "perfect": {
      "orchestrator": {
        "model": [
          "openai/gpt-5.5",
          "anthropic/claude-opus-4-8",
          "xai/grok-4.3",
          "deepseek/deepseek-v4-pro"
        ],
        "skills": ["*"],
        "mcps": ["*", "!context7"]
      },
      "oracle": {
        "model": [
          "anthropic/claude-opus-4-8",
          "openai/gpt-5.5",
          "xai/grok-4.3",
          "deepseek/deepseek-reasoner"
        ]
      },
      "fixer": {
        "model": [
          "deepseek/deepseek-v4-flash",
          "openai/gpt-5.4-mini",
[...]

Issue #180, opened 2026-03-16
Fallback chains do not work for foreground agents; balanceProviderUsage appears unimplemented


AlejandroAkbal:
## Summary
The current config/schema suggests that fallback chains and provider balancing apply generally, but in practice runtime fallback only appears to execute in the background task manager path used for subagents/background tasks. Foreground/main agent usage does not appear to get runtime failover, and `balanceProviderUsage` appears to be declared in config/schema without runtime behavior.

This is misleading for users configuring the plugin, because the config shape implies broader support than the installed runtime actually provides.

## Environment
- oh-my-opencode-slim: `0.8.3`
- inspected from local installed package

### 1) Runtime fallback exists in the background task manager path
In the installed build, the fallback loop is implemented in the background manager:
- `dist/index.js:18166-18188` — `resolveFallbackChain(agentName)`
- `dist/index.js:18243-18279` — `startTask(...)` builds `chain`, iterates models, and retries sequentially

That code clearly uses `fallback.enabled`, `fallback.timeoutMs`, and `fallback.chains`.

### 2) Foreground/main agent config does not appear to get runtime failover
In the plugin config path:
- `dist/index.js:33577-33636`

The plugin merges agents into `opencodeConfig.agent` and, when `_modelArray` exists, resolves only the **first** model whose provider exists in configured providers:
- `dist/index.js:33600-33623`

That looks like config-time provider presence resolution, not runtime fallback on request failure / rate limit / provider error.

### 3) `balanceProviderUsage` appears unimplemented
I found the setting declared in schema/config:
- `dist/index.js:17057`
- `dist/cli/index.js:13680`
- `dist/config/schema.d.ts:129`

But I could not find runtime logic actually using it. In the installed build it appears to be a dead config flag.

## User-visible effect
With config like this:
```json
{
  "fallback": {
    "enabled": true,
    "timeoutMs": 15000,
    "chains": {
      "orchestrator": ["openai/gpt-5.4", "github-copilot/gpt-5.4"],
      "fixer": ["openai/gpt-5.3-codex", "github-copilot/gpt-5.3-codex"]
    }
  },
  "balanceProviderUsage": true
}

PR #593, opened 2026-06-23
fix(fallback): resolve three deadlock points in foreground fallback + task-session-manager

### LINKED ISSUE #592: [Bug] Three fallback deadlock points: chain-exhausted permanent deadlock + unknown agents (compaction) excluded + session.error erases terminal job state
## Summary

Three independent deadlock points in `ForegroundFallbackManager` + `task-session-manager` that break fallback in v2.0.4/v2.0.5 when the OpenAI/GPT primary model hits its usage limit (ChatGPT subscription OAuth → `"The usage limit has been reached"`).

Each is reproducible by exhausting the primary model's quota, observing omos logs (`~/.local/share/opencode/log/oh-my-opencode-slim.*.log`), and cross-referencing `opencode.log`.

## Environment

- `oh-my-opencode-slim`: **2.0.5** (also affects 2.0.3, 2.0.4)
- opencode: **1.17.7**
- OS: **Ubuntu 22.04 (Linux)**
- Provider: `openai/gpt-5.4` (ChatGPT OAuth, no API key — usage limit is the OAuth subscription cap)
- Fallback chain: `[openai/gpt-5.4, new-api/glm-5.2, new-api/deepseek-v4-pro]` (3-level)

## Investigation methodology

Diagnosed by reading `dist/index.js` (bundle) and correlating with runtime logs:

- **omos log pattern for Bug A**: `"fallback chain exhausted"` appearing consecutively without any `"switched to fallback"` in between → chain exhausted, session permanently stuck.
- **omos log pattern for Bug B**: `"no chain configured" {"agentName":"compaction"}` — compaction/title agents have no chain.
- **opencode.log pattern for Bug C**: `AI_APICallError: The usage limit has been reached` → fallback succeeds → but `"parsed task output status"` → `"ignored status for unknown background job"` repeats in a busy/idle oscillation loop → orchestrator cannot dispatch new tasks.

---

### Location

`ForegroundFallbackManager.tryFallback()` — `dist/index.js` lines ~24395–24406 (v2.0.5).

### Root cause

When all models in the chain have been tried (e.g., `tried == {"openai/gpt-5.4", "new-api/glm-5.2", "new-api/deepseek-v4-pro"}`), the code does a bare `return` **without resetting `tried`**. Since `sessionTried` is only cleared on `session.deleted`, the session is permanently deadlocked — every subsequent message retries the primary model (gpt), trigger the error, hit `fallback chain exhausted` immediately, and give up.

### Native code (deadlock)

```js
// dist/index.js v2.0.5, lines 24392–24406
if (!this.sessionTried.has(sessionID)) {
  this.sessionTried.set(sessionID, new Set);
}
const tried = this.sessionTried.get(sessionID);
if (currentModel)
  tried.add(currentModel);
const nextModel = chain.find((m) => !tried.has(m));
if (!nextModel) {
  log("[foreground-fallback] fallback chain exhausted", {
    sessionID,
    agentName,
    tried: [...tried]
  });
  return;  // ← deadlock: tried set never reset
}
```

### Patched code (sticky exhaustion)

When `chain.length > 1`, reset `tried` to only contain the primary model, then stick to the deepest fallback model (`chain[chain.length - 1]`). Subsequent messages will go: primary fails → chain.find skips primary (it's in tried) → selects the sticky fallback directly, no more exhaustion.

```js
let tried = this.sessionTried.get(sessionID);
if (currentModel)
  tried.add(currentModel);
let nextModel = chain.find((m) => !tried.has(m));
if (!nextModel) {
  if (chain.length > 1) {
    const primary = chain[0];
    const stickyFallback = chain[chain.length - 1];
    log("[foreground-fallback] resetting tried set for re-fallback", {
      sessionID, agentName, currentModel,
      prevTried: [...tried], nextModel: stickyFallback
    });
    tried = new Set();
    if (primary) tried.add(primary);
    if (currentModel && currentModel !== primary) tried.add(currentModel);
    this.sessionTried.set(sessionID, tried);
    nextModel = stickyFallback;
  } else {
    log("[foreground-fallback] fallback chain exhausted", {
      sessionID, agentName, tried: [...tried]
    });
    return;
  }
}
```

---

### Location

`ForegroundFallbackManager.resolveChain()` — `dist/index.js` lines ~24476–24478 (v2.0.5).

### Root cause

`resolveChain` returns `[]` immediately for any known agent name that has no chain configured:

```js
resolveChain(agentName, currentModel) {
  if (agentName) {
    return this.chains[agentName] ?? [];  // ← compaction/title not in chains → []
  }
  // ...
}
```

This was an intentional design choice from PR #199 (to prevent `oracle` from bleeding into `orchestrator`'s chain). However, OpenCode's built-in agents like `compaction` and `title` have no preset config — they ALWAYS fall into this `?? []` path. They have no chain to fall through to, so they never get fallback at all. The log is filled with:

```
[foreground-fallback] no chain configured {"sessionID":"ses_xxx","agentName":"compaction"}
```

Meanwhile `opencode.log` shows compaction retrying `openai/gpt-5.4` endlessly with `"The usage limit has been reached"`.

### Patched code (fall through to model-matching for unknown agents)

```js
resolveChain(agentName, currentModel) {
  if (agentName) {
    const chain = this.chains[agentName];
    if (chain) return chain;
    // Fall through: unknown agent (compaction, title) — try model matching below
  }
  if (currentModel) {
    for (const chain of Object.values(this.chains)) {
      if (chain.includes(currentModel)) return chain;
    }
  }
  // ...
}
```

Now compaction's gpt-5.4 matches orchestrator's chain → fallback works for compaction too.

---

### Location

`createTaskSessionManagerHook()` event handler — `dist/index.js` lines ~25360–25365 (v2.0.5).

### Root cause

Both `ForegroundFallbackManager` and `task-session-manager` listen for `"session.error"`. When a rate-limit fires:

1. **FL Manager** correctly catches it → `abort` + delayed `promptAsync` with fallback model (good).
2. **Task-session-manager** also catches it → unconditionally **deletes** `terminalJobsInjectedByParent` for that session (bad).

```js
// dist/index.js v2.0.5, lines 25360–25365
if (input.event.type === "session.error") {
  const sessionId2 = input.event.properties?.info?.id ?? input.event.properties?.sessionID;
  if (sessionId2 && options.shouldManageSession(sessionId2)) {
    terminalJobsInjectedByParent.delete(sessionId2);  // ← deletes ON ANY session.error
  }
  return;
}
```

After the deletion, the orchestrator's fallback re-prompt succeeds (model switched to glm/deepseek), but the orchestrator has lost track of which background tasks completed. The task-session-manager's `hasTerminalUnreconciled` never flips, and the orchestrator enters a busy/idle oscillation, unable to dispatch follow-up tasks.

### Patched code (only delete for non-rate-limit errors)

```js
if (input.event.type === "session.error") {
  const sessionId2 = input.event.properties?.info?.id ?? input.event.properties?.sessionID;
  if (sessionId2 && options.shouldManageSession(sessionId2)) {
    const props = input.event.properties;
    if (!props?.error || !isRateLimitError(props.error)) {
      terminalJobsInjectedByParent.delete(sessionId2);
    }
  }
  return;
}
```

`isRateLimitError` is already a standalone function in the same bundle scope (line ~24279), so this doesn't introduce a new dependency. Rate-limit errors are recovered by ForegroundFallbackManager → the terminal job state should survive.

### Relationship to issue #560

Issue #560 describes a deeper form of this problem: the foreground fallback re-prompt result is **never registered as a tracked background job** at all, so `hasTerminalUnreconciled` never flips regardless of `terminalJobsInjectedByParent`. Bug C fixes the state-erasure on `session.error`, but the deeper "orphaned result" problem in #560 remains and needs upstream attention (registering the fallback re-prompt as a tracked job in BackgroundJobBoard).

---

## Related issues

| Issue | Relation |
|---|---|
| [#560](https://github.com/alvinunreal/oh-my-opencode-slim/issues/560) | Describes the orphaned-result problem; Bug C is a partial fix |
| [#526](https://github.com/alvinunreal/oh-my-opencode-slim/issues/526) / [#530](https://github.com/alvinunreal/oh-my-opencode-slim/issues/530) | Cascade/cancel bugs related to Bug A but not the same deadlock |
| [#180](https://github.com/alvinunreal/oh-my-opencode-slim/issues/180) → [#199](https://github.com/alvinunreal/oh-my-opencode-slim/pull/199) | Introduced ForegroundFallbackManager; intentional `[]` return for unknown agents causes Bug B |
| [#511](https://github.com/alvinunreal/oh-my-opencode-slim/issues/511) | Forbidden/401 not in RATE_LIMIT_PATTERNS — a separate gap |

## Patches

All three patches are applied to the running instance and verified via logs. A self-contained Python script that applies them by code-feature matching (no backup dependency, idempotent) is available. Happy to contribute these as a PR if there's interest.

The combined effect: the 3-level chain `[gpt-5.4 → glm-5.2 → deepseek-v4-pro]` now properly falls back through all models, sticks to the deepest fallback after exhaustion, covers compaction/title agents, and preserves background task state through rate-limit recovery so the orchestrator can continue dispatching.

henlii:

Three independent deadlock points break model fallback in v2.0.5 when the primary model hits its usage limit (e.g. ChatGPT OAuth → `"The usage limit has been reached"`). Each was diagnosed by correlating `dist/index.js` source with runtime logs (`oh-my-opencode-slim.*.log` + `opencode.log`).

Full details in #592.

---

### Bug A — chain-exhausted permanent deadlock (`foreground-fallback`)

`tryFallback()` returns without resetting `tried` when all chain models have been attempted. Since `sessionTried` is only cleared on `session.deleted`, the session is **permanently deadlocked** — every subsequent message retries the dead primary, hits the error, sees `fallback chain exhausted`, and gives up.

**Fix**: when `chain.length > 1`, reset `tried` (keeping only the primary marked) and stick to the deepest fallback model (`chain[chain.length - 1]`). Subsequent messages go: primary fails → skip primary (in tried) → select sticky fallback directly.

### Bug B — non-omos agents excluded from fallback (`foreground-fallback`)

`resolveChain()` returns `[]` for any agent without a configured chain. This was intentional for **known omos agents** (preserves cross-agent isolation from PR #199 — prevents oracle from borrowing orchestrator's models). But OpenCode built-in agents like `compaction` and `title` are **not** omos agents and never have chains — they are silently excluded from fallback entirely, leading to infinite retries on the dead primary.

**Fix**: fall through to model-matching for agents that are **not** in `ALL_AGENT_NAMES` (i.e. not omos built-ins). Known omos agents without chains keep the isolation (`return []`). Non-omos agents (compaction, title) inherit a chain from a configured agent that shares their model.

### Bug C — `session.error` erases terminal job state (`task-session-manager`)

Both `ForegroundFallbackManager` and `task-session-manager` listen for `session.error`. FFM correctly recovers via abort + reprompt with fallback model. But TSM **unconditionally deletes** `terminalJobsInjectedByParent` on any `session.error` — losing track of completed background tasks. After the fallback reprompt succeeds, the orchestrator has lost its background task state and cannot dispatch follow-ups.

**Fix**: only clear `terminalJobsInjectedByParent` for **non-rate-limit** errors. Rate-limit errors are recovered by FFM; the terminal job state must survive so the orchestrator can continue. Uses the existing `isRateLimitError()` from `foreground-fallback/index.ts`.

This partially addresses #560 (the deeper orphaned-result problem — fallback re-prompt never registered as a tracked job — still needs upstream work).

---

## Testing

- `bun run check:ci` — passes (Biome lint + format)
- `bun run typecheck` — passes (`tsc --noEmit`)
- `bun test` — all relevant tests pass (31/31 foreground-fallback, 41/41 task-session-manager). Added a new test for compaction fallthrough. 2 pre-existing failures in unrelated modules (`auto-update-checker/skill-sync`, `codemap`).

## Related

- Closes #592
- Partially addresses #560 (orphaned foreground fallback result)
- Relates to #526, #530 (cascade/cancel bugs in the same area)
- Relates to #180 / PR #199 (introduced ForegroundFallbackManager + the intentional isolation that Bug B refines)

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -698,4 +698,39 @@ describe('ForegroundFallbackManager resolveChain cross-agent isolation', () => {
     expect(call[0].body.model.providerID).toBe('openai');
     expect(call[0].body.model.modelID).toBe('gpt-4o');
   });
+
+  test('falls through to model matching for non-omos agents (e.g. compaction)', async () => {
+    // compaction is an OpenCode built-in agent that is NOT an omos agent,
+    // so it has no chain configured. It should fall through to model
+    // matching and inherit a chain from a configured agent that shares
+    // its model, instead of being silently excluded from fallback.
+    const { client, mocks } = createMockClient();
+    const mgr = new ForegroundFallbackManager(
+      client,
+      { orchestrator: ['openai/gpt-5.4', 'new-api/glm-5.2'] },
+      true,
+    );
+
+    await mgr.handleEvent({
+      type: 'message.updated',
+      properties: {
+        info: {
+          sessionID: 'compaction-sess',
+          agent: 'compaction', // NOT a known omos built-in agent
+          providerID: 'openai',
+          modelID: 'gpt-5.4',
+          error: { message: 'rate limit exceeded' },
+        },
+      },
+    });
+
+    // compaction's model (openai/gpt-5.4) matches orchestrator's chain
+    // → should fall back to the next untried model in that chain
+    expect(mocks.promptAsync).toHaveBeenCalledTimes(1);
+    const call = mocks.promptAsync.mock.calls[0] as [
+      { body: { model: { providerID: string; modelID: string } } },
+    ];
+    expect(call[0].body.model.providerID).toBe('new-api');
+    expect(call[0].body.model.modelID).toBe('glm-5.2');
+  });
 });
--- a/src/hooks/foreground-fallback/index.ts
+++ b/src/hooks/foreground-fallback/index.ts
@@ -15,6 +15,7 @@
  */
 
 import type { PluginInput } from '@opencode-ai/plugin';
+import { ALL_AGENT_NAMES } from '../../config/constants';
 import { log } from '../../utils/logger';
 import {
   abortSessionWithTimeout,
@@ -236,17 +237,37 @@ export class ForegroundFallbackManager {
         this.sessionTried.set(sessionID, new Set());
       }
       // biome-ignore lint/style/noNonNullAssertion: We just set this above
-      const tried = this.sessionTried.get(sessionID)!;
+      let tried = this.sessionTried.get(sessionID)!;
       if (currentModel) tried.add(currentModel);
 
-      const nextModel = chain.find((m) => !tried.has(m));
+      let nextModel = chain.find((m) => !tried.has(m));
       if (!nextModel) {
-        log('[foreground-fallback] fallback chain exhausted', {
-          sessionID,
-          agentName,
-          tried: [...tried],
-        });
-        return;
+        if (chain.length > 1) {
+          // Chain exhausted but we have fallbacks: reset tried set and
+          // stick to the deepest fallback model so we stop re-trying the
+          // dead primary model on every subsequent message.
+          const primary = chain[0];
+          const stickyFallback = chain[chain.length - 1];
+          log('[foreground-fallback] resetting tried set for re-fallback', {
+            sessionID,
+            agentName,
+            currentModel,
+            prevTried: [...tried],
+            nextModel: stickyFallback,
+          });
+          tried = new Set();
+          if (primary) tried.add(primary);
+          if (currentModel && currentModel !== primary) tried.add(currentModel);
+          this.sessionTried.set(sessionID, tried);
+          nextModel = stickyFallback;
+        } else {
+          log('[foreground-fallback] fallback chain exhausted', {
+            sessionID,
+            agentName,
+            tried: [...tried],
+          });
+          return;
+        }
       }
       tried.add(nextModel);
 
@@ -352,9 +373,18 @@ export class ForegroundFallbackManager {
     currentModel: string | undefined,
   ): string[] {
     if (agentName) {
-      // Agent is known: use its chain exactly, or no chain at all.
-      // Never fall through to cross-agent chains when the agent is identified.
-      return this.chains[agentName] ?? [];
+      // Agent is known: use its chain exactly if configured.
+      const chain = this.chains[agentName];
+      if (chain) return chain;
+      // Known omos built-in agent (oracle, librarian, …) without a
+      // configured chain: keep isolation — do NOT bleed into other
+      // agents' chains (preserves the cross-agent isolation contract
+      // from PR #199).
+      if ((ALL_AGENT_NAMES as readonly string[]).includes(agentName)) return [];
+      // Unknown agent (e.g. OpenCode built-in "compaction" or "title"
+      // that don't appear in the user preset): fall through to
+      // model-matching so they can inherit a chain from a configured
+      // agent that shares their model.
     }
 
     // Agent unknown: try to infer from the current model.
--- a/src/hooks/task-session-manager/index.ts
+++ b/src/hooks/task-session-manager/index.ts
@@ -12,6 +12,7 @@ import {
 } from '../../utils';
 import { isRecord as isObjectRecord } from '../../utils/guards';
 import { log } from '../../utils/logger';
+import { isRateLimitError } from '../foreground-fallback/index';
 import type { MessagePart, MessageWithParts } from '../types';
 
 interface TaskArgs {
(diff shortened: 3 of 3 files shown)


alvinunreal:
thanks @824156793

PR #602, opened 2026-06-24
fix: clear interview fallback timer

Zaradacht:
## Summary
- Clear the interview fallback polling interval after the final registered session is deleted.
- Add focused regression coverage for the timer cleanup path.

## Validation\n- bun test src/interview/manager.test.ts src/interview/interview.test.ts src/interview/dashboard.test.ts\n- bun run typecheck\n- bun run check:ci\n- bun test\n\n## Scope\n- Implements one small resource cleanup candidate only.\n- Leaves other follow-up candidates for separate work.


Zaradacht:
@greptileai

Issue #124, opened 2026-02-09
[Bug]: tmux integration doesn't cleanup subagent sessions after task completion - resource leak


rizal72:
## Bug Description

When tmux integration is enabled, subagent sessions spawned for task execution are never cleaned up after task completion. This causes:
- Persistent tmux panes remaining open indefinitely
- CPU waste (~24-30% per idle session)
- Memory consumption (~200-400MB per session)
- Socket connection bloat (each session maintains 10+ WebSocket connections)
- System instability over time

## Steps to Reproduce

1. Enable tmux integration in config:
```json
{
  "tmux": {
    "enabled": true,
    "layout": "main-vertical",
    "main_pane_size": 60
  }
}
```

2. Start OpenCode in server mode
3. Launch any task (synchronous or background) that spawns subagents
4. Wait for task completion
5. Observe that tmux panes remain open even after task finishes
6. Check process list: `ps aux | grep opencode | grep attach`

## Expected Behavior

- tmux panes should auto-terminate after subagent task completion
- Sessions should be cleaned up automatically
- No persistent idle processes consuming resources

## Actual Behavior

- tmux panes remain open indefinitely after task completion
- Multiple `opencode attach http://127.0.0.1:4096/ --session ses_XXX` processes remain active
- Each process consumes ~24-30% CPU even when idle
- Sessions are not tracked in `opencode session list` (these are transient worker sessions)

## Impact

- **Resource leak**: CPU, memory, and sockets are not released
- **System performance**: Significant CPU usage even when not using the system
- **Scalability**: Issue compounds with each new task executed
- **User experience**: tmux integration becomes unusable due to resource constraints

## Workaround

Disable tmux integration in config:
```json
{
  "tmux": {
    "enabled": false
  }
}
```

Manual cleanup command (temporary):
```bash
pkill -f "opencode attach"
```

## Environment

- **Operating System**: macOS 26.2 Tahoe
- **OpenCode Version**: 1.1.51
- **Plugin Version**: oh-my-opencode-slim (latest)
- **tmux version**: 3.5a

## Process Hierarchy Example

```
tmux (42382)
├── opencode --port 4096 -c (42470) [Main server]
├── opencode attach ... (43030) [Orphaned - no parent cleanup]
├── opencode attach ... (43142) [Orphaned]
├── opencode attach ... (43028) [Orphaned]
└── opencode attach ... (43141) [Orphaned]
```

## Additional Context

[...]

PR #603, opened 2026-06-24
fix: clean up subagent attach sessions

Zaradacht:
## Summary
- close tracked multiplexer panes when background jobs complete, fail, cancel, or time out
- stop timed-out running jobs from deferring idle cleanup forever
- keep cleanup scoped to plugin-tracked pane/session IDs without broad process matching

## Validation
- bun test src/multiplexer/session-manager.test.ts
- bun test src/utils/background-job-board.test.ts
- bun test src/multiplexer/session-manager.test.ts src/utils/background-job-board.test.ts
- bun run typecheck
- bun run check:ci
- bun test

## Related issue
- #124

### DIFF
--- a/src/multiplexer/session-manager.test.ts
+++ b/src/multiplexer/session-manager.test.ts
@@ -522,6 +522,53 @@ describe('MultiplexerSessionManager', () => {
       }
     });
 
+    test('deferred idle close retries when the background job times out', async () => {
+      const ctx = createMockContext();
+      const board = new BackgroundJobBoard();
+      board.registerLaunch({
+        taskID: 'timed-out-deferred',
+        parentSessionID: 'parent-1',
+        agent: 'explorer',
+      });
+      mockMultiplexer.spawnPane.mockResolvedValue({
+        success: true,
+        paneId: 'p-timed-out-deferred',
+      });
+      const manager = new MultiplexerSessionManager(
+        ctx,
+        defaultMultiplexerConfig,
+        board,
+      );
+      board.setTerminalStateListener((taskID) => {
+        void manager.retryDeferredIdleClose(taskID);
+      });
+
+      await manager.onSessionCreated({
+        type: 'session.created',
+        properties: {
+          info: { id: 'timed-out-deferred', parentID: 'parent-1' },
+        },
+      });
+      await manager.onSessionStatus({
+        type: 'session.status',
+        properties: {
+          sessionID: 'timed-out-deferred',
+          status: { type: 'idle' },
+        },
+      });
+
+      board.updateStatus({
+        taskID: 'timed-out-deferred',
+        state: 'running',
+        timedOut: true,
+      });
+      await Promise.resolve();
+
+      expect(mockMultiplexer.closePane).toHaveBeenCalledWith(
+        'p-timed-out-deferred',
+      );
+    });
+
     test('deferred idle close retries on markCancelled', async () => {
       const ctx = createMockContext();
       const board = new BackgroundJobBoard();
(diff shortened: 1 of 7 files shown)


Zaradacht:
@greptileai

Zaradacht:
@greptileai

Zaradacht:
@alvinreal

mhenke:
## Review: ✅ Approve with non-blocking polish

### Logic
- Second commit (`8f2e653`) correctly tightened `retryDeferredIdleClose` condition and renamed `terminalStateListener` → `ActionableStateListener`
- Timed-out-without-prior-defer path is covered (`session-manager.test.ts:503`)

### Non-blocking suggestions

1. **Add JSDoc to `ActionableStateListener`** (`background-job-board.ts:61`)
   - The type alias has no doc — explain "actionable" means terminal states + timed-out running jobs

2. **Add test: listener fires once for timeout→terminal transition** (`background-job-board.test.ts`)
   - The `!existing.timedOut` guard prevents double-firing when a timed-out job later goes terminal, but no test documents that contract

3. **Extract `isActiveRunningJob()` helper** (`background-job-board.ts`)
   - `job.state === 'running' && !job.timedOut` appears in two places in `session-manager.ts` (`isRunningBackgroundJob` + inline in `retryDeferredIdleClose`) — extract to the job board module for reuse

alvinunreal:
Quick heads-up: this PR is currently showing `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Could you please rebase/update onto latest `master` and repush so we can clear conflicts and re-run checks before merging?

mhenke:
@greptile-apps please recheck this

mhenke:
@greptile-apps

mhenke:
@greptile-apps please recheck these changes and update your chart/rating

mhenke:
CI is green and this PR is mergeable. Two follow-up commits landed after the original feature work.

`2a5f098` fixed the failing `test (latest)`. `handleTerminalState` now notifies listeners on any terminal or timeout state, but the old test still expected notification only after a prior defer. The test now matches the behavior.

`5930358` removed `retryDeferredClose`, which had no callers left after the change above. It is gone from the `BackgroundJobStore` interface, the board and coordinator implementations, and its tests. The one stale test that referenced it by name was renamed. The `terminalStateListener` rename Greptile suggested was skipped: it touches several files and changes nothing about how the code runs.

Also resolved: a stray cosmetic edit to `foreground-fallback/index.ts` that predated this work conflicted with the branch and was dropped from the PR to keep the change focused.

Totals vs master: 7 files, +133 and -71. Greptile's two reviews were non-blocking and rated safe to merge. One thing left behind on purpose: `deferredIdleCloses` in `background-job-coordinator.ts` is now written but never read. It is harmless dead state and can be cleaned up later if anyone cares.

mhenke:
**Closing this PR.**

After reviewing against current master, this PR is mostly superseded by the cleanup infrastructure that's been merged since it was opened. A few reasons we're not merging it as-is:

1. **`promptSafe()` removal is a regression risk** — it's actively used in 8 places to HTML-escape user-controlled strings before injecting them into system prompts. Removing it opens prompt injection vectors.
2. **Removing `CmuxSessionLifecycle` entirely** drops mature orphan detection, activity tracking, and close retry logic that's already working on master.
3. **The `notifyTimedOut` logic** (the one genuinely useful piece) is better addressed by the stuck-agent detection on `fix/67-stuck-agent-detection`, which handles the full lifecycle — timeout, abort, cleanup, with proper grace periods.

Thanks for the work on this — the investigation was valuable and informed the stuck-agent detection approach.

Issue #600, opened 2026-06-24
Second round memory/resource optimization follow-ups


Zaradacht:
## Summary
Follow up on the resource-retention audit from #597 after PR #598 delivered the first small fix. This second round should pick the next highest-value small chunk from the remaining concrete findings, not broaden into a large refactor.

## Source context
- Source issue: #597
- First-round PR: #598
- First-round outcome: dashboard shutdown now releases its cleanup timer and active SSE clients.

## Candidate follow-up chunks
Use the first-round audit evidence to choose one small safe chunk, such as:
- Multiplexer/plugin teardown review: determine whether a plugin/session lifecycle hook can safely call `MultiplexerSessionManager.cleanup()` without regressing tmux/zellij behavior.
- Background job-board alias-counter pruning: prune parent/agent alias counters after all jobs for a parent/prefix are removed.
- Interview manager fallback timer cleanup: stop the dashboard-client fallback polling interval when no sessions remain registered.
- Interview service terminal record pruning: bound or prune abandoned/completed interview records and related browser-open tracking.
- Browser reconnect timer dedupe: avoid queuing repeated `/interview` SSE reconnect timers in the browser UI.
- Companion manager listener cleanup: remove process exit listeners and clear detached child references when manager lifecycle ends.
- Smartfetch binary cache policy: evaluate whether binary bodies should be cached or metadata-only cached under some options.

## Expected outcome
- Re-read the relevant first-round evidence and code before choosing the chunk.
- Pick only one high-confidence, low-risk resource optimization.
- Implement that small fix only if evidence supports it.
- Add focused regression tests for the changed lifecycle/resource path.
- Keep the remaining candidates as follow-up notes instead of expanding scope.

## Validation hints
- `bun run check:ci`
- `bun run typecheck`
- `bun test`
- Targeted tests based on chosen area, for example:
  - `bun test src/multiplexer/session-manager.test.ts`
  - `bun test src/utils/background-job-board.test.ts`
  - `bun test src/hooks/task-session-manager/index.test.ts`
  - `bun test src/interview/manager.test.ts src/interview/interview.test.ts src/interview/dashboard.test.ts`
  - `bun test src/companion/manager.test.ts`
  - `bun test src/tools/smartfetch/cache.test.ts src/tools/smartfetch/tool.test.ts`

## Open questions
- Which remaining candidate gives the best risk/reward after #598?
[...]

Issue #597, opened 2026-06-24
Audit memory leaks and resource optimization opportunities


Zaradacht:
## Summary
Audit oh-my-opencode-slim for memory leaks, long-lived resource retention, and practical performance/resource optimizations. Keep the work incremental: first understand the overall situation, then propose and implement small, low-risk chunks rather than a large refactor.

## Context
No broad optimization pass has been done yet. Likely hotspots from bounded repo discovery include:
- `src/index.ts` plugin bootstrap and session event fan-out/cleanup.
- `src/multiplexer/session-manager.ts` global session/pane tracking and deferred cleanup state.
- `src/utils/background-job-board.ts` in-memory job registry, alias retention, and context-file tracking.
- `src/interview/service.ts` and `src/interview/ui.ts` long-lived maps, sets, polling intervals, and browser timers.
- `src/companion/manager.ts` detached process handling, busy-session tracking, and process exit listeners.
- `src/companion/updater.ts` temp download/extraction cleanup paths.
- `src/tools/smartfetch/cache.ts` 50 MB in-memory LRU cache and TTL behavior.

A narrower existing task covers tmux/subagent cleanup. This issue should stay broader and identify the next small chunks across lifecycle, timers, detached processes, caches, and profiling opportunities.

## Expected outcome
- Produce a concise inventory of concrete memory/resource risks with evidence.
- Identify where profiling or benchmarking is practical; if not practical, explain why and use targeted diagnostics/tests instead.
- Implement only the first safe, small optimization/fix chunk when evidence supports it.
- Avoid broad rewrites or speculative optimization.

## Validation hints
- `bun run check:ci`
- `bun run typecheck`
- `bun test`
- Targeted tests such as multiplexer/session-manager, background task/session-manager, companion manager/updater, interview, auto-update checker, and smartfetch cache tests.
- Consider `bun run build` and `bun run verify:host-smoke` if runtime/plugin loading behavior changes.

## Open questions
- Which first chunk gives the best risk/reward after the audit?
- Should benchmarking/profiling be added as durable tooling, or is targeted test/diagnostic coverage enough for the first pass?

Zaradacht:
https://github.com/alvinunreal Action required: review draft PR #598 and the remaining resource-audit findings before deciding the next optimization chunk.
Delivery status: accepted
PR: https://github.com/alvinunreal/oh-my-opencode-slim/pull/598
Contract: GitHub issue #597
[...]

PR #613, opened 2026-06-25
fix(background-job-board): prune leaked alias counters on job removal

Zaradacht:
## Summary

Second-round resource fix following the audit in #597 and the follow-up tracker #600.

`BackgroundJobBoard` kept an alias `counters` Map keyed by `${parentSessionID}:${prefix}` that grew monotonically via `nextAlias()` and was **never pruned**. The job-deletion paths (`clearParent()`, `drop()`, `trimReusable()`) removed the job records but left the counter entries behind, leaking one entry per (parent session × agent prefix) for the life of long-running serve/dashboard processes.

This is the "background job-board alias-counter pruning" candidate listed in #600.

## Changes

- Extract prefix derivation into a shared `aliasPrefix()` helper.
- Add `pruneCounter(parentSessionID, agent)` that releases a counter entry **only when no remaining job shares that parent+prefix**. Resetting the count is safe under this invariant because aliases only need to stay unique among jobs that still exist — a reset count can never collide with a live alias.
- Wire pruning into all three job-deletion paths: `drop()`, `clearParent()` (collects distinct agents, prunes each), and `trimReusable()`.

## Tests

Added 3 focused regression tests (observable via alias numbering, since `counters` is private):
- Counter resets to `exp-1` after the last job for a prefix is dropped.
- Counter is preserved (`exp-3`) while another same-prefix job is still live — guards against alias collision.
- `clearParent` releases only that parent's counters, leaving sibling parents untouched.

## Validation

- `bun run typecheck` ✅
- `bun test src/utils/background-job-board.test.ts` → 34 pass ✅
- `bun test` (full suite) → 1217 pass / 0 fail ✅

## Scope

Intentionally one small chunk per #600's guidance. Remaining verified audit candidates left as follow-ups: interview-service record pruning (`interviewsById`/`browserOpened`), multiplexer `cleanup()` teardown + orphaned poll timer on re-init, and companion `process.on('exit')` listener accumulation.

### DIFF
--- a/src/utils/background-job-board.test.ts
+++ b/src/utils/background-job-board.test.ts
@@ -698,4 +698,82 @@ describe('BackgroundJobBoard', () => {
     const prompt = board.formatForPrompt('parent-1', 9_000);
     expect(prompt).toContain('running [resumed, 4s ago]');
   });
+
+  test('resets the alias counter after the last job for a prefix is dropped', () => {
+    const board = new BackgroundJobBoard();
+    const first = board.registerLaunch({
+      taskID: 'ses_1',
+      parentSessionID: 'parent-1',
+      agent: 'explorer',
+    });
+    expect(first.alias).toBe('exp-1');
+
+    board.drop('ses_1');
+
+    // No explorer jobs remain for parent-1, so the counter is released and
+    // the next alias restarts at 1 instead of leaking to exp-2.
+    const second = board.registerLaunch({
+      taskID: 'ses_2',
+      parentSessionID: 'parent-1',
+      agent: 'explorer',
+    });
+    expect(second.alias).toBe('exp-1');
+  });
+
+  test('keeps the alias counter while another job of the same prefix remains', () => {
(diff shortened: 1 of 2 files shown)


Zaradacht:
@greptileai please review

alvinunreal:
Quick heads-up: this PR is currently showing `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Could you please rebase/update onto latest `master` and repush so we can clear conflicts and re-run checks before merging?

mhenke:
Thanks for this, @Zaradacht. The alias-counter leak analysis is solid, and the pruneCounter approach reads clean.

It's been sitting since July while master moved through several background-job-board changes, so it now conflicts with master and needs a rebase before it can merge. Rather than leave it open and gathering conflicts, we're closing it as stale for now.

If you're still interested in picking it up, a fresh PR (or a reopen of this one after a rebase) is welcome. The candidate item is still tracked in #600.

PR #614, opened 2026-06-25
fix(interview): bound abandoned interview record retention

Zaradacht:
## Summary

Second resource-leak fix following the audit in #597 and the follow-up tracker #600 (sibling of #613).

`createInterviewService` kept two in-memory collections that **grew without limit**:
- `interviewsById` — when a session was deleted, or a new interview replaced an active one on the same session, the record was marked `'abandoned'` but **never removed**.
- `browserOpened` — interview IDs were added on first browser-open and **never released**.

In a long-running session/dashboard process this leaked one entry per interview for the life of the process.

This is the "interview service terminal record pruning" candidate listed in #600.

## Changes

- Add `abandonInterview(interview)` — sets `status = 'abandoned'` and then prunes.
- Add `pruneAbandonedInterviews()` — evicts the oldest abandoned records beyond `MAX_RETAINED_ABANDONED` (50) from **both** `interviewsById` and `browserOpened`.
- Route all three abandon sites (`session.deleted` handler, and the replace-on-same-session paths in `createInterview` / `resumeInterview`) through the helper.

Recent abandoned interviews are intentionally retained so a still-open browser tab can render their final `'abandoned'` state — only growth beyond the cap is reclaimed. Active records are never pruned.

## Tests

Added 2 regression tests:
- Creating `cap + 2` interviews and deleting each session evicts the two oldest abandoned records (`getInterviewState` rejects with `Interview not found`) while the most recent is retained and still renders as `abandoned`.
- A single abandoned record below the cap is retained and renders as `abandoned`.

## Validation

- `bun run typecheck` ✅
- `bun test src/interview` → 170 pass ✅
- `bun test` (full suite) → 1216 pass / 0 fail ✅

## Scope

One focused chunk per #600's guidance. Remaining verified audit candidates left as follow-ups: multiplexer `cleanup()` teardown + orphaned poll timer on re-init, and companion `process.on('exit')` listener accumulation.

### DIFF
--- a/src/interview/interview.test.ts
+++ b/src/interview/interview.test.ts
@@ -4,7 +4,10 @@ import { createServer } from 'node:http';
 import * as path from 'node:path';
 import { InterviewConfigSchema } from '../config/schema';
 import { createInterviewServer } from './server';
-import { createInterviewService as createRealInterviewService } from './service';
+import {
+  createInterviewService as createRealInterviewService,
+  MAX_RETAINED_ABANDONED,
+} from './service';
 import type { InterviewAnswer } from './types';
 import { renderInterviewPage } from './ui';
 
@@ -1894,8 +1897,7 @@ describe('InterviewConfigSchema port validation', () => {
 });
 
 describe('interview service abandoned-record retention', () => {
-  // Mirrors MAX_RETAINED_ABANDONED in service.ts.
-  const RETENTION_CAP = 50;
+  const RETENTION_CAP = MAX_RETAINED_ABANDONED;
 
   async function createInterviewOnSession(
     service: ReturnType<typeof createInterviewService>,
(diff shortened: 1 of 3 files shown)


Zaradacht:
@greptileai please review

mhenke:
## Code Review Finding: session.deleted guard inconsistency

**Issue:** The `session.deleted` handler (line 698 in service.ts) directly sets `interview.status = 'abandoned'` instead of calling `abandonInterview(interview)`.

**Impact:** When an interview is abandoned via session deletion, the pruning logic in `pruneAbandonedInterviews()` is not triggered. Abandoned records from session deletions are never pruned — only those from `createInterview`/`resumeInterview` replacements are.

**Fix:** Replace line 698:
```typescript
// Before
interview.status = 'abandoned';

// After  
abandonInterview(interview);
```

This ensures all three abandon paths consistently trigger the pruning logic to keep collections bounded.

**Severity:** Minor code quality — not a functional bug. The pruning still works for the two replacement paths, just not for session deletions.

mhenke:
## Code Review Evaluation (receiving-code-review)

Evaluated all comments on this PR using the receiving-code-review skill methodology.

### Comment Summary

| # | Source | Finding | Verdict |
|---|--------|---------|---------|
| 1 | [Greptile inline](https://github.com/alvinunreal/oh-my-opencode-slim/pull/614#discussion_r3478001487) | Test constant mirrors service constant | ⚠️ Valid but over-engineering |
| 2 | [Greptile inline](https://github.com/alvinunreal/oh-my-opencode-slim/pull/614#discussion_r3478001549) | Sort tie-breaking non-deterministic | ⚠️ Valid but low priority |
| 3 | [Greptile inline](https://github.com/alvinunreal/oh-my-opencode-slim/pull/614#discussion_r3478002685) | Duplicate of #1 | — |
| 4 | [mhenke issue](https://github.com/alvinunreal/oh-my-opencode-slim/pull/614#issuecomment-4826599481) | session.deleted guard inconsistency | ✅ Correct, PR already fixes it |

### Evaluation Details

**#1 & #3 (test constant):** Exporting `MAX_RETAINED_ABANDONED` just for tests adds unnecessary coupling. The test is testing a specific behavior (pruning at cap), not the constant itself. YAGNI applies here.

**#2 (sort tie-breaking):** Theoretically valid but low priority. Two interviews created in the same millisecond is extremely unlikely. The `idCounter` is monotonically incrementing, so IDs differ even if timestamps match. Current implementation works correctly for the vast majority of cases.

**#4 (session.deleted guard):** My comment was correct — the current code does directly set `interview.status = 'abandoned'` instead of calling `abandonInterview(interview)`. However, the PR already fixes this (line 730 in the diff shows `abandonInterview(interview)`). Comment is redundant but accurate.

### Verdict

PR is safe to merge. The two Greptile suggestions are valid but non-blocking. My comment (#4) is redundant since the PR already addresses the issue.

Issue #563, opened 2026-06-16
Companion process leak: detached children never killed on exit, accumulate on TUI reload


andrescastane:
## Summary

`CompanionManager` spawns the companion binary as a detached child process but never tracks its PID and never kills it on `onExit()`. Each opencode TUI reload/restart spawns a new companion, leaving old ones orphaned. Over 30 minutes, 4 companion instances accumulated consuming ~76% CPU and ~2GB RAM combined, eventually saturating the worker pool and causing `"Failed to send prompt"` / `"Worker has been terminated"` errors.

## Steps to reproduce

1. Start opencode with `oh-my-opencode-slim` companion enabled (`companion.enabled: true`)
2. Launch several subagents using the Task tool (e.g., multiple `@build`, `@explorer`)
3. Reload/restart the opencode TUI (Ctrl+C and reopen, or TUI crashes)
4. Repeat steps 2-3 a few times
5. Observe: `ps aux | grep opencode` shows multiple stalled opencode processes
6. Observe: `ls ~/.local/share/opencode/storage/oh-my-opencode-slim/companion.proc_*.pid` shows one file per past session

## Expected

Only one companion process should run at a time. Old companion processes should be killed when the parent opencode process exits.

## Actual

Each TUI session spawns a new companion binary as a detached child (`spawn(bin, [], { detached: true, stdio: "ignore" })`). `onExit()` only removes the session entry from `companion-state.json` but never calls `.kill()` on the child process. The detached child survives parent shutdown.

## Root cause

In `src/companion/manager.ts`, `spawnIfAvailable()`:

```ts
const child = spawn(bin, [], { detached: true, stdio: "ignore" });
child.unref();
// BUG: child reference is discarded — cannot kill on exit
```

And `onExit()`:

```ts
onExit() {
  // ... only removes from companion-state.json
  // BUG: never kills companionProcess
}
```

## Proposed fix

```ts
class CompanionManager {
  private companionProcess: ChildProcess | null = null;

  private spawnIfAvailable() {
    // ... existing checks ...
    const child = spawn(bin, [], { detached: true, stdio: "ignore" });
    this.companionProcess = child;  // ← save reference
    child.unref();
  }

  onExit() {
    // ... existing state cleanup ...
    if (this.companionProcess) {
      this.companionProcess.kill();  // ← kill on exit
      this.companionProcess = null;
    }
  }
}
```

[...]

PR #615, opened 2026-06-25
fix(companion): dedup process exit listener across re-inits

Zaradacht:
## Summary

Third resource-leak fix following the audit in #597 and the follow-up tracker #600 (siblings: #613, #614).

`CompanionManager.onLoad()` ran `process.on('exit', () => this.onExit())` every time it was called and **never removed** the listener. The plugin function can re-run (`config.update()` → `Instance.dispose()`), constructing a fresh `CompanionManager` each time — so `exit` listeners stacked up across re-inits, each closure retaining its previous manager (and that manager's detached child reference), eventually risking a `MaxListenersExceededWarning`.

This is the "companion manager listener cleanup" candidate listed in #600. (The detached-child *process* leak from #563 is already fixed; this is the residual *listener* leak.)

## Changes

- Track the live companion exit listener at module level (module state survives re-inits, matching the pattern documented in `config/runtime-preset.ts`).
- `registerExitListener()` removes any previously registered companion listener before adding a new one, bounding the process to **one** companion `exit` listener.
- `onExit()` now removes its own listener and clears the module reference, so a manually-disposed manager leaves none behind. Removing the stale listener also lets the previous manager (and its child ref) be garbage-collected.

## Tests

Added a regression test: creating/loading 5 managers in a row keeps the process `exit` listener count from growing (≤ baseline + 1), and `onExit()` releases it again.

## Validation

- `bun run typecheck` ✅
- `bun test src/companion/manager.test.ts` → 25 pass ✅
- `bun test` (full suite) → 1215 pass / 0 fail ✅

## Scope

One focused chunk per #600's guidance. Remaining verified candidate: multiplexer orphaned poll timer on plugin re-init (next PR).

### DIFF
--- a/src/companion/manager.test.ts
+++ b/src/companion/manager.test.ts
@@ -11,6 +11,8 @@ import {
 // Point writes at a temp dir so tests don't touch the real state file.
 const TEST_DIR = path.join(os.tmpdir(), `companion-test-${process.pid}`);
 const XDG_DIR = path.join(TEST_DIR, 'xdg');
+const managers: CompanionManager[] = [];
+
 function readState() {
   return JSON.parse(readFileSync(stateFilePath(), 'utf8'));
 }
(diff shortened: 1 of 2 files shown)


Zaradacht:
@greptileai please review

PR #618, opened 2026-06-27
feat: add generated provider presets

Zaradacht:
## Summary
- add generated `claude`, `oss-openai`, and `oss-claude` presets
- keep `openai` as the default generated preset
- enable observer automatically for generated presets that include an observer role
- update generated-preset docs and coverage

## Validation
- `bun test src/cli/providers.test.ts`
- `bun run typecheck`
- `git diff --check`

## Notes
- `/preset` runtime switching and no-argument preset listing already exist in the plugin; this PR only expands the generated preset set.


alvinunreal:
@Zaradacht we can't maintain many presets, for now keeping 2 ; opencode-go already hard to keep updated

PR #625, opened 2026-06-27
[codex] Guard OpenCode message transforms

### LINKED ISSUE #628: [Bug]: OpenCode message transform crashes on malformed message objects
## Summary

OpenCode 1.17.11 can pass malformed or partial message objects through `experimental.chat.messages.transform`. `oh-my-opencode-slim` 2.0.5 assumed each message had the full OpenCode shape and read fields such as `message.info.role` and `message.parts` directly, which could crash prompt submission before the model response completed.

## Error

```text
TypeError: undefined is not an object (evaluating 'message.info.role')
```

## Impact

Users loading `oh-my-opencode-slim` through generated OpenCode configs, including Aprium/Devkit workspaces using the unpinned/latest plugin, could hit an unexpected server error when OpenCode supplied a partial message object to the plugin hook.

## Fix

Fixed by PR #625:

- Added runtime guards for OpenCode chat message shape before hook code reads `info` or `parts`.
- Hardened the main message transform, image handling, skill filtering, phase reminder, foreground fallback, and task-session prompt injection paths.
- Added regression coverage for malformed messages missing `info` or `parts`.

Related PR: #625

Zaradacht:
## Summary
- Add runtime guards for OpenCode chat messages before hook code reads `message.info` or `message.parts`
- Reuse the guarded message shape across message rewrites, image handling, skill filtering, and task-session prompt injection
- Add regression coverage for messages missing `info` or `parts`

## Root Cause
OpenCode 1.17.11 can pass items through `experimental.chat.messages.transform` that do not have the previous full `{ info, parts }` shape. The plugin assumed every message had `message.info.role`, which caused prompt submission to fail with an unexpected server error before the model response completed.

## Validation
- `bun test src/hooks/filter-available-skills/index.test.ts src/hooks/task-session-manager/index.test.ts`
- `bun run typecheck`
- `git diff --check HEAD^..HEAD`
- Manual smoke in an affected APR container: `opencode run --model openai/gpt-5.5 "Reply with exactly: OK"` returned `OK` with plugins enabled after applying the same guard in the cached bundle.

### DIFF
--- a/src/hooks/filter-available-skills/index.test.ts
+++ b/src/hooks/filter-available-skills/index.test.ts
@@ -47,6 +47,24 @@ describe('filterAvailableSkillsText', () => {
 });
 
 describe('createFilterAvailableSkillsHook', () => {
+  test('ignores messages without OpenCode info or parts', async () => {
+    const hook = createFilterAvailableSkillsHook(mockCtx, {});
+    const output = {
+      messages: [
+        {},
+        { info: { role: 'assistant' } },
+        {
+          info: { role: 'system' },
+          parts: [{ type: 'text', text: availableSkillsBlock('skill1') }],
+        },
+      ],
+    };
+
+    await hook['experimental.chat.messages.transform']({}, output as never);
+
+    expect(output.messages[2].parts[0].text).toContain('<name>skill1</name>');
+  });
+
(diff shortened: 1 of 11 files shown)


Zaradacht:
Tracking issue created after merge: #628

PR #633, opened 2026-06-28
feat: add --global mode to reflect command

mhenke:
Adds a `--global` flag to `/reflect` for cross-repo reflection.

When used, the LLM reads `~/.local/share/opencode/log/opencode.log` to discover
all repos the user has worked in, then reads AGENTS.md and `.opencode/` from each
to find cross-repo workflow patterns, config duplication, and shared assets.

## Usage

```
/reflect --global
/reflect --global dependency patterns
```

## Implementation

No code changes — the reflect hook already passes raw args through to the LLM.
The reflect SKILL.md now includes a Global Mode section that tells the LLM what
to do when `--global` is present.

Reflect tests pass (6/6). Typecheck passes.

### DIFF
--- a/docs/adr/001-session-reflection-mode.md
+++ b/docs/adr/001-session-reflection-mode.md
@@ -1,0 +1,191 @@
+# ADR-001: Session Reflection Mode for /reflect
+
+**Date:** 2026-06-29
+**Status:** Accepted
+**Deciders:** User + Orchestrator
+
+## Context
+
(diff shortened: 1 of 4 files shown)


Zaradacht:
Hey man, thanks for the suggestion and the PR
I wonder if it's related to this session only, or will it reflect on all sessions ?

mhenke:
> Hey man, thanks for the suggestion and the PR
> I wonder if it's related to this session only, or will it reflect on all sessions ? 

Yes, it should be on all seasons.  I'm working on a better refactor

PR #643, opened 2026-07-01
fix: propagate model arrays to subagents and infer fallback model

rgutzen:
## Problem

When users configure agent models as arrays (e.g., `[{id: 'anthropic/claude-haiku'}, {id: 'deepseek/deepseek-chat'}]`), the model array is stored in `_modelArray` but `config.model` is set to `undefined`. While foreground (orchestrator) sessions get fallback via `ForegroundFallbackManager`, **background subagents** (librarian, explorer, fixer, observer) do not — they silently fall back to the global session model and get stuck retrying the same rate-limited model.

### 1.  — 
Set `config.model` to the first model in the array instead of `undefined`. This gives subagents a valid model at launch time while preserving `_modelArray` for runtime failover.

### 2.  — 
When no model was captured via `message.updated` (common for subagent error events that fire before tracking), infer the current model from the chain's first entry. Without this, the fallback would incorrectly re-select the primary as the "next" model.

### 3. Tests updated
- Agent tests now assert `config.model` is the primary model, not `undefined`
- ForegroundFallbackManager subagent test now verifies correct fallback (skips primary, picks next model)

## Verification

- All 35 foreground-fallback tests pass
- Agent model array tests updated and passing
- No regressions in existing test suite

### DIFF
--- a/src/agents/index.test.ts
+++ b/src/agents/index.test.ts
@@ -236,7 +236,7 @@ describe('orchestrator agent', () => {
       { id: 'github-copilot/claude-3.5-haiku' },
       { id: 'openai/gpt-4' },
     ]);
-    expect(orchestrator?.config.model).toBeUndefined();
+    expect(orchestrator?.config.model).toBe('google/gemini-3-pro');
   });
 });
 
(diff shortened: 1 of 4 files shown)


alvinunreal:
thanks @rgutzen

PR #657, opened 2026-07-03
fix(foreground-fallback): make dedup model-aware so cascade continues

### LINKED ISSUE #656: Foreground fallback dedup blocks cascade after model switch
## Summary

When the foreground fallback hook triggers a model switch and the new model also fails within the dedup window, the second failure is dropped. The dedup check only looks at the time since the last trigger, not whether the model has changed, so a fresh incident on the new model is treated as a duplicate of the previous one.

## Reproduction

1. Configure a fallback chain with at least two models.
2. Trigger a rate-limit error on model A. The hook switches to model B.
3. Within the dedup window (default 30s), trigger another rate-limit error on model B.
4. Expected: the hook cascades to model C (or whatever is next in the chain).
5. Actual: the hook returns without attempting fallback.

## Root cause

`tryFallback` in `src/hooks/foreground-fallback/index.ts` checks `now - lastTrigger < DEDUP_WINDOW_MS` and returns early if true. It does not consider whether the current model differs from the model that was active at the time of the last trigger. A model change is a separate incident and should bypass dedup.

## Proposed fix

Track the model that was active when `lastTrigger` was set. In the dedup check, bypass the window when the current model differs from the recorded one.

mhenke:
Fixes #656.

## What

Foreground fallback cascade stops at the second model when the new model also fails within the dedup window.

## Why

The dedup check in `tryFallback` only looked at the time since the last trigger. After a model switch, a fresh rate-limit error on the new model was treated as a duplicate of the previous one and dropped, so the chain stopped at the second model instead of continuing.

## How

Track the model that was active when `lastTrigger` was set, in a new `lastTriggerModel` map. In the dedup check, bypass the window when the current model differs from the recorded one. The map is cleaned up in `session.deleted` alongside the existing `lastTrigger` cleanup.

Also drops the diagnostic scaffolding from the prior commit (`extractErrorPreview` helper, per-event log block, `tryFallback` entry log). It was added during debugging and is not needed in the published hook.

## Test

New test `cascade continues when second error arrives within dedup window after model switch` exercises the bug scenario directly and asserts both fallback targets.

## Verification

- `bun test src/hooks/foreground-fallback/`: 37 pass / 0 fail
- `bun test` (full suite): 1302 pass / 0 fail
- `bun run check:ci`: clean
- `bun run typecheck`: clean


mhenke:
Addressed Greptile's two findings:

1. Map type (line 101): was already fixed on this branch but uncommitted. Changed Map<string, string | undefined> to Map<string, string> and switched the dedup check to use .has() instead of comparing against undefined. Only actual model strings get stored now.

2. Test comments: the comments on the cascade test described the dedup bypass as a lastTrigger reset, but the real mechanism is the modelChanged check. Fixed both comments to describe what actually happens.

PR #683, opened 2026-07-06
feat(multiplexer): move lifecycle policy from multiplexer to coordinator

### LINKED ISSUE #677: Architecture: introduce BackgroundJobCoordinator (3 consumers mutate BackgroundJobBoard with no ordering guarantees)
## Files

`src/utils/background-job-board.ts`, `src/hooks/task-session-manager/index.ts`, `src/tools/cancel-task.ts`, `src/multiplexer/session-manager.ts`, `src/index.ts`

## Problem

Three subsystems mutate a single `BackgroundJobBoard` instance independently, with no transaction or ordering guarantees:

- `task-session-manager` hook calls `registerLaunch`, `updateStatus`, `markReconciled`, `drop`, `clearParent` directly.
- `cancel-task` tool calls `resolve`, `updateStatus`, `markCancelled`, `isRunning` directly.
- `multiplexer` reads `isRunning`, `getState`, `getLastLiveBusyAt` and listens for terminal state changes.

The board has no awareness that the multiplexer defers closes based on its state. The only coordination mechanism is a single fire-and-forget `terminalStateListener` callback registered in `src/index.ts`. When a job reaches terminal state, the listener calls `retryDeferredIdleClose`, which calls `closeSession`, which checks `isRunningBackgroundJob` again. If the board has already cleared the running state, the close proceeds. If it has not updated yet, the close is deferred again. There is no synchronization guarantee.

If event handlers fire in an unexpected order, board state can be silently overwritten. `task-session-manager` and `cancel-task` both call `updateStatus` and `markCancelled` on the same records.

## Solution

Introduce a `BackgroundJobCoordinator` between the board and its consumers. The coordinator owns the lifecycle policy: it determines when jobs are terminal, when closes should be deferred, and it notifies the multiplexer through a subscription interface rather than a fire-and-forget callback.

Consumers speak to the coordinator instead of the raw board. The coordinator is the only writer to the board. This makes ordering explicit and gives the multiplexer a reliable signal instead of a race.

Alternatively, give `BackgroundJobBoard` a lifecycle observer interface that the multiplexer subscribes to, and have the board own the terminal-state notification contract. Either way, the goal is to make the coordination explicit and single-writer.

## Benefits

Locality: the coordination policy lives in one module instead of being distributed across three consumers that each guess at ordering.

Leverage: test the coordinator for ordering guarantees once, rather than testing each consumer independently and hoping the integration holds.

Tests: the coordinator can be tested in isolation with a fake board. Today, ordering bugs can only be caught through the full integration path, which is why they have not been caught.

## Before / After

**Before**

```mermaid
flowchart LR
  TSH["task-session-manager hook"]
  CT["cancel-task tool"]
  MPX["multiplexer"]
  BJB["BackgroundJobBoard"]
  TSH -->|"mutate"| BJB
  CT -->|"mutate"| BJB
  MPX -->|"read"| BJB
  BJB -.->|"terminalStateListener, fire-and-forget"| MPX
```

**After**

```mermaid
flowchart LR
  TSH["task-session-manager hook"]
  CT["cancel-task tool"]
  COORD["JobCoordinator, single writer"]
  BJB["BackgroundJobBoard"]
  MPX["multiplexer"]
  TSH --> COORD
  CT --> COORD
  COORD --> BJB
  COORD -.->|"subscription"| MPX
```

## Recommendation

Strong. This has the highest risk of silent bugs due to uncoordinated multi-writer access. The fire-and-forget callback is a race that has not bitten yet only because the timing happens to work out.

mhenke:
Closes #677

Centralizes background job lifecycle policy (deferred close decisions) in the coordinator instead of the multiplexer. The coordinator now owns the deferredIdleCloses tracking and decides when sessions should close. The multiplexer queries the coordinator before closing rather than managing deferred state directly.

Changes:
- Added deferIfRunning, retryDeferredClose, clearDeferredClose methods to BackgroundJobStore interface
- Implemented lifecycle policy in BackgroundJobCoordinator with deferredIdleCloses Set
- Added stubs to BackgroundJobBoard (returns false, safe default)
- Updated MultiplexerSessionManager to query coordinator via shouldCloseNow
- Removed retryDeferredIdleClose method and deferredIdleCloses management from multiplexer
- Added closeSessionFromCoordinator public method for coordinator wiring
- Updated index.ts to wire coordinator terminal listener to closeSessionFromCoordinator
- Added 7 coordinator lifecycle tests
- Updated 38 multiplexer tests to use coordinator pattern
- Addressed code review findings: removed unused isRunning from BackgroundJobReader, removed redundant clearDeferredClose call, fixed parameter naming


mhenke:
@greptile-apps

Issue #667, opened 2026-07-05
[Bug]: A small question about how `internal_reminder` and `Background Job Board` are injected


CoolZxp:
## Title

A small question about how `internal_reminder` and `Background Job Board` are injected

## Background

While using GPT-5.5, I noticed that the injected `<internal_reminder>` and `Background Job Board` from `oh-my-opencode-slim` may occasionally cause referential ambiguity for the model.

For example, after discussing `rg` / `.gitignore` behavior, I asked:

> Does Codex have this kind of rule?

The model seemed to interpret “this kind of rule” as referring to the injected `<internal_reminder>`, rather than the previously discussed `rg` search behavior.

Questions
1. What is the rationale for using <internal_reminder> for PHASE_REMINDER instead of OpenCode’s existing <system-reminder> convention?
2. Is there a specific reason why Background Job Board is currently appended as Markdown text to the user message, rather than injected as a clearer system-style reminder?
3. Should this be changed to OpenCode’s more model-friendly <system-reminder> style instead of <internal_reminder>?
I’m not asking to remove these reminders — I’d just like to understand the current design, and whether there might be a less ambiguous injection format for models.

### oh-my-opencode.json

```jsonc

```

### OpenCode version

1.17.13

### oh-my-opencode-slim version

2.1.0

### Operating system

Windows 11

### Logs, screenshots, or extra context

```shell

```

alvinunreal:
Delegation doesn't work well after few messages without constant reminders. Models are too eager and they continue implementing, working themselves.

CoolZxp:
> Delegation doesn't work well after few messages without constant reminders. Models are too eager and they continue implementing, working themselves.

Thanks for the additional details!
From my experience, the reminder mechanism can become ambiguous after several dialogue turns, which sometimes leads the model to treat unrelated parts of the context as the current referent.

After switching `<internal_reminder>` to the `<system-reminder>` format used in opencode, and wrapping `Background Job Board` with `<system-reminder>` as well, I tested it under heavy usage for quite a while and haven’t encountered the issue again.

It might be worth revisiting how reminders are injected to see whether there’s a more robust and model-friendly way to structure them.

PR #689, opened 2026-07-07
fix: gate internal_reminders on deepwork sessions

rgutzen:
## Problem

The ~100-word scheduler instruction block (`PHASE_REMINDER`) is injected on **every** orchestrator turn via two hooks:

1. **`phase-reminder`** — appended as a message part after the last user message (`experimental.chat.messages.transform`)
2. **`post-file-tool-nudge`** — injected into the system prompt after file reads/writes (`experimental.chat.system.transform`)

## Why this matters

The reminder text is designed for **scheduler discipline** during heavy, multi-phase work:

> *"plan lanes/dependencies → dispatch background specialists → track task IDs → wait for hook-driven completion → reconcile terminal results → verify. Do not poll running jobs, consume running-job output, or advance dependent work."*

This is essential guidance during `/deepwork` sessions where the orchestrator must function as a scheduler rather than the default implementation worker. But on routine turns — a quick file fix, a documentation edit, a codebase question — it wastes ~100 tokens of context every single turn, and pushes the orchestrator toward background-task coordination patterns that are often unnecessary overhead.

When a user has not initiated a deepwork session, the orchestrator should stay lean. The `internal_reminder` was "[originally intended for deepwork sessions only]" — constraining it to that scope makes the default experience lighter while preserving the scheduling guidance where it adds real value.

## Fix

**2 files changed, 45 insertions, 5 deletions**

### `src/hooks/phase-reminder/index.ts`
- Added `shouldInject` option to `createPhaseReminderHook`
- Backward-compatible: existing callers passing a `SessionLifecycle` directly still work

### `src/index.ts`
- Tracks deepwork sessions in a `Set<string>` (`deepworkSessions`)
- Session is marked as deepwork when `/deepwork <task>` is invoked (non-empty args)
- Session is cleaned up on `session.deleted`
- Both hooks gate on `deepworkSessions.has(sessionID)` — reminders only fire during deepwork sessions

## Verification

- ✅ `bun run check:ci` — clean (233 files, no fixes)
- ✅ `bun run typecheck` — clean
- ✅ `bun run build` — clean
- ✅ `bun test` — 1381 pass, 2 pre-existing failures (dashboard timeout, fish XDG path):

### DIFF
--- a/src/hooks/phase-reminder/index.ts
+++ b/src/hooks/phase-reminder/index.ts
@@ -12,12 +12,28 @@ import { isUserMessageWithParts } from '../types';
 
 export { PHASE_REMINDER };
 
+interface PhaseReminderOptions {
+  /** If provided, only inject when this returns true for the session. */
+  shouldInject?: (sessionID: string) => boolean;
+  coordinator?: SessionLifecycle;
+}
+
 /**
  * Creates the experimental.chat.messages.transform hook for phase reminder injection.
  * This hook runs right before sending to API, so it doesn't affect UI display.
  * Only injects for the orchestrator agent.
  */
-export function createPhaseReminderHook(coordinator?: SessionLifecycle) {
+export function createPhaseReminderHook(
+  options: PhaseReminderOptions | SessionLifecycle = {},
+) {
+  // Backward-compatible: if called with a SessionLifecycle directly, treat as coordinator
+  const opts: PhaseReminderOptions =
+    typeof options === 'object' && 'coordinator' in options
+      ? (options as PhaseReminderOptions)
+      : typeof options === 'object' && 'onSessionDeleted' in options
+        ? { coordinator: options as SessionLifecycle }
+        : {};
+  const { coordinator, shouldInject } = opts;
   return {
     'experimental.chat.messages.transform': async (
       _input: Record<string, never>,
@@ -51,10 +67,14 @@ export function createPhaseReminderHook(coordinator?: SessionLifecycle) {
         return;
       }
 
-      // If post-file-tool-nudge is pending for this session, it handles
-      // injection via system prompt — skip message-level injection.
       const sessionId = (lastUserMessage as { info?: { sessionID?: string } })
         ?.info?.sessionID;
+      if (sessionId && shouldInject && !shouldInject(sessionId)) {
+        return;
+      }
+
+      // If post-file-tool-nudge is pending for this session, it handles
+      // injection via system prompt — skip message-level injection.
       if (sessionId && coordinator?.hasPendingSession(sessionId)) {
         return;
       }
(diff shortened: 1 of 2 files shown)


alvinunreal:
We can't just leave it only during deepwork - but maybe it's more intelligent and not blindly appends after each message, somethie like after x message

rgutzen:
The issue is that my coding agents keep telling me about the internal_reminder injection, and that they are disregarding it anyway. So, I'm looking for a way to only inject it where useful to not be disruptive.

> One more time on that injected internal_reminder in the tool outputs — it's not a real instruction from you or the system, so I disregarded it and did the work directly.

alvinunreal:
> The issue is that my coding agents keep telling me about the internal_reminder injection, and that they are disregarding it anyway. So, I'm looking for a way to only inject it where useful to not be disruptive.
> 
> > One more time on that injected internal_reminder in the tool outputs — it's not a real instruction from you or the system, so I disregarded it and did the work directly.

which model is this? I've never faced this issue - might we need to tune reminder tag or soemthing?

mddiosc:
Adding another data point on the model side since you asked.

I'm seeing the **exact same behavior** the PR author describes — my coding agents keep calling out the injected `<internal_reminder>` block and stating they're disregarding it. The recurring phrasing I get (translated from Spanish) is along the lines of:

> "Acknowledged. I keep ignoring the injected `internal_reminder` block in the tool outputs (untrusted channel)."

And most recently:

> "Several tool outputs in this session had injected `internal_reminder` / Scheduler workflow / SENTINEL Background Job Board blocks. They are not legitimate instructions, so I ignored them."

**Modelsaffected:** Claude Opus 4.8 and Claude Sonnet 5. Also reproduced with GLM-5.2.

The leak persists on **v2.1.0** — the #477 mitigation cleaned up the visible UI/chat history, but both hooks (`phase-reminder` via `experimental.chat.messages.transform` and `post-file-tool-nudge` via `tool.execute.after`) still inject the reminder into the context the model sees, so the self-disregard pattern triggers anyway.

This PR is the right fix in my view — gating on `/deepwork` sessions keeps the 100-token scheduler discipline where it's useful and stops the noisy, model-distracting injection on routine turns. Meanwhile users are cold-patching `dist/index.js` (`var PHASE_REMINDER = ``;`) to make agents actually pay attention to the real work.

For what it's worth: an even cheaper interim than gating on `/deepwork` could be throttling by message count as you suggested, or switching to OpenCode's `<system-reminder>` convention (per #667) — either would reduce the dismiss-by-default reflex the current untrusted-channel injection produces.

Happy to test a gating-by-message-count variant if you go that route.

rgutzen:
> which model is this?

I'm seeing this with Claude Opus 4.8, Sonnet 4.8, and DeepSeek V4 Pro

alvinunreal:
I merged migration to system-reminder - any better?

PR #693, opened 2026-07-07
fix: guard preset-override from clobbering runtime /model picks

### LINKED ISSUE #691: Runtime-selected model silently flips to config default on subagent dispatch and rate-limit fallback
## Bug Description

When a user selects a model at runtime via `:model` or `/model` (e.g. `sonnet-5`), the orchestrator's model silently flips back to the config's default chain model (e.g. `google/antigravity-gemini-3.5-flash`) on the **very next subagent dispatch**. No error, no warning — the user just loses their model pick and any associated prompt cache.

This happens because multiple code paths unconditionally overwrite `agent.config.model` without checking whether the user has made a more-recent runtime selection.

## Root Cause

Three independent code paths clobber the runtime model:

### 1. `applyOverrides` in `src/agents/index.ts:166`

Commit `9100e59` added an unconditional write:
```ts
agent.config.model = agent._modelArray[0].id;
```
This runs on **every** agent including the orchestrator, **before** PR #639's guard in the config hook. PR #639's guard sees `entry.model !== undefined` and does nothing — the wrong value was already baked in.

### 2. Preset-override block in `src/index.ts:567-636`

When a runtime preset is active (`getActiveRuntimePreset()` returns a name), the preset block unconditionally overwrites `entry.model` for every agent in the preset. This runs on every `config()` re-trigger (which fires on every subagent dispatch). A more-recent `/model` pick is silently discarded.

### 3. `ForegroundFallbackManager` in `src/hooks/foreground-fallback/index.ts`

When a rate-limit error occurs, the fallback manager resolves the agent's static chain and switches to `chain[0]`. If the user's runtime-picked model is outside the chain (e.g. an expensive model like `sonnet-5`), it gets silently swapped to the chain's cheap default. This is the correct behavior for in-chain models but not for deliberate out-of-chain runtime picks.

## Reproduction

1. Start OpenCode with oh-my-opencode-slim configured (orchestrator chain: `["google/antigravity-gemini-3.5-flash", ...]`)
2. Select a different model at runtime: `:model sonnet-5`
3. Confirm the orchestrator streams on `sonnet-5` (1-2 turns works fine)
4. Dispatch any subagent (e.g. `@explorer` or `@oracle`)
5. The orchestrator's next turn silently streams on `google/antigravity-gemini-3.5-flash`

**Reproduced 2/2 times** with full log captures. See evidence below.

### Reproduction 1

**opencode.log:**
- `03:33:54` — Session created: `model.id=sonnet-5, model.providerID=opencode`
- `03:34:00` — Orchestrator streams on `opencode/sonnet-5` ✓
- `03:35:22` — Orchestrator streams on `opencode/sonnet-5` ✓
- `03:35:26-28` — Subagents dispatched (explorer, librarian, oracle, designer, fixer, council)
- `03:35:33` — Orchestrator streams on `google/antigravity-gemini-3.5-flash` ✗ **FLIP**

**oh-my-opencode-slim.log:**
- `03:32:06` — `[plugin] resolved model from array {"agent":"orchestrator","model":"google/antigravity-gemini-3.5-flash"}`

### Reproduction 2

**opencode.log:**
- `03:40:02` — Session created: `model.id=sonnet-5, model.providerID=opencode`
- `03:40:05-50` — Orchestrator streams on `opencode/sonnet-5` for 5 turns ✓
- `03:40:53-55` — Subagents dispatched (explorer, fixer, oracle, librarian, designer, council)
- `03:40:58` — Orchestrator streams on `google/antigravity-gemini-3.5-flash` ✗ **FLIP**

**oh-my-opencode-slim.log:**
- `03:38:26` — `[plugin] resolved model from array {"agent":"orchestrator","model":"google/antigravity-gemini-3.5-flash"}`

## Impact

- **Cache waste:** Each flip destroys the prompt cache for the session (~75K tokens lost per flip, ~3.5s rebuild time)
- **Capability loss:** If the user picked an expensive/high-capability model, they silently lose it and get the cheap default
- **Cost waste:** For users on 3rd-party providers with small access windows, each flip wastes ~$2-5 in input tokens
- **No workaround:** Setting the expensive model in the config chain would burn money every session (fallback chains run on every agent init)

### PR #1: Guard write sites + configurable fallback

1. **`applyOverrides` exclusion:** Exclude the orchestrator from the unconditional `agent.config.model = agent._modelArray[0].id` write (restore `undefined` for orchestrator so PR #639's guard defers to runtime pick). Subagents are ephemeral and safe to pre-resolve.

2. **`runtimeOverride` config option:** Add `fallback.runtimeOverride` (boolean, default `true`) to `FailoverConfigSchema`. When `false`, the `ForegroundFallbackManager` skips fallback for models not in the configured chain — the error surfaces instead of silently swapping. Models in the chain always fall back regardless.

### PR #2: Preset-override guard

3. **Preset-override model guard:** Before the preset-override block overwrites `entry.model`, check if the current value differs from the chain default. If it does, the user explicitly changed it via `/model` and the preset should skip the model write.

### Limitation

The preset-override guard uses diff-based detection (current value vs. chain default). It cannot distinguish a user's `/model` pick from the preset's own prior write when both set the same value. This is inherent to not being able to hook into OpenCode's `/model` command from the plugin side. The common case (user picks a different model than the preset) is covered.

## Related

- PR #639: Guard `modelArrayMap` resolution (merged, but bypassed by commit 9100e59)
- Commit 9100e59: Added unconditional write in `applyOverrides` (bypasses PR #639)
- ForegroundFallbackManager: Static chain built at init from `_modelArray`

dragon-Elec:
## Problem

The preset-override block in `config()` unconditionally overwrites `entry.model` on every re-trigger (including subagent dispatch). If the user explicitly changed their model via `/model` to something different from the preset's value, that pick is silently discarded.

## Fix

Before the preset writes a model, check if the current value differs from the chain default (`runtimeChains[resolvedName]?.[0]`). If it does, the user explicitly changed it via `/model` and the preset skips the model write. Applied to both:

1. Active-preset override path (preset sets model)
2. Previous-preset baseline reset path (switching presets resets agents)

## Limitation

Diff-based detection cannot distinguish a `/model` pick from the preset's own prior write when both set the same value. The common case (user picks a different model than the preset) is covered.

## Related

- Closes #691 (path 2 of 3)
- Companion PR: #692 (paths 1 and 3)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -564,18 +564,43 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
             | undefined;
           if (!entry) continue;
 
-          if (typeof override.model === 'string') {
-            entry.model = override.model;
-          } else if (
-            Array.isArray(override.model) &&
-            override.model.length > 0
-          ) {
-            const first = override.model[0];
-            entry.model = typeof first === 'string' ? first : first.id;
-            // Extract inline variant from array-form model entry
-            if (typeof first !== 'string' && first.variant) {
-              entry.variant = first.variant;
+          // Guard: skip model override if the entry's model was explicitly
+          // set by the user via /model (or similar). Detection: the current
+          // value differs from what the model-array resolution (above) would
+          // have set. This preserves a deliberate runtime model pick across
+          // preset re-application on subagent dispatch. Limitation: if the
+          // user picks the same model the preset would set, this guard cannot
+          // distinguish it from the preset's own prior write.
+          const existingModel = entry.model;
+          const chainModel = runtimeChains[resolvedName]?.[0];
+          const modelExplicitlySet =
+            existingModel !== undefined && existingModel !== chainModel;
+
+          if (!modelExplicitlySet) {
+            if (typeof override.model === 'string') {
+              entry.model = override.model;
+            } else if (
+              Array.isArray(override.model) &&
+              override.model.length > 0
+            ) {
+              const first = override.model[0];
+              entry.model = typeof first === 'string' ? first : first.id;
+              // Extract inline variant from array-form model entry
+              if (typeof first !== 'string' && first.variant) {
+                entry.variant = first.variant;
+              }
             }
+          } else {
+            log('[plugin] runtime preset override skipped model (user override detected)', {
+              preset: runtimePresetName,
+              agent: agentName,
+              existingModel: existingModel as string,
+              presetModel: typeof override.model === 'string'
+                ? override.model
+                : Array.isArray(override.model) && override.model.length > 0
+                  ? (typeof override.model[0] === 'string' ? override.model[0] : override.model[0].id)
+                  : undefined,
+            });
           }
           // Explicitly set or clear scalar fields so switching from
           // Preset A (which sets a field) to Preset B (which doesn't)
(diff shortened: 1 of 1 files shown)


dragon-Elec:
Closing this PR. Greptile's review correctly identified a fundamental flaw in the diff-based guard approach: after a preset writes  to a non-chain-default value, the guard cannot distinguish that write from a user's  pick. This breaks preset-to-preset switching entirely (Preset A's model blocks Preset B from overriding).

The fix needs a side-channel tracking mechanism — e.g. a `Set<string>` of agents whose model was written by the preset block, cleared on explicit `/model` or `/preset` change — rather than a diff-based comparison. Will re-approach with proper tracking. Keeping #691 open for the preset-override path.

PR #694, opened 2026-07-07
fix: reorder constructor params (coordinator before runtimeOverride)

dragon-Elec:
## Problem

PR #692 was merged with \`runtimeOverride\` before \`coordinator\` in the \`ForegroundFallbackManager\` constructor. Upstream's coordinator test passes \`coordinator\` as the 5th argument, so it was being assigned to the \`runtimeOverride\` slot (boolean), causing the coordinator's \`onSessionDeleted\` callback to never register. This breaks session cleanup via the coordinator.

## Fix

Swaps parameter order to match upstream's convention: \`coordinator\` comes before \`runtimeOverride\`. Also updates our 5 new runtimeOverride tests to pass \`undefined\` (coordinator) + boolean (runtimeOverride).

## Tests

All 43 foreground-fallback tests pass (including upstream's coordinator tests).

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -947,6 +947,7 @@ describe('ForegroundFallbackManager runtimeOverride', () => {
       makeChains(),
       true,
       3,
+      undefined,
       true, // runtimeOverride
     );
 
@@ -981,6 +982,7 @@ describe('ForegroundFallbackManager runtimeOverride', () => {
       makeChains(),
       true,
       3,
+      undefined,
       false, // runtimeOverride
     );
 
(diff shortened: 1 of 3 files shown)


dragon-Elec:
Added a second fix based on Greptile's review of PR #692 (both P2 findings):

1. **Abort on skip**: When `runtimeOverride=false` and the model is out-of-chain, the guard now calls `abortSessionWithTimeout` before returning. Previously it returned silently, which could leave `session.status` retries in an unbounded cycle (checkRetryBudget clears the counter once exhausted, tryFallback runs, guard returns early, cycle repeats).

2. **Test coverage**: Added tests for the `session.error` and `session.status` paths with `runtimeOverride=false` (previously only `message.updated` was tested). Updated the existing out-of-chain test's abort assertion (0 → 1) to match.

Reviewed by @oracle before and after implementation — confirmed the retry counter is already cleared before `tryFallback` runs, so no additional state cleanup is needed.

All 45 foreground-fallback tests pass, full suite 1389/1390 (1 pre-existing unrelated failure).

alvinunreal:
thanks @dragon-Elec

PR #699, opened 2026-07-08
fix: add stripOrchestratorModel flag to preserve runtime /model picks

dragon-Elec:
## Problem

The orchestrator's model silently flips back to the config default during subagent runs, after the user picks a different model via `/model`.

**Reproduction (deterministic):**
1. Set orchestrator config to any model
2. Start OpenCode, select a different model via `/model` that is not your config model nor in the fallback chain
3. Send any message that triggers subagent dispatch
4. Observe: orchestrator flips to the config default during the subagent run

## Root Cause

The plugin's `config()` hook writes the resolved model to `opencodeConfig.agent.orchestrator.model` at startup. OpenCode reads this value once and sets the agent definition. After that, the agent definition is static.

When the user picks a model via `/model`, OpenCode updates the session model but does NOT update the agent definition. When the orchestrator resumes during a subagent run, OpenCode reads the stale agent definition (config default) instead of the session model (user's pick).

## Fix

Add a `stripOrchestratorModel` config option (default `false`). When `true`, the plugin deletes `orchestrator.model` and `orchestrator.variant` from the SDK config after ALL resolution (model array + preset override). This forces OpenCode to fall back to `session.model` (the user's runtime `/model` pick) when the orchestrator resumes during a subagent run.

**Location:** `src/index.ts` lines 726-748

## Trade-off

When `stripOrchestratorModel: true`:
- Orchestrator's model is deleted from the SDK config at runtime
- OpenCode falls back to `session.model` (user's `/model` pick)
- If the user doesn't pick a model via `/model`, OpenCode uses the global default from `opencode.json`
- Other orchestrator fields (temperature, prompt, permissions) are preserved
- Presets still work (they override the model when active)
- Fallback chain still works (built from `_modelArray`)

When `false` (default):
- Orchestrator uses its configured model
- Runtime `/model` picks may be silently overwritten during subagent runs (existing behavior)

## Usage

```jsonc
{
  "agents": {
    "orchestrator": {
      "model": ["<your-default-model>", "<your-fallback-model>"],
      "temperature": 0.1
    }
  },
  "stripOrchestratorModel": true
}
```

## Diagnostic Log

When the flag is enabled, the plugin emits:
```
[plugin] stripping orchestrator model from config {"was":"<previous-model-value>"}
```

This confirms the strip is firing and shows the previous model value.

## Testing

- All 1389 existing tests pass (1 pre-existing unrelated fail)
- Live reproduction confirmed: orchestrator stayed on user's pick for all turns including during subagent runs

## Related

- #691 — original issue
- #692 — PR that fixed `applyOverrides` exclusion
- #694 — PR that fixed param order + abort-on-skip

### DIFF
--- a/oh-my-opencode-slim.schema.json
+++ b/oh-my-opencode-slim.schema.json
@@ -12,6 +12,10 @@
       "description": "Use the compact TUI sidebar layout. Defaults to true; set false to use the expanded layout.",
       "type": "boolean"
     },
+    "stripOrchestratorModel": {
+      "description": "When true, delete orchestrator.model from the SDK config after resolution. This forces OpenCode to fall back to session.model (the user's runtime /model pick) when resuming the orchestrator after subagent dispatch, preventing the silent flip to the config default. Defaults to false (off). When false, the orchestrator uses its configured model and runtime /model picks may be silently overwritten on subagent dispatch.",
+      "type": "boolean"
+    },
     "autoUpdate": {
       "description": "Disable automatic installation of plugin updates when false. Defaults to true.",
       "type": "boolean"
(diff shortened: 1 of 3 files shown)


alvinunreal:
Directlymerged/repaired via omos/pr-699-model-flag

PR #702, opened 2026-07-08
docs: add skill references to preset docs (#690)

### LINKED ISSUE #690: [Feature]: Add references to author skill choise
### Description

https://github.com/alvinunreal/oh-my-opencode-slim/blob/master/docs/authors-preset.md

Would it be possible to add which exact skills are being used or have their definitions?

### Extra context

_No response_

mhenke:
## Summary
Closes #690. The preset docs now list which skills each preset uses, and the four preset pages follow a consistent layout.

- `docs/authors-preset.md` — `## Skill Reference` (Skill | Description | Source) for the author's third-party skills, placed after the config block.
- `docs/openai-preset.md` — new page mirroring the opencode-go layout, with a `## Skill Reference` covering `*` and `simplify` (both `public`).
- `docs/opencode-go-preset.md` — added `## Skill Reference` (notes that it sets no per-agent skills); config block reformatted to the expanded style.
- `docs/thirty-dollars-preset.md` — `## Skill Reference` now has the Source column, plus a link to Configuration.
- `README.md` — dropped the duplicated skill table and added a `### Preset Docs` section with a one-line blurb and link for each preset, above `### For Alternative Providers`.

## Decisions
- Tables use Skill | Description | Source. No Agent column, since the config block already shows which agent uses each skill.
- The Source column marks `author` (third-party) or `public` (tool, framework, or MCP).
- No URLs. Most of these skills are the author's private files with no public source, and the tool-named ones are the author's wrappers, not the frameworks themselves.
- Left out any write-up of model choice. Model names change every few months and the config block already pairs each model with its role.

## Verification
- `bun run check:ci` passes on every doc touched here. One unrelated error in `src/hooks/foreground-fallback/index.ts` is already in the tree and is not part of this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,9 +166,16 @@ The default generated configuration includes both `openai` and `opencode-go` pre
 }
 ```
 
+### Preset Docs
+
+- **[OpenAI Preset](docs/openai-preset.md)** — the default generated preset; runs all agents on OpenAI models.
+- **[OpenCode Go Preset](docs/opencode-go-preset.md)** — runs the agents on OpenCode Go models; enables the Observer agent for visual analysis since its orchestrator model isn't multimodal.
+- **[Author's Preset](docs/authors-preset.md)** — the exact config the author runs day to day, with third-party skills.
+- **[$30 Preset](docs/thirty-dollars-preset.md)** — a mixed-provider setup built around Codex Plus and GitHub Copilot Pro for about $30/month.
+
 ### For Alternative Providers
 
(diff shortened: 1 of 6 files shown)


mhenke:
Fixed the Source column quoting inconsistency across all four preset docs. All Skill Reference tables now use backticks for `author` and `public` values to match the style in `authors-preset.md`.

Thanks for the catch @greptile-apps.

PR #722, opened 2026-07-10
fix: fail over stalled subagent retries

ShreeMulay:
## Summary
- classify structured 500/502/503/504 and bounded transport failures as failover-worthy
- budget every OpenCode `session.status: retry` event while preserving retry counts across `busy`
- register the resolved agent before the first provider response so child fallback chains remain isolated
- preserve `isRateLimitError` as a deprecated compatibility alias for `isFailoverError`

## Why
OpenCode 1.17.18 can retry provider 5xx/transport failures indefinitely. OMO 2.1.1 only recognized a subset of retry messages, so a delegated child could remain busy without advancing its configured fallback chain.

Numeric 5xx matching remains structured-field-only; arbitrary free-text `503` does not trigger fallback. Explicit retry events are debounced by the existing `maxRetries` budget, and interleaved `busy` events no longer reset that budget.

## Validation
- `bun test src/hooks/foreground-fallback/index.test.ts` — 59 pass
- `bun test` — 1,439 pass
- `bun run typecheck`
- `bun run lint`
- `bun run build`

`fallback.timeoutMs` and `retryDelayMs` wiring is intentionally out of scope.

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -1,6 +1,10 @@
 import { beforeEach, describe, expect, mock, test } from 'bun:test';
 import { SessionLifecycle } from '../session-lifecycle';
-import { ForegroundFallbackManager, isRateLimitError } from './index';
+import {
+  ForegroundFallbackManager,
+  isFailoverError,
+  isRateLimitError,
+} from './index';
 
 type ForegroundFallbackClient = ConstructorParameters<
   typeof ForegroundFallbackManager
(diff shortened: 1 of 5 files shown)


ShreeMulay:
Follow-up runtime evidence after commit `a76322bcc9ad517b2d87306a03948053bcac8c98`:

- OpenCode 1.17.18 delegated Oracle, generic HTTP 503: switched to configured fallback and returned the expected token within 45s
- refused transport: switched and returned token within 11s
- HTTP 429 control: switched and returned token within 49s
- HTTP 400 negative control: no fallback, bounded completion within 10s
- full source suite: 1,442 pass; focused fallback suite: 62 pass; typecheck/lint/build green

The second commit aborts the active OpenCode retry loop before re-prompting, includes the known agent in the fallback prompt, and makes the first registered session-agent identity sticky. Trade-off: a foreground session that intentionally switches agents mid-session retains the original fallback chain. This is bounded by runtimeOverride/out-of-chain handling and prevents delegated child identity from being overwritten during fallback.

mhenke:
@greptile-apps please update your summary

alvinunreal:
Directly merged/repaired via omos/pr-722-retry-fallback

PR #740, opened 2026-07-12
feat: add native cmux multiplexer support

HighColdHC:
## Summary
- add a native cmux multiplexer adapter with right-side, vertically stacked, equal-height agent panes
- add cmux-specific session lifecycle handling for readiness, stable idle detection, deferred spawning, bounded orphan recovery, and instance disposal
- resolve and quote the host OpenCode executable, wire configuration/auto-detection, and document cmux setup

## Validation
- `bun test` — 1528 passed, 0 skipped
- `bun test src/multiplexer` — 137 passed, 0 skipped
- `bun run typecheck`
- `bun run build`
- `git diff --check`

## Notes
- Requires cmux 0.64.14 or newer.
- Orphan recovery is process-global across instance reloads; hard process crashes are not persisted.


alvinunreal:
thanks @HighColdHC

PR #767, opened 2026-07-13
fix(tui-state): scope state file per project to stop model override leaks

### LINKED ISSUE #741: tui-state.json has no project isolation — stale model overrides leak across projects causing 'Model not found'
## Summary

`tui-state.json` stores runtime agent model overrides at a **fixed global path** with no project isolation. Runtime model selections from one project leak into every other project on the same machine, causing `Model not found` errors when a project's config doesn't define the same providers/models.

## Environment

- **Plugin version:** `oh-my-opencode-slim@2.1.1`
- **OpenCode:** `1.1.44`
- **OS:** Windows 11 (PowerShell 7)
- **Config preset:** `hybrid`

## Root cause

`src/tui-state.ts` computes the state file path with no project discriminator:

```ts
// dist/index.js:34763-34769  (and dist/tui.js:783-789)
var STATE_DIR = "oh-my-opencode-slim";
var STATE_FILE = "tui-state.json";
function dataDir() {
  return process.env.XDG_DATA_HOME ?? path.join(os.homedir(), ".local", "share");
}
function getTuiStatePath() {
  return path.join(dataDir(), "opencode", "storage", STATE_DIR, STATE_FILE);
}
```

The path resolves to `~/.local/share/opencode/storage/oh-my-opencode-slim/tui-state.json` for **every** project. `recordTuiAgentModels()` (line 34818) and `recordTuiAgentModel()` (line 34824) write agent model selections into this single file. On the next project's startup, `readTuiSnapshot()` (line 34790) returns the previous project's models before the new config is applied.

## Reproduction

1. **Project A** — config uses `openai/gpt-5.4-mini` for `explorer`, `librarian`, `designer`, `fixer` (e.g. via `/model` runtime switch or an `openai` preset). Start OpenCode there. `tui-state.json` now contains:
   ```json
   {"agentModels":{"explorer":"openai/gpt-5.4-mini","librarian":"openai/gpt-5.4-mini","designer":"openai/gpt-5.4-mini","fixer":"openai/gpt-5.4-mini", ...}}
   ```
2. **Project B** (`markt.de`) — config uses preset `hybrid` with `omniroute/cx/gpt-5.4-mini` and `ollama-cloud/...` models. The global `opencode.jsonc` defines providers `omniroute`, `freebuff`, `google`, `llmproxy` — **no `openai` provider**.
3. Start OpenCode in Project B. The stale `tui-state.json` from Project A is read before the new config is applied. Agents resolve to `openai/gpt-5.4-mini`, which doesn't exist in Project B's providers.

## Error

```
Sitzungsfehler
Fehler: Model not found: openai/gpt-5.4-mini. Did you mean: gpt-5.4-mini, gpt-5.4-mini-fast?
```

The suggestion list comes from a different provider (the OmniRoute proxy at `127.0.0.1:18080`) that does offer `gpt-5.4-mini` — confirming the model string is being looked up under the wrong provider.

## Observed state file

```json
{
  "version": 1,
  "updatedAt": 1783862931336,
  "agentModels": {
    "orchestrator": "ollama-cloud/glm-5.2",
    "explorer": "openai/gpt-5.4-mini",
    "librarian": "openai/gpt-5.4-mini",
    "oracle": "openai/gpt-5.5",
    "designer": "openai/gpt-5.4-mini",
    "fixer": "openai/gpt-5.4-mini"
  }
}
```

`orchestrator` was overwritten by Project B's startup, but the subagents kept Project A's `openai/...` models — indicating the overwrite is partial or loses the race against agent instantiation.

## Impact

- Any multi-project setup where projects use different providers/models is affected.
- The error is confusing because the suggestion list comes from whichever provider happens to offer a similarly-named model.
- Workaround is to delete `tui-state.json` before switching projects, which loses legitimate runtime overrides.

## Suggested fix

Scope `tui-state.json` per project. Options:

1. **Project-scoped path** — include the project hash in the path:
   ```ts
   function getTuiStatePath() {
     const projectHash = /* from opencode project context */;
     return path.join(dataDir(), "opencode", "storage", STATE_DIR, projectHash, STATE_FILE);
   }
   ```

2. **Project field in the snapshot** — keep a single file but key `agentModels` by project:
   ```ts
   { "version": 2, "projects": { "<hash>": { "agentModels": {...} } } }
   ```

3. **Validate against active providers on load** — if a stored model's provider isn't registered, drop it and fall back to the preset default instead of surfacing a hard error.

Option 1 is the cleanest and matches OpenCode's own `storage/session/<projectHash>/` layout.

mhenke:

Issue #741: `tui-state.json` stores runtime agent model overrides (set via `/model`) at one fixed global path with no project isolation. A user who switches models in project A and then opens project B (different providers/models) gets project A's stale overrides read before B's config applies, which throws `Model not found: openai/gpt-5.4-mini` in B. The only workaround was deleting `tui-state.json` when switching projects, which also dropped legitimate overrides.

## What does this change?
Scopes `tui-state.json` per project by hashing the project directory into the storage path (`.../oh-my-opencode-slim/<sha256(projectDir)[:12]>/tui-state.json`) instead of a single global file. Every read/write caller now receives the project directory.

## Why this approach?
Three options were considered. Per-project path (chosen) matches OpenCode's own `storage/session/<projectHash>/` layout and is the smallest correct fix. A single shared file keyed by project would keep a global file and complicate the schema. Validating stored models against active providers on load only hides the symptom. The per-project path avoids any cross-project read.

## Related work
- Related: #741 (the issue this fixes). Adjacent but distinct: #639 (preserve /model across plugin re-inits), #737 (disable fallback when user picks a model via /model). No existing PR fixes the cross-project isolation.

## Acceptance testing
1. Confirm two projects get separate files. Start OpenCode in project A, then in project B. Check the storage dir:
   ```
   ls ~/.local/share/opencode/storage/oh-my-opencode-slim/
   ```
   You should see two different 12-char hash directories, one per project. Before the fix there was a single `tui-state.json` at that level with no hash folder.
2. To confirm a directory matches a project, compute the expected hash (matches `path.resolve` + sha256, first 12 hex):
   ```
   printf '%s' "$(realpath -s /abs/path/to/projectA)" | sha256sum | cut -c1-12
   ```
   That string should equal one of the directory names. `realpath -s` avoids symlink resolution, same as `path.resolve`.

## AI assistance
- Model / harness: OpenCode (hy3-free) with @oracle (hy3-free) subagents for the code review and @fixer (hy3-free) for the test and import change.
- Human reviewed the full diff? Yes. The change was manually acceptance-tested by opening two projects and confirming separate 12-char hash state directories are created, with no model leak across projects.

## Checklist
- [x] `bun run check:ci`, `bun run typecheck`, and `bun test` pass (1550 tests, 0 fail)
- [x] Docs updated if behavior/commands changed: no doc change needed. This is an internal state-file location change with no user-facing command, config, or behavior surface.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -146,6 +146,9 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
   let multiplexerSessionManager: MultiplexerSessionManager;
   let autoUpdateChecker: ReturnType<typeof createAutoUpdateCheckerHook>;
   let sessionAgentMap: Map<string, string>;
+  // ponytail: cache sessionID -> project directory so TUI model writes
+  // land in the right per-project file after a project switch (ctx.directory is stale)
+  const sessionDirectories = new Map<string, string>();
   let sessionLifecycle: SessionLifecycle;
 
   let chatHeadersHook: ReturnType<typeof createChatHeadersHook>;
@@ -865,6 +868,7 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
               modelID?: string;
             };
             sessionID?: string;
+            directory?: string;
           };
           sessionID?: string;
           id?: string;
@@ -897,7 +901,8 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
               model,
               variant: variant ?? null,
             },
-            ctx.directory,
+            (info?.sessionID && sessionDirectories.get(info.sessionID)) ??
+              ctx.directory,
           );
         }
       }
@@ -907,6 +912,11 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
         const parentSessionId = event.properties?.info?.parentID;
         if (depthTracker && childSessionId && parentSessionId) {
           depthTracker.registerChild(parentSessionId, childSessionId);
+        }
+        const createdSessionId = event.properties?.info?.id;
+        const createdSessionDir = event.properties?.info?.directory;
+        if (createdSessionId && createdSessionDir) {
+          sessionDirectories.set(createdSessionId, createdSessionDir);
         }
       }
 
@@ -987,6 +997,7 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
         }
         if (sessionID) {
           sessionAgentMap.delete(sessionID);
+          sessionDirectories.delete(sessionID);
         }
       }
     },
--- a/src/index.ts
+++ b/src/index.ts
@@ -772,10 +772,13 @@ const OhMyOpenCodeLite: Plugin = async (ctx) => {
           tuiAgentVariants[agentDef.name] = resolvedVariant;
         }
       }
-      recordTuiAgentModels({
-        agentModels: tuiAgentModels,
-        agentVariants: tuiAgentVariants,
-      });
+      recordTuiAgentModels(
+        {
+          agentModels: tuiAgentModels,
+          agentVariants: tuiAgentVariants,
+        },
+        ctx.directory,
+      );
 
       applyOrchestratorModelConfig({
         agents: configAgent,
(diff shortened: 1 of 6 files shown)


mhenke:
@greptile-apps please re-run your review/check on the latest changes — the message.updated handler now reads the live project directory from the event (info.path.cwd) instead of the stale ctx.directory.

mhenke:
@greptile-apps please re-run your review/check. The previous info.path.cwd approach was dead code (UserMessage carries agent but no path); replaced it with a sessionID->directory cache populated from session.created events, so the message.updated handler now writes TUI model state to the correct per-project file via the live session directory (fallback ctx.directory).

PR #786, opened 2026-07-15
Remove redundant subagent depth limiting

adikpb:

OpenCode introduced native subagent depth limiting (https://github.com/anomalyco/opencode/pull/37124/commits, `subagent_depth` config, default 1). The plugin's own `SubagentDepthTracker` (default max depth 3) was redundant and is now removed.

**Why it was safe to remove:**
- The native limit is enforced inside `TaskTool.execute`, which the plugin's councillors don't use — they spawn via the OpenCode SDK `session.create` path.
- Councillors are already denied both `task` (TaskTool) and `council_session` (the council tool), so the council→councillor spawn path is already a recursion dead-end. The tracker was pure defense-in-depth.

**Changes:**
- Deleted `src/utils/subagent-depth.ts` and its test.
- Removed the depth pre-check, child registration, and cleanup from `src/council/council-manager.ts`.
- Removed the `depthTracker` wiring from `src/index.ts` and the `DEFAULT_MAX_SUBAGENT_DEPTH` constant from `src/config/constants.ts`.
- Added a one-line comment in `src/agents/index.ts` noting `council_session: 'deny'` on councillors is now the recursion guard.
- Updated docs (codemap.md, src/codemap.md, src/council/codemap.md, docs/council.md, docs/loop-engineering-research.md).

**Verification:** `bun run typecheck` clean, `bun run check:ci` clean, `bun test` = 1562 pass / 0 fail.


adikpb:
## CI note: the `test (latest)` failure is unrelated to this PR

The only failing check is one test:

```
(fail) system > isOpenCodeInstalled detects opencode in ~/.opencode/bin
```

This lives in `src/cli/system.test.ts`, which this PR does **not** touch. The change here only removes the redundant `SubagentDepthTracker` (council-manager.ts, index.ts, constants.ts, agents/index.ts, docs).

**Why it fails in CI but not locally:**
- It passes locally — `bun test src/cli/system.test.ts` → 6/6.
- The test writes a fake `opencode` binary into a temp dir and asserts detection via `process.env.HOME`. In the CI "latest" runner that detection returns `false` (the runner's temp filesystem is likely `noexec`, or `HOME`/`PATH` resolution differs). It's an environment-dependent test, not a logic regression.
- CI ran 1562 tests with exactly this 1 failure (1561 pass / 1 fail) — i.e. the same suite minus this environmental case.

**Conclusion:** pre-existing/environmental failure, safe to re-run or fix separately. No change needed in this PR. Greptile Review passed.

alvinunreal:
thanks @adikpb

PR #804, opened 2026-07-17
Make background job board injection cache-safe

DanielMaly:
The Background Job Board was prepended transiently to the latest orchestrator user message on every request, so each turn mutated message history and broke the provider prompt cache (exact byte-prefix matching), zeroing cached_tokens on every turn while background tasks were active.

This change makes the injection cache-safe:

- Tagged synthetic board parts are stripped from all messages first (whole synthetic board messages are pruned), so history stays byte-stable.
- The current board is appended as its own trailing user message rather than mutating an existing one. In long tool loops the latest user message becomes deep history; rewriting it on board state changes would invalidate the cache for everything after it. At the tail, a board state change only costs the board itself (~170 tokens).
- The board step runs last in the chat.params pipeline so other injectors cannot displace it.

Measured with a recording proxy against Copilot gpt-5.6-sol (opencode 1.18.3), orchestrator sessions with background subagents launched mid-session: every turn hits 97.0-99.1% prompt cache (steady state ~98%), versus 0% before this change under the same scenario. Byte-prefix analysis shows perfect prefix extension on quiet turns and exactly one board-message-sized divergence (699 bytes) on turns where the board changed. Complements #790.

### DIFF
--- a/src/hooks/task-session-manager/index.test.ts
+++ b/src/hooks/task-session-manager/index.test.ts
@@ -75,6 +75,14 @@ function createMessages(sessionID: string, text = 'user message') {
   };
 }
 
+async function transformMessages(
+  hook: ReturnType<typeof createTaskSessionManagerHook>,
+  messages: { messages: unknown[] },
+) {
+  await hook['experimental.chat.messages.transform']({}, messages as never);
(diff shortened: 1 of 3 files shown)


alvinunreal:
nice @DanielMaly

PR #814, opened 2026-07-17
fix(fixer): add design-boundary constraint to fixer prompt

### LINKED ISSUE #444: [Bug]: UI/frontend tasks often bypass designer and get handled by orchestrator/fixer

I noticed this in both `omo` and `omo-slim`.

When a task involves frontend logic or a small UI change, it often gets handled directly by the `orchestrator` or delegated to `fixer` instead of `designer`.
In my setup, both `orchestrator` and `fixer` are `gpt-5.5`, so the UI often gets replaced with the typical GPT-slop design pattern: uppercase eyebrow/subheading above the main heading, generic SaaS cards, gradients, spacing, etc. I think most people recognize this style immediately.

The problem is that once the UI has been rewritten into that style, asking `designer` to fix it later is much harder. The designer is now working from already-wrong context and usually only adjusts it slightly instead of restoring the original design direction.

**Expected behavior:**
If a task affects user-facing UI or frontend visual behavior, it should be routed to `designer` first, even if the change is small.
This includes styling, layout, component markup, Tailwind/CSS classes, responsive behavior, visible states, or frontend logic that changes what users see.
`fixer` should probably stay focused on headless implementation, tests, backend logic, or non-visual frontend plumbing.

### Steps to reproduce

_No response_

### oh-my-opencode.json

```jsonc

```

### OpenCode version

1.14.30

1.0.6

### Operating system

_No response_

```shell

```

mhenke:
Fixes #444

UI/frontend tasks get routed to fixer (or handled by orchestrator) instead of designer. Fixer has no design constraint in its own prompt — it knows not to research or delegate, but it was never told not to do design. When the orchestrator model misses the routing rule, or a user calls @fixer directly, fixer implements UI with whatever taste its model has (usually generic slop). Designer can’t easily restore the original direction from already-wrong context.

## What does this change?

One constraint line in the fixer system prompt: layout, styling, visual hierarchy, responsive behavior, animation, and component feel belong to @designer.

## Why this approach?

The orchestrator prompt already has routing rules telling it to send UI work to @designer and not @fixer. This adds defense in depth: the fixer itself now knows its boundary. A code-level hook for file-type detection would be over-engineered for a one-line constraint. No alternatives considered.

## Related work

- [x] Checked open and closed PRs/issues for duplicates
- Related: none found

## AI assistance

- Model / harness: OpenCode with @oracle (claude-sonnet-4-20250514) for code review
- [x] Human reviewed the full diff

## Checklist

- [x] `bun run check:ci`, `bun run typecheck`, and `bun test` pass
- [x] Docs: no behavior/command changes, README + docs/ unaffected
- [x] One logical change
- [x] PR targets `master`

### DIFF
--- a/src/agents/fixer.ts
+++ b/src/agents/fixer.ts
@@ -18,11 +18,12 @@ ${WRITABLE_FILE_OPERATIONS_RULES}
 
 **Constraints**:
 - NO external research (no websearch, context7, gh_grep)
-- NO delegation or spawning subagents
+- NO spawning subagents; telling the caller which specialist to use is fine
 - No multi-step research/planning; minimal execution sequence ok
 - If context is insufficient: use grep/glob/read directly - do not delegate
 - Only ask for missing inputs you truly cannot retrieve yourself
 - Do not act as the primary reviewer; implement requested changes and surface obvious issues briefly
+- No design work — layout, styling, visual hierarchy, responsive behavior, animation, component feel. Refuse and tell the caller to use @designer.
 
 **Output Format**:
 <summary>
--- a/src/agents/orchestrator.ts
+++ b/src/agents/orchestrator.ts
@@ -162,7 +162,8 @@ Review available agents and lane rules. Before beginning non-trivial work, ident
 
 **Routing threshold:**
 - Handle directly only for one isolated, clear, low-risk action where delegation would cost more than execution.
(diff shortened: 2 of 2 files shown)


mhenke:
@greptile-apps

PR #820, opened 2026-07-18
fix(windows): detect opencode installed via .cmd shims and add Windows install paths

GiuseppeBellamacina:
## Problem

On Windows, `bunx oh-my-opencode-slim install` fails immediately after the companion prompt with:

```
[x] OpenCode is not installed on this system.
```

even when opencode is installed and running. This is reported by users running opencode right now — the check is wrong, not the install.

## Root cause

The check in `src/cli/system.ts` (`isOpenCodeInstalled` -> `resolvePathCommand` + `canExecute`) breaks on Windows in three compounding ways:

1. **`where opencode` returns multiple shims** (`opencode`, `opencode.cmd`, `opencode.ps1`). The code picked the first line via `.find(Boolean)`, which is the **extensionless npm wrapper** that Node cannot spawn on Windows.
2. **Node's CVE-2024-27980 patch blocks `.cmd`/`.bat` via `spawnSync` without `shell: true`**, throwing `Error: spawn EINVAL`. The `try/catch` swallows it and returns `false`.
3. **The fallback `getOpenCodePaths` list had zero Windows entries** (only Unix paths: `~/.local/bin`, homebrew, snap, nix, cargo, npm-global…), so the check never recovered.

The flow in `src/cli/install.ts` is:

```ts
const companionInstall = await shouldInstallCompanion(config); // asks companion? y/n
...
const { ok } = await checkOpenCodeInstalled();                 // broken check runs next
if (!ok) return 1;                                             // exits 1
```

So answering `y` to the companion prompt is immediately followed by the broken detection, which exits. The companion answer doesn't gate the check — it just happens to be the preceding step.

## Fix

- `resolvePathCommand`: pass `shell: true` to `spawnSync` on win32, and prefer executable shims (`.cmd`/`.exe`/`.ps1`/`.bat`) over the extensionless npm wrapper when parsing `where` output.
- `canExecute`: pass `shell: true` to `spawnSync` on win32 so `.cmd` shims can execute.
- `getOpenCodePaths`: add Windows install paths (npm `%APPDATA%\npm`, pnpm, yarn, opencode.ai installer at `%LOCALAPPDATA%\Programs\opencode`, Scoop, Chocolatey).
- `checkOpenCodeInstalled` in `install.ts`: print Windows-specific install hints (PowerShell `irm | iex`, `winget install opencode`) instead of the Unix-only `curl | bash` / `export PATH=...`.

## Verification

- `bunx tsc --noEmit` -> exit 0, no errors.
- `bunx @biomejs/biome lint src/cli/system.ts src/cli/install.ts` -> clean.
- Full `bun run build` fails on Windows for a **pre-existing, unrelated** reason (the `--external @opencode-ai/plugin/*` glob in `build:plugin` doesn't expand on Windows shells), so it's not in scope here.

## Impact

Unblocks Windows installs. Without this fix, no Windows user can pass the `checkOpenCodeInstalled` step when opencode is installed via npm/pnpm/yarn/scoop/chocolatey — the installer always exits 1 right after the companion prompt.

### DIFF
--- a/src/cli/install.ts
+++ b/src/cli/install.ts
@@ -126,15 +126,30 @@ async function checkOpenCodeInstalled(): Promise<{
 }> {
   const installed = await isOpenCodeInstalled();
   if (!installed) {
+    const isWindows = process.platform === 'win32';
     printError('OpenCode is not installed on this system.');
     printInfo('Install it with:');
-    console.log(
-      `     ${BLUE}curl -fsSL https://opencode.ai/install | bash${RESET}`,
-    );
-    console.log();
-    printInfo('Or if already installed, add it to your PATH:');
-    console.log(`     ${BLUE}export PATH="$HOME/.local/bin:$PATH"${RESET}`);
-    console.log(`     ${BLUE}export PATH="$HOME/.opencode/bin:$PATH"${RESET}`);
+    if (isWindows) {
+      console.log(
+        `     ${BLUE}powershell -NoProfile -ExecutionPolicy Bypass -Command "irm https://opencode.ai/install.ps1 | iex"${RESET}`,
+      );
+      console.log();
+      printInfo('Or with winget:');
+      console.log(`     ${BLUE}winget install opencode${RESET}`);
+      console.log();
+      printInfo('Or if already installed, add it to your PATH:');
+      console.log(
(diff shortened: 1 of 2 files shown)


alvinunreal:
thanks @GiuseppeBellamacina

PR #829, opened 2026-07-18
fix(foreground-fallback): prevent background tasks reported as cancelled during fallback (issue #765)

### LINKED ISSUE #765: Background task reported as 'cancelled' when it actually completed during foreground fallback - rate limited
I sent a simple `@oracle hello` and the oracle responded (Qwen3.6 Plus replied with a greeting), but the task tool returned "Task cancelled" to the orchestrator.

Looking at the logs, the issue seems to be a race between foreground fallback and idle reconciliation:

1. Oracle session starts with `opencode/big-pickle`
2. Model fails, fallback kicks in: `big-pickle` → `qwen3.6-plus` → `qwen3.6-plus`
3. Session goes idle at some point
4. The background job board shows `runningJobForSession: false` when idle fires, meaning the job was never tracked as "running" on the board
5. Since the board doesn't think the job is running, the idle reconciliation path at line 693 in `task-session-manager/index.ts` never fires
6. The task gets reported as cancelled even though the oracle actually responded

Relevant log lines:

```
[22:56:44.618Z] [task-session-manager] idle/status idle observed {"sessionID":"ses_0a7746404ffeB6J8JiKCvFYh1e","managesSession":false,"terminalJobsPending":0,"runningJobForSession":false}
```

The `runningJobForSession: false` is the problem. The job was registered in `tool.execute.after` but the board's state doesn't reflect it as running when idle fires.

This only happens during foreground fallback. Without fallback, the normal flow works fine.

Versions: oh-my-opencode-slim current main, OpenCode latest.

mhenke:
Closes #765

## Problem
Background tasks were reported as 'cancelled' during foreground fallback when the model failed and the fallback chain kicked in. The race between the abort/re-prompt cycle and idle reconciliation caused the orchestrator to lose track of the task.

## Root Cause
The background job board never registered the task as 'running' when the session went idle during the fallback abort/re-prompt cycle. Since the board didn't think the job was running, the idle reconciliation path never fired, and the task was reported as cancelled even though the oracle actually completed.

## Fix
1. **Early board registration in session.created** — registers the job before tool.execute.after, so the board knows the job is running when idle fires
2. **Retry-without-abort before abort** — aborting triggers 'Task cancelled' in the parent; retrying promptAsync without abort first avoids this
3. **isFallbackInProgress drop guard** — prevents onSessionDeleted from dropping the job during the fallback cycle

## Simplifications (from code review)
- Removed wasFallbackRecent grace window (~190 lines) — speculative defense on top of isFallbackInProgress
- Removed deprecated isRateLimitError wrapper

## Verification
- All 100+ tests pass
- Net reduction: -221 lines across 4 files


mhenke:
Closing to recreate with proper template and gates.

Issue #595, opened 2026-06-24
[Bug]: Fallback is detached from background task lifecycle, causing lost failures and orphaned successes


kumar-shivang:
### What happened, and what did you expect?

I configured a subagent with multiple models as a fallback chain and launched it using task(..., background: true).

When the primary model failed with a fallback-eligible error such as Insufficient balance, fallback correctly switched to another model. However, the original background task did not receive the final outcome.

I observed both cases:

1. Fallback succeeds: the fallback model completes the work in the child session, but the parent/orchestrator is never notified. The task remains shown as running indefinitely.
2. All models fail: the parent/orchestrator never receives a final task failure indicating that the fallback chain was exhausted. The task may remain running or disappear through cancellation.

I expected the background task to remain logically running across all fallback attempts and produce exactly one final outcome:

primary succeeds → completed
primary fails, fallback succeeds → completed
all configured models fail → error

The fallback result should be delivered to the parent through the same task-completion mechanism as the original model attempt.

It appears that fallback currently aborts the child session and re-prompts it asynchronously. This detaches the fallback attempt from the original OpenCode background job that owns the parent completion notification.

### Steps to reproduce

1. Clear the OpenCode cache and restart:
rm -rf ~/.cache/opencode
2. Configure a subagent with at least two models:
{
  "agents": {
    "explorer": {
      "model": [
        "provider-a/primary-model",
        "provider-b/fallback-model"
      ]
    }
  },
  "fallback": {
    "enabled": true
  }
}
Make the primary model return a fallback-eligible error, for example by using a provider account with insufficient balance.
Launch the subagent in the background:
Use the explorer subagent in background mode to investigate a simple task.

Equivalent task arguments:

{
  "subagent_type": "explorer",
  "description": "Test background fallback",
  "prompt": "Investigate this simple task and return a summary.",
  "background": true
}
Test both outcomes:
Case A: fallback succeeds

Allow the second model to complete successfully.

Observed:

the fallback output appears in the child session;
the parent does not receive a completed task result;
the background task remains running indefinitely.
Case B: all models fail

Make both configured models return an insufficient-balance, quota, or rate-limit error.

Observed:

the parent does not receive a final state="error" result;
the task is not reliably marked failed after the fallback chain is exhausted.

Expected in Case A:

<task id="child-session-id" state="completed">
  <summary>Background task completed: Test background fallback</summary>
  <task_result>
    Result produced by the fallback model
  </task_result>
</task>

[...]

PR #837, opened 2026-07-19
fix(task-session): parallel early-reg mis-attribution + idle/FG false-complete race

Jiajun0413:
## Problem

When launching **parallel** task tools (e.g. council `councillor-reviewer-a/b/c`), two bugs caused false cancel/complete while the child kept running:

### 1. Early board registration attributes all children to the first agent

`session.created` early registration called `peekByParent(parentID)`, which returns the **first** pending call in insertion order. With three parallel task tools in flight:

```
before(a) → before(b) → before(c)
session.created(child-b) → peek → a's pending  // wrong
session.created(child-c) → peek → a's pending  // wrong
```

Log evidence (`oh-my-opencode-slim.20260719T081220.log`):
- cou-2 / cou-3 both registered as `agent: councillor-reviewer-a`

### 2. Immediate idle reconcile races FG fallback (false complete)

Child idle was reconciled to `completed` **immediately**. OpenCode can emit idle for a rate-limited child **before** `ForegroundFallbackManager` sets `isFallbackInProgress`:

```
08:34:09.362 idle → reconciled running job from idle (cou-2)
08:34:09.555 FG switched zfy → shuo
08:34:09.632 busy with previousState: reconciled, terminalState: completed
… session keeps working for minutes
```

Board says done; session still runs. Parent sees cancelled/completed while child continues — the "错误 cancel" UX.

### 3. `session.deleted` doesn't cancel pending child idle-reconcile timer

Greptile P1: during FG teardown, idle → schedule timer → FG aborts → `onSessionDeleted` preserves board entry (correct) but leaves the timer running → FG finishes (`isFallbackInProgress=false`) → timer fires and falsely reconciles while the re-prompted session keeps working.

## Fix

1. **`peekByParentAndAgent(parent, info.agent)`** — OpenCode sets `SessionInfo.agent` to the subagent name at create (`sst/opencode` `task.ts` → `sessions.create({ agent: next.name })`). Prefer the pending call whose `agentType` matches; fall back to oldest only if no match.

2. **Delay child idle-reconcile** (same `idleReconcileDelayMs` window as parent continuation):
   - Re-check `isFallbackInProgress` after delay
   - Skip if `lastLiveBusyAt` is after the idle was observed
   - Cancel pending timer on live `busy` status

3. **Cancel `childIdleReconcileTimers` on `session.deleted`** — matches `server.instance.disposed` path. A deleted session has no future idle/busy; any pending timer can only mis-reconcile.

## Tests

- Parallel early-reg: a/b/c children get correct agent/description
- FG race: idle then busy before delay → stays `running`
- FG teardown race: idle → session.deleted before delay → stays `running`
- Existing idle/fallback tests updated for delayed reconcile

1621 tests pass across the full suite. CI green.

## Relation to #822 and issues

- **#822** stopped FG from aborting agents with **no** chain (plain `councillor`). Parallel **reviewer-*** agents **have** chains and still use FG — this PR fixes the board/lifecycle side of that path.
- **#595** "Fallback is detached from background task lifecycle, causing lost failures and orphaned successes" — the root issue. The idle/FG race fixed here is the mechanism behind the false-complete/orphaned-success symptom.
- **#765** "Background task reported as cancelled when it actually completed during foreground fallback" — fixed the early-registration side; this PR extends that protection to the parallel + idle-reconcile-race scenarios.

### DIFF
--- a/src/cache-safety-tripwire.test.ts
+++ b/src/cache-safety-tripwire.test.ts
@@ -63,6 +63,10 @@ const ALLOWLIST = new Map<string, string>([
     'Date.now() records lastReadAt for internal recency ordering; formatted prompt output (background job board) is confined to the volatile trailing message.',
   ],
   [
+    'hooks/task-session-manager/index.ts',
+    'Date.now() captures idleObservedAt to detect post-idle busy recovery from foreground-fallback re-prompts; never serialized into prompt content.',
+  ],
+  [
     'hooks/image-hook.ts',
     'Date.now() throttles temp-image cleanup; extracted image paths are deterministic per part id.',
   ],
--- a/src/hooks/task-session-manager/index.test.ts
+++ b/src/hooks/task-session-manager/index.test.ts
@@ -25,6 +25,11 @@ async function flushContinuation(): Promise<void> {
   await new Promise((resolve) => setTimeout(resolve, 0));
 }
 
+/** Flush delayed child idle-reconcile timers when idleReconcileDelayMs is 0. */
+async function flushChildIdleReconcile(): Promise<void> {
+  await new Promise((resolve) => setTimeout(resolve, 5));
+}
+
 function createHook(options?: {
   shouldManageSession?: (sessionID: string) => boolean;
   registerSessionAsOrchestrator?: (sessionID: string) => void;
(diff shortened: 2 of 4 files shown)


alvinunreal:
@greptileai

PR #847, opened 2026-07-20
fix(preset): set OpenAI xhigh variants

alvinunreal:
## Summary
- Only the OpenAI preset orchestrator, oracle, and fixer variants change to \`xhigh\`.
- All model IDs and other agent variants remain unchanged.

## Validation
- \`bun run check:ci\`, \`bun run typecheck\`, and \`bun run build\` pass.
- Full \`bun test\` has the unrelated existing deepwork prompt assertion failure expecting \`simplify/readability\`.

### DIFF
--- a/README.ja-JP.md
+++ b/README.ja-JP.md
@@ -153,12 +153,12 @@ bun run build
   "preset": "openai",
   "presets": {
     "openai": {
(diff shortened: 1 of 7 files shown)


jeshtan:
@alvinunreal may I know the reason behind changing the OpenAI preset to xhigh for terra and luna model? In this case, does sol (medium) would performan better than terra (xhigh)?

HorikitaSuzuneTsundere:
@alvinunreal I'm also curious about the change. I was expecting changes for the OpenCode Go preset as well.

PR #848, opened 2026-07-20
refactor(task-session-manager): decompose monolithic index into focused subsystems

adikpb:

The `task-session-manager` entry point (`src/hooks/task-session-manager/index.ts`) had grown to ~1300 lines, mixing board injection, continuation evaluation, idle reconciliation, event routing, input-wait tracking, and tool-execute hooks in one file. This made the module hard to navigate, review, and reason about, and increased the risk of cache-safety regressions when editing prompt-assembly paths.

This PR is a behavior-preserving refactor that extracts single-responsibility subsystems:

- `board-injection.ts` — Background Job Board injection + synthetic completion processing (still routed through `../cache-safe-injection.ts`).
- `continuation-evaluator.ts` — pure continuation-nudge evaluation, taking explicit dependency objects to avoid circular imports.
- `continuation-token-manager.ts` — per-session continuation token + in-flight evaluation tracking.
- `event-router.ts` — routes lifecycle events (created/disposed/idle/error/status/deleted) to the right subsystems.
- `idle-reconciliation.ts` — delayed idle reconciliation with fallback/input-wait guards.
- `status-utils.ts` — shared status predicates and task-summary extraction.
- `tool-execute-hooks.ts` — `tool.execute.before`/`.after` handling (pending calls, launch/read tracking).
- `input-wait-tracker.ts` — permission/question ask/resolution wait tracking.

`index.ts` now composes these subsystems and delegates; no behavioral logic remains inline. Tests (`src/hooks/task-session-manager/index.test.ts`) and the cache-safety tripwire were updated to track the new module boundaries. `bun run typecheck` is clean and the full task-session-manager suite (104 tests) passes.

No user-facing behavior change — this is internal structure only.

### DIFF
--- a/src/agents/orchestrator.test.ts
+++ b/src/agents/orchestrator.test.ts
@@ -11,5 +11,4 @@ describe('orchestrator prompt', () => {
     expect(prompt).toContain('small bounded set of options');
     expect(prompt).toContain('ordinary dialogue that does not block work');
   });
-
 });
--- a/src/cache-safety-tripwire.test.ts
+++ b/src/cache-safety-tripwire.test.ts
@@ -63,7 +63,7 @@ const ALLOWLIST = new Map<string, string>([
     'Date.now() records lastReadAt for internal recency ordering; formatted prompt output (background job board) is confined to the volatile trailing message.',
   ],
   [
-    'hooks/task-session-manager/index.ts',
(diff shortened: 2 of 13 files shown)


alvinunreal:
good one @adikpb

PR #849, opened 2026-07-20
fix(prompt): remove misleading 'do not poll' instructions

adikpb:
1. What changed, and why was it needed?

The Background Job Board prompt (`src/utils/background-job-board.ts`) and the scheduler `PHASE_REMINDER_TEXT` (`src/config/constants.ts`) both instructed the orchestrator to "not poll running jobs." This instruction was inaccurate: the agent has no poll/status tool — background-job completion is hook-driven. The misleading clause was removed from both prompt surfaces, and the golden snapshot (`src/hooks/__snapshots__/cache-payload.snapshot.test.ts.snap`) was updated deliberately to reflect the trimmed text.

Two commits:
- `37a3d78` — remove the line from the job board prompt + snapshot
- `7b3684a` — remove the same clause from `PHASE_REMINDER_TEXT` + snapshot (5 snapshot sites)

Verification: `bun test src/hooks/cache-payload.snapshot.test.ts` passes (3 pass, 0 fail); snapshot is in sync with the trimmed source. A repo-wide search confirms no remaining "do not poll" / "consume running-job" / "advance dependent" instructions in any prompt surface.

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -48,7 +48,7 @@ export const POLL_INTERVAL_BACKGROUND_MS = 2000;
 export const MAX_POLL_TIME_MS = 5 * 60 * 1000; // 5 minutes
 
 // Workflow reminders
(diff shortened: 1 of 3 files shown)


alvinunreal:
nice @adikpb

Issue #815, opened 2026-07-17
[Bug]: Continuation nudge triggers a second orchestrator response while waiting for user input


slowdream:
### What happened, and what did you expect?

When the orchestrator encounters a blocker and asks the user for clarification, permission, or command output, the session becomes idle while some todos remain incomplete.

After approximately two seconds, oh-my-opencode-slim sends an internal synthetic prompt:

Continue coordinating the remaining incomplete todos. Do not finalize while work remains.

This triggers another orchestrator turn immediately after the question. The orchestrator then responds a second time even though it cannot continue until the user answers.

Example:

Assistant:
Please run `docker compose logs app` and send me the output.

Slim internal continuation:
Continue coordinating the remaining incomplete todos.

Assistant:
I still need the command output before I can continue.

This also happens when the orchestrator asks a normal clarification question while keeping the current task todos open.

Expected behavior

The continuation nudge should not run when the orchestrator is intentionally waiting for user input.

Incomplete todos do not necessarily mean the orchestrator stopped prematurely. They may remain incomplete because the next step depends on:

clarification from the user;
command output from the user's machine;
user approval or permission;
credentials or other external information.

The plugin should distinguish between:

idle with unfinished autonomous work

and:

idle because the orchestrator is waiting for the user

### Steps to reproduce

Start a task using the orchestrator agent.
Let the orchestrator create one or more todos.
Cause the task to require user input. For example, ask it to diagnose something that requires running a command on another machine.
The orchestrator asks the user to run a command or provide clarification while at least one todo remains pending or in_progress.
Do not answer the question.
Wait approximately two seconds.

Observe that the plugin injects a synthetic message containing:

Observe that the orchestrator starts another turn and produces a second response despite still waiting for the user.

### oh-my-opencode.json

```jsonc

```

### OpenCode version

1.18.3

### oh-my-opencode-slim version

2.2.3

### Operating system

ubuntu

### Logs, screenshots, or extra context

```shell

```

Issue #854, opened 2026-07-21
TaskSessionManager injects duplicate forced continuation nudges after idle with incomplete todos


ricatix:
## Environment

- oh-my-opencode-slim: `2.2.4`
- OpenCode: `1.17.x`
- Orchestrator session with background subagents and incomplete todos

## Actual behavior

After the orchestrator returns a response and becomes idle, `TaskSessionManager.evaluateContinuation()` injects the internal prompt below through `sessionSdk.promptAsync()`:

```text
Continue coordinating the remaining incomplete todos. Do not finalize while work remains.
```

In one local session, the same continuation nudge was injected **four times at the same timestamp**. The session then completed ten separate loop cycles. This is forced execution after an apparently final response, without a user decision.

Relevant local log shape:

```text
[task-session-manager] synthetic part missing task status
{textPreview:"Continue coordinating the remaining incomplete todos..."}
```

The source path in `2.2.4` is `dist/index.js`:

- `CONTINUATION_NUDGE` near line 26561
- `evaluateContinuation()` near line 26694
- `sessionSdk.promptAsync()` near lines 26753-26756
- idle reconciliation delay: `IDLE_RECONCILE_DELAY_MS = 2000`

## Expected behavior

1. At most one internal continuation prompt may be dispatched per session/idle transition.
2. Continuation should be explicit opt-in, or wait for the user when the orchestrator has produced a final response.
3. Users need a supported config switch to disable automatic continuation while retaining task/background-job orchestration.

## Why this is harmful

Duplicate `promptAsync()` dispatches can create overlapping assistant work, token waste, repeated subagent delegation, and responses continuing after the UI appears complete. This also amplifies known OpenCode `promptAsync` lifecycle races.

## Suggested fix

Minimal:

- Add a documented `backgroundJobs.disableContinuation` (or equivalent) config option.
- Return before `evaluateContinuation()` injects its nudge when enabled.
- Reserve/dedupe before `promptAsync()` so concurrent idle events cannot pass the guard.

Robust:

Use one shared internal-prompt gate: per-session reservation, queue, semantic dedupe, post-dispatch hold, timeout, and current-session/tool-state checks. `oh-my-openagent` uses this approach in its `prompt-async-gate` to avoid competing internal prompt injections.

## Related

- #167: duplicate background-task notifications (related, different trigger)
- #244: orchestrator stopping with todos (opposite failure mode)
- #560: foreground fallback orphaned result (related async lifecycle race)
[...]

Issue #852, opened 2026-07-20
[Bug]: orchestrator may continue execution unexpectedly after subagent completion or user replies


shizhx:
## Description

The latest version of oh-my-opencode-slim has an issue where injected prompts and hidden actions can cause the orchestrator to proceed unexpectedly.

## Issues observed

1. When a subagent finishes, oh-my-opencode-slim injects an additional prompt into the orchestrator context.

   This can cause the orchestrator to continue execution immediately instead of waiting for all subagents to complete.

2. After I send a reply to the orchestrator, it sometimes asks me a question, but immediately triggers a hidden action that makes the orchestrator answer itself instead of waiting for my response.

## Expected behavior

- The orchestrator should wait until all subagents have completed before deciding the next step.
- When the orchestrator asks the user a question, it should wait for the user's response and not trigger an automatic continuation.

## Actual behavior

- A subagent completion event can cause premature orchestration.
- User interaction flow can be interrupted by an unexpected hidden action that triggers self-continuation.

## Steps to reproduce

1. Run a task that uses the [grilling-with-docs(https://github.com/mattpocock/skills/tree/main/skills) skill.
2. During the execution of the skill, ask the orchestrator to dispatch subagents for exploration.
3. Wait for one of the exploration subagents to complete before the others.
4. Observe that the orchestrator may continue execution immediately instead of waiting for all subagents to finish.
5. In some cases, after the orchestrator asks a question, replying to it can trigger a hidden action that causes the orchestrator to continue by itself instead of waiting for the user's response.
## Impact

These behaviors can cause incorrect task progression and make human-in-the-loop interactions unreliable.

## Version

This issue only occurs in the latest version. Previous versions did not exhibit this behavior.

### oh-my-opencode.json

```jsonc

```

### OpenCode version

1.18.3

### oh-my-opencode-slim version

v2.2.4

### Logs, screenshots, or extra context

<img width="2131" height="1122" alt="Image" src="https://github.com/user-attachments/assets/e42eed00-9ccc-48df-a058-ec8784b4e9bb" />

<img width="2122" height="903" alt="Image" src="https://github.com/user-attachments/assets/d3764460-b85a-46e8-acf4-4a1d9702a155" />

shizhx:
I tried v2.2.5, but the issue still exists. I eventually had to roll back to v2.2.2, which works correctly.

alvinunreal:
[...]

PR #861, opened 2026-07-21
fix: add explicit wait_for_user for text-only HITL

### LINKED ISSUE #860: [Bug]: Continuation nudge still fires during text-only waits for user input

PR #819 fixed #815 by suppressing automatic continuation while OpenCode reports a pending `question` or `permission` request. The same behavior still occurs in oh-my-opencode-slim 2.2.5 when the human-in-the-loop boundary is expressed only in the assistant's text.

For example, the orchestrator can give the user a sequence of manual deployment steps, ask the user to perform them and report back, and end its turn with an incomplete todo. No user reply has arrived, but about two seconds later the plugin injects:

```text
Continue coordinating the remaining incomplete todos. Do not finalize while work remains.
<!-- SLIM_INTERNAL_INITIATOR -->
```

That synthetic message starts another orchestrator turn before the user has performed the requested operation. This happens at the first HITL boundary; it does not require the user to say "stop" first.

This is distinct from #854's duplicate-dispatch problem: this reproduction observes exactly one `promptAsync` call, but that one call is semantically wrong for a text-only HITL boundary.

Expected: after the orchestrator gives manual steps and intentionally waits for the result, the session remains idle until a new external user message arrives.

Actual: an incomplete todo causes an automatic continuation turn even though the next step depends on the user.

### Steps to reproduce

1. Install oh-my-opencode-slim 2.2.5 and start an orchestrator session with the default continuation behavior.
2. Send this prompt:

   ```text
   Create a todo for verifying a manual deployment step.
   Give me the exact steps to run `printf 'ready\n'` in another terminal, then wait for me to paste the output before doing anything else.
   Keep the verification todo in progress. Give the instructions in ordinary assistant text rather than opening a question or permission request.
   ```

3. Do not send another message after the orchestrator gives the manual steps.
4. Wait slightly longer than two seconds.
5. Observe the internal continuation prompt and a new assistant turn before any user input.

The same control path can be reduced deterministically at the public plugin-hook boundary: register an orchestrator user message, leave one todo pending, emit `session.idle`, and wait 2.1 seconds without emitting `question.asked`, `permission.asked`, or another user message. With the 2.2.5 built artifact, the stubbed `session.promptAsync()` is called once.

### Preliminary cause analysis

The 2.2.5 input-wait tracker records only `question.asked` and `permission.asked`, then clears those records on the corresponding reply or rejection events. An ordinary assistant response containing manual steps emits none of those events, so `hasInputWait()` remains false. When idle reconciliation finds an incomplete todo, the continuation evaluator therefore reaches `session.promptAsync()`.

This appears to leave #815 only partially covered: tool-backed question and permission waits are recognized, while a text-only HITL wait is indistinguishable from an orchestrator that stopped with autonomous work remaining. The process-local deduplication merged in #856 limits duplicate submissions, but it does not change this semantic distinction. A clean current-master control at `9105d5268681b6eae8e3fb78732588e89e9ef039` still submits one continuation for this case.

### oh-my-opencode.json

```jsonc
{}
```

No user configuration is required to reproduce the issue.

### OpenCode version

1.18.4 for the observed live session. The deterministic reproduction invokes the packaged plugin at its SDK boundary and does not require a running OpenCode host.

2.2.5 (`5b4ce76eb53cbcf3654c2ca2e59be8050b97efdb`)

### Operating system

macOS 15.7.1 (arm64)

The deterministic public-hook reduction was run with temporary, isolated OpenCode config, data, cache, log, and project directories. Its 2.2.5 control result was:

```text
pending todo calls: 1
promptAsync calls:   1
```

Related: #815, #819, #852, #854, and #856. #852 includes a partially overlapping HITL symptom as well as a separate subagent-completion path; this report isolates only the text-only external-user-wait case.

nettee:

Fixes #860.

Follow-up to #819 / #815. Builds on the process-global continuation gate from #856 / #854.

Related to #852; this addresses only the text-only external-user-wait path, not premature wakeups from subagent-completion events.

## Root cause

The incomplete-todo continuation hook treats an idle orchestrator with pending work as abandoned and queues an internal `promptAsync` nudge.

Native question and permission flows already suppress that behavior because they emit machine-readable wait events. A normal assistant message asking the user to perform external work emits no such event, so the session becomes idle and continues before the user can respond.

Inferring a HITL boundary from assistant prose would be brittle. This change makes that state transition explicit.

## Solution

- Add an orchestrator-only `wait_for_user` tool for text-only HITL boundaries. The orchestrator gives concrete manual steps first, then calls the tool as its final action. Immediate answers, choices, clarifications, and pasted output continue to use the native `question` tool.
- Encode the explicit wait as `waiting-for-user` in the existing process-global continuation-attempt gate.
  - Arming the wait cancels scheduled continuation work and revokes in-flight reservations.
  - Pre-upgrade continuation reservation/commit logic sharing the gate also fails closed.
- Release the wait only when a distinct external user text, file, or image message arrives.
  - Duplicate delivery of the preceding user message does not release it.
  - Synthetic/internal messages and foreground-fallback replays do not release it.
- Preserve the wait across hook/plugin recreation and fallback teardown in the same process, while clearing it on genuine session deletion.
- Mark fallback replays with internal provenance and share fallback-in-progress state across manager instances so teardown races cannot mimic a new external message or genuine deletion.
- Register the tool, orchestrator permissions, prompt guidance, docs, and `disabled_tools` behavior. When the tool is disabled, the prompt uses `question` as the blocking boundary and plugin health thresholds account for the intentionally disabled tool.

The tool's `reason` is diagnostic only. The implementation does not parse assistant prose. The latch is intentionally process-local and does not survive a process restart.

## Impact

Users can receive manual instructions without an incomplete todo immediately restarting the orchestrator. Normal continuation becomes eligible again after their next distinct message.

This adds a tool and changes the static orchestrator prompt, so deployment has the expected one-time prompt-cache prefix change. It introduces no per-turn mutation of the static/system prompt; variable `reason` text appears only in the ordinary tool result.

## Packaged RED -> GREEN reproduction

The same isolated harness was run against the public root export of two packed artifacts. It redirects OpenCode config, data, cache, logs, and project paths to a fresh temporary directory, then:

1. submits an orchestrator user message;
2. invokes `wait_for_user` when the package exposes it;
3. emits `session.idle` with one pending todo and no active sessions; and
4. counts todo reads and automatic `promptAsync` calls.

Both package manifests resolve the root export to `./dist/index.js`. Each extracted `package/dist/index.js` matched its source checkout byte-for-byte.

| Result | 2.2.5 baseline | This fix |
| --- | ---: | ---: |
| `waitToolAvailable` | `false` | `true` |
| `todoCalls` | `1` | `0` |
| `promptAsyncCalls` | `1` | `0` |
| Harness exit code | `1` | `0` |

A clean pre-patch build of `master` at `9105d5268681b6eae8e3fb78732588e89e9ef039` produced the same RED result, so the GREEN result is isolated from the commits between release 2.2.5 and current master.

<details>
<summary>Artifact SHA-256 values</summary>

- 2.2.5 baseline tarball: `9f6f14cd0d18a07585f455323545f4aae9addbcd74b83b86caad17c97552f29e`
- Fixed tarball: `ed1e7386d6418ddffeaacff20771359527926d51286a3c3b3d827527745dd953`
- Baseline extracted `dist/index.js`: `c5694504a177b54b4b6cde4b0a7d6f7907605381a4c57a20d55ced9875d3d81d`
- Fixed extracted `dist/index.js`: `301e1ca8df1d569872d514f7e1fd3a64fff07144bfe044f219158456a54e1d44`

</details>

## Validation

- `bun test`: 1,694 passed, 0 failed; 3 snapshots and 3,759 expectations across 102 files.
- `bun run lint`: passed.
- `bun run typecheck`: passed.
- `bun run build`: passed.
- `bun run verify:release`: passed machine-path scanning, npm packing, clean temporary installation, and package entrypoint import.
- `bunx biome check <16 changed TypeScript files>`: passed.
- `git diff --check upstream/master`: passed.

`bun run check:ci` also checks formatting for the entire repository and reports pre-existing formatter drift in unchanged `src/utils/env.test.ts` on current master. This PR does not modify that file; all changed TypeScript files pass the same Biome check. The repository's PR CI uses `bun run lint`, which passes.

### DIFF
--- a/docs/background-orchestration.md
+++ b/docs/background-orchestration.md
@@ -37,6 +37,7 @@ The required native/background-control tools are:
 | `task(..., background: true)` | Start a specialist in the background and immediately return a task ID |
 | hook-driven completion | OpenCode injects terminal background task results automatically |
 | `cancel_task` | Plugin-provided tool to cancel a tracked background task by task ID or Background Job Board alias |
+| `wait_for_user` | Plugin-provided orchestrator tool that pauses automatic continuation while the user performs external manual work |
 
 If these are not available, the scheduler cannot use the default background
 workflow. Configure the environment variable through the installer or use the
@@ -345,18 +346,32 @@ prompts and todo updates do not rearm.
 Continuation is suppressed when the SDK reports the parent or any direct child
 as active, when a terminal child result has not yet been reconciled, during
 foreground fallback, while OpenCode is waiting for a question or permission
-response, or whenever SDK data is unavailable or malformed. A matching reply, or
-a rejected question, clears that wait but does not itself inject a nudge; the
-normal session lifecycle decides whether a later nudge is needed.
+response, after the orchestrator calls `wait_for_user`, or whenever SDK data is
+unavailable or malformed. A matching reply, or a rejected question, clears its
+tool-backed wait but does not itself inject a nudge; the normal session lifecycle
+decides whether a later nudge is needed.
+
+For external manual work, the orchestrator first gives the user concrete steps,
+then calls `wait_for_user` as its final tool action. This explicit signal covers
+text-only HITL turns without attempting to infer intent from assistant prose. The
+wait remains armed across hook/plugin recreation in the same process and is
+cleared only by a distinct real external user message or genuine session
+deletion. Re-observing the user message that preceded the wait, synthetic/internal
+messages (including foreground-fallback replays), fallback teardown, session
+errors, and idle/busy events do not clear it. Immediate choices, clarifications,
+and pasted command output continue to use the `question` tool. If
+`wait_for_user` is intentionally listed in `disabled_tools`, the orchestrator
+uses the `question` tool as the blocking boundary instead.
 
 This is a best-effort runtime check, not a scheduler or persisted state. The
-one-attempt guard is process-local (shared across hook instances in the same JS
-process via an internal gate). It does not survive process restart or
-cross-process boundaries. Recreating a hook/plugin instance inside the same
-process does **not** rearm a consumed epoch; only a real external user message
-(or genuine session deletion) does. After a process restart, the in-memory job
-board cannot establish prior result reconciliation, and the SDK's current
-session/todo status remains the liveness authority.
+one-attempt guard and explicit `wait_for_user` state are process-local (shared
+across hook instances in the same JS process via an internal gate). They do not
+survive process restart or cross-process boundaries. Recreating a hook/plugin
+instance inside the same process does **not** rearm a consumed or waiting epoch;
+only a distinct real external user message (or genuine session deletion) does.
+After a process restart, the in-memory job board cannot establish prior result
+reconciliation, and the SDK's current session/todo status remains the liveness
+authority.
 
 ### Background Job Board Injection
 
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -39,15 +39,23 @@ Fast, structural code search and refactoring - more powerful than plain text gre
 | Tool | Description |
 |------|-------------|
 | `cancel_task` | Cancel a tracked background specialist task by native task ID or Background Job Board alias |
+| `wait_for_user` | Pause automatic incomplete-todo continuation until the next distinct external user message |
 
 `cancel_task` is orchestrator-only. It only cancels background tasks tracked for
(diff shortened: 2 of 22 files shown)


alvinunreal:
thanks @nettee

PR #870, opened 2026-07-22
fix: add reconcile_task tool to prevent self-amplifying task reconciliation loop

### LINKED ISSUE #863: [Bug]: Terminal task reconciliation can self-amplify into new unreconciled tasks
## Summary

I am the coding assistant filing this report after encountering the failure during a code-review session.

A completed background specialist task remained shown as **unreconciled** in the Background Job Board. The scheduler instruction then required reconciliation before a final response, but there is no explicit reconciliation operation. Calling `task()` to acknowledge/reconcile without a `task_id` starts a **new** task session. That new session completes and itself becomes another terminal task to reconcile, creating a self-amplifying loop.

I made the first orchestration mistake by omitting `task_id`; the product issue is that this dangerous path is easy to enter and lacks an explicit, non-recursive reconciliation mechanism or guardrail.

Expected: receiving/processing a terminal task result should mark it reconciled, or the system should offer an explicit reconciliation API that cannot create another task.

## Steps to reproduce

1. Start one or more background `task()` calls.
2. Wait for a task to complete and return a result.
3. Observe the task as completed but unreconciled in the Background Job Board, while the scheduler instruction says terminal tasks must be reconciled before final response.
4. Attempt to reconcile it with `task()` but accidentally omit `task_id` (or follow an ambiguous instruction that does not provide a dedicated reconciliation tool).
5. The call creates a fresh specialist session instead of reconciling the original task.
6. Its completion produces another terminal task, repeating the same blocker.

Observed session sequence: initial read-only review lanes `ora-1`, `ora-2`, and `exp-1`; then accidental fresh Oracle sessions (`ora-3` onward) created while trying to reconcile. Each returned `Reconciled.` but created another completed terminal job. No write-capable agent was involved and no repository files were changed.

## Suggested fix

- Auto-reconcile a terminal task once its completion is delivered to and consumed by the parent agent.
- Add a dedicated `reconcile_task(task_id)` operation that changes board state only; it must not invoke a specialist.
- If reconciliation is implemented through `task()`, require `task_id` for resume/reconcile mode and reject a missing ID with an actionable error rather than silently creating a new task.
- Make task mode explicit (`start` vs `resume` vs `reconcile`) and show the target task ID in the confirmation.
- Avoid blocking final responses solely because a terminal result has not passed through an implicit reconciliation transition.

## Environment

- OpenCode: `1.18.4`
- OS: macOS (Darwin)
- oh-my-opencode-slim version: not surfaced in this session
- Model: `bifrost-openai/openai/gpt-5.6-terra`

## Extra context

Relevant scheduler guidance in the session included both: “Before final response, reconcile any terminal jobs shown in the Background Job Board” and the instruction to resume a specialist by passing its existing `task_id`. The first instruction defines a required state transition, but no direct state-transition tool was available; the second is easy to miss when the task call itself is used as the only available mechanism.

kaze-gif:
### Summary

The orchestrator prompt instructs "reconcile any terminal jobs shown in the Background Job Board" before final response, but no tool existed to perform this operation. The only available mechanism was `task()`, which creates new specialist sessions when called without a `task_id`. This created a self-amplifying loop: an orchestrator attempting to reconcile a completed task would accidentally spawn a new session, which itself would complete and become another unreconciled terminal task requiring reconciliation.

The existing idle-based auto-reconciliation (`idle-reconciliation.ts`) is a heuristic - it marks tasks as reconciled when the parent session goes idle, but the orchestrator has no way to do this explicitly. This forces the LLM into a dead end where the only tool available (`task()`) does the opposite of what's needed.

Fixes: #863

### Steps to reproduce

1. Start one or more background `task()` calls.
2. Wait for a task to complete and return a result.
3. Observe the task as completed but unreconciled in the Background Job Board, while the scheduler instruction says terminal tasks must be reconciled before final response.
4. Attempt to reconcile it with `task()` but accidentally omit `task_id`.
5. The call creates a fresh specialist session instead of reconciling the original task.
6. Its completion produces another terminal task, repeating the same blocker.

### Changes

- New `src/tools/reconcile-task.ts` - state-only tool that marks a terminal task as reconciled via `BackgroundJobBoard.markReconciled()`. Validates calling agent is orchestrator, resolves task_id/alias to board entry, checks terminal-unreconciled state, and marks reconciled. No specialist session is invoked.
- Register in `src/tools/index.ts` and `src/index.ts` (import, instantiation, tools object, `BASELINE_TOOL_NAMES`, health check comment).
- Update orchestrator prompt in `src/agents/orchestrator.ts` - line 213 now reads: *"Before final response, use `reconcile_task` to mark any terminal jobs shown in the Background Job Board as reconciled."*
- Documentation updates in `docs/tools.md`, `docs/background-orchestration.md`, `src/tools/codemap.md`, and root `codemap.md`.

### Tests

Added `src/tools/reconcile-task.test.ts` with 11 focused regression tests:

| Test | What it covers |
|------|----------------|
| Reconcile completed by task ID | Primary success path |
| Reconcile by alias | Alias resolution |
| Reconcile errored | Terminal state coverage |
| Reconcile cancelled | Terminal state coverage |
| Includes reason | Optional field handling |
| Already-reconciled message | Non-terminal jobs |
| Unknown task ID | Missing task |
| Cross-parent rejection | Security isolation |
| Non-orchestrator denial | Agent enforcement |
| Unmanaged session denial | Session scope |
| Requires non-empty task_id | Input validation |

### Validation

- `bun run typecheck` ✅
- `bun run check:ci` (biome) ✅ - 0 errors, 1 info (biome deprecation notice, pre-existing)
- `bun test src/tools/reconcile-task.test.ts` → 11 pass ✅
- `bun test --update-snapshots` → snapshots updated for orchestrator prompt change ✅
- `bun test` (full suite) → 1585 pass / 130 fail (all pre-existing Windows `/tmp` / multiplexer failures - no regressions)

### Scope

The fix is intentionally minimal: a single new tool that closes the reconciliation gap. Idle-based auto-reconciliation (`idle-reconciliation.ts`) continues to work as a fallback. The `cancel_task` tool and all other task lifecycle management are unchanged.

### DIFF
--- a/codemap.md
+++ b/codemap.md
@@ -41,7 +41,7 @@ This codemap covers the plugin repository itself and excludes the nested `openco
 | `src/hooks/json-error-recovery/` | JSON/tool-output recovery helpers for malformed model responses. | [View Map](src/hooks/json-error-recovery/codemap.md) |
 | `src/hooks/phase-reminder/` | Message-transform reminder enforcing orchestrator workflow phases. | [View Map](src/hooks/phase-reminder/codemap.md) |
 | `src/hooks/post-file-tool-nudge/` | Post-read/write reminder path that nudges delegation-aware next steps. | [View Map](src/hooks/post-file-tool-nudge/codemap.md) |
-| `src/hooks/task-session-manager/` | Resumable `task` session tracking, short alias resolution, prompt injection, and stale-session cleanup. | [View Map](src/hooks/task-session-manager/codemap.md) |
+| `src/hooks/task-session-manager/` | Resumable `task` session tracking, short alias resolution, prompt injection, stale-session cleanup, and terminal task reconciliation. | [View Map](src/hooks/task-session-manager/codemap.md) |
 | `src/interview/` | `/interview` feature: per-session and dashboard prompt/state orchestration, persistence, local UI, and cross-process coordination. | [View Map](src/interview/codemap.md) |
 | `src/mcp/` | Built-in MCP registry and per-provider MCP definitions. | [View Map](src/mcp/codemap.md) |
 | `src/multiplexer/` | Terminal multiplexer abstraction layer with backend selection, session mirroring, polling fallback, and shutdown lifecycle orchestration. | [View Map](src/multiplexer/codemap.md) |
@@ -97,7 +97,7 @@ This codemap covers the plugin repository itself and excludes the nested `openco
   multiplexer behavior remains on the upstream path.
 - Council mode is implemented in `src/agents/`; the orchestrator dispatches councillors as subagents and the council agent synthesizes responses.
 - `src/tools/preset-manager.ts` hooks command execution and updates runtime agent models from configured presets.
-- `src/hooks/task-session-manager/` depends on `src/utils/background-job-board.ts` and `src/utils/task.ts` to support background task tracking, task output parsing, and safe alias reuse.
+- `src/hooks/task-session-manager/` depends on `src/utils/background-job-board.ts` and `src/utils/task.ts` to support background task tracking, task output parsing, terminal task reconciliation, and safe alias reuse.
 - `src/hooks/filter-available-skills/` and agent permission logic rely on shared skill names from the CLI/config layer.
 - `src/interview/` hooks into plugin command/event surfaces exposed by `src/index.ts`.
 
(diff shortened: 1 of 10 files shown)


alvinunreal:
nice @kaze-gif

atomlink-ye:
I have the same problem here.

when will it release? the latest version v2.2.8 does not contain this change, although it shows it's released in v2.2.7.

atomlink-ye:
> I have the same problem here.
> 
> when will it release? the latest version v2.2.8 does not contain this change, although it shows it's released in v2.2.7.

https://github.com/alvinunreal/oh-my-opencode-slim/pull/876

this PR revert the change. 
@alvinunreal Could you elaborate on the design intent behind the automatic task continuation logic?
In my use case, the orchestrator takes full control over all task workflows. I don’t intend to trigger any additional prompts for task resumption, and I want to reuse existing task sessions to prevent redundant preloading and repeated file reads.

Currently, subagent session reuse is only allowed once the main session enters an idle state to make the task auto reconcil, which strikes me as an unreasonable constraint.

PR #871, opened 2026-07-22
fix: keep running task tool_result parts byte-stable to protect prompt cache

tsankotsanev:
## Problem

While a background `task()` lane is running, its `tool_result` part sits **mid-history** in the parent orchestrator's request. If those bytes change between consecutive requests (streamed progress snapshots, non-deterministic placeholders), the Anthropic prompt-cache prefix is invalidated at that point — in large orchestrator sessions this re-writes a ~150k-token tail as cache on **every call** while the lane runs (write-never-read loop, observed via request dumps: cache reads frozen while cache writes grew per call).

## Fix

- New `stabilizeRunningTaskParts(messages)` in `board-injection.ts`, invoked at the top of `experimental.chat.messages.transform`.
- It rewrites any `{ type:'tool', tool:'task' }` part in `state: running` to a **deterministic placeholder keyed only on the task ID** (`renderRunningTaskPlaceholder` in `src/utils/task.ts`, round-trips through `parseTaskStatusOutput`).
- Terminal parts (completed/error/cancelled) are never touched — the child's result still materializes exactly once and then stays byte-stable.
- Foreground (`wait: true`) tasks unaffected. Background Job Board injection unaffected (it stays in the trailing volatile zone, which is cache-cheap).

This guarantees the invariant at the plugin layer regardless of how the host streams child progress into running task parts.

## Relation to #870

Complementary: #870 fixes the self-amplifying reconciliation loop (board/state side); this PR fixes byte-stability of running task tool_results in the request body (cache side). No overlap in code paths.

## Tests

- New `running-task-cache-safety.test.ts` (4 tests): running part byte-identical across consecutive transforms while the child progresses; terminal result materializes once then stays stable; running→terminal mutates exactly once, no duplicate completed results; foreground defensive case.
- 2 unit tests for `renderRunningTaskPlaceholder` determinism.
- Full suite: **1721 pass / 0 fail** (`bun test`), `tsc --noEmit` clean, `biome check` clean.

### DIFF
--- a/src/hooks/task-session-manager/board-injection.ts
+++ b/src/hooks/task-session-manager/board-injection.ts
@@ -12,7 +12,12 @@ import type {
   BackgroundJobStore,
   ContextFile,
 } from '../../utils';
-import { isInternalInitiatorPart, parseTaskStatusOutput } from '../../utils';
+import {
+  isInternalInitiatorPart,
+  parseTaskStatusOutput,
+  renderRunningTaskPlaceholder,
+} from '../../utils';
+import { isRecord } from '../../utils/guards';
 import { log } from '../../utils/logger';
 import {
   appendTrailingVolatileMessage,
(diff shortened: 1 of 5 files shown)


alvinunreal:
thanks @tsankotsanev

PR #876, opened 2026-07-22
fix(tasks): prevent duplicate active task launches

alvinunreal:
## Findings
- The task-session manager silently removed task_id for an active known task. The native task call then created a new child session, turning an attempted continuation into a duplicate worker.
- reconcile_task duplicated idle-based automatic reconciliation. Cancellation can also delete a job-board entry before manual reconciliation, producing unknown-or-unowned-task errors.

## Changes
- Fail closed when a caller tries to reuse an active task, rather than stripping its ID and spawning a new session.
- Teach the orchestrator to queue additive amendments, wait for completion, and only resume a task from Reusable Sessions.
- Remove the manual reconcile_task tool, registration, tests, prompts, and documentation. Automatic reconciliation remains the sole lifecycle path.
- Update the deliberate orchestrator-prompt cache snapshot for the new prompt surface.
- No file or write-lock system was added.

## Verification
- bun test
- bun run typecheck


alvinunreal:
@greptileai

PR #880, opened 2026-07-22
fix(task-session): reconcile false-cancelled FG-fallback task parts (#595)

Jiajun0413:
## Problem

When launching a **foreground** task (e.g. `@fixer`) whose child agent has a multi-model fallback chain, and the primary model hits a rate-limit, the ForegroundFallbackManager correctly switches to the fallback model which **completes successfully** — but the parent orchestrator receives **"Task cancelled"** instead of the result.

This is the "orphaned success" symptom of #595.

## Root cause (verified across opencode core + omos)

The FG-fallback path uses `abortSessionWithTimeout` to break opencode's same-model infinite-backoff retry loop (`retry.ts` — no max-attempts), then re-prompts the **same session** with model 2 via `promptAsync`.

The abort poisons the lifecycle:
1. `runState.cancel` (`run-state.ts:77`) calls `cancelBackgroundJobs` → the BackgroundJob the task tool is awaiting is settled as `"cancelled"` (`background-job.ts:148`)
2. `runner.cancel` interrupts the runLoop → `Cause.hasInterruptsOnly` → status `"cancelled"`
3. `background.wait()` returns cancelled → `task.ts:329` throws `"Task cancelled"`

Model 2 then completes on an orphan runLoop (post-abort) with **no awaiter**. The board later records the real outcome via idle reconciliation, but the orchestrator's tool result already says cancelled — **board truth and tool result diverge**.

Key findings that shaped the fix:
- `Agent.Info.model` is a **single** model — opencode has **no native model fallback chain**; it's entirely an omos feature
- `ensureRunning` (`runner.ts:120`) discards new work while a runLoop is Running → omos **must** abort to swap models (no omos-only mechanism can avoid it)
- `background-job.ts` has **no public fulfill/reattach API** (only internal `settle`) → no omos-only mechanism can fix the current turn
- `tool.execute.after` fires on fail but `result=undefined` → opencode writes `status:"error", error:"Tool execution failed: Task cancelled"` (`prompt.ts:414-428`), with `state.metadata.sessionId` pointing at the child session

## Fix

In the `experimental.chat.messages.transform` hook, **after** `stabilizeRunningTaskParts`, detect task tool parts written as `status:'error'` whose error message matches `cancelled` and whose `state.metadata.sessionId` matches a board job in a **non-cancelled terminal state** (completed/reconciled/error). Rewrite them to reflect the board's authoritative terminal outcome.

**Board-gated**: only acts when the board holds completed/error (not cancelled) — true user cancels leave the board in `cancelled` and are preserved unchanged.

**Idempotent**: after rewrite the part is no longer `error` with a cancelled message, so subsequent transforms skip it.

This is the single intentional exception to `stabilizeRunningTaskParts`' "terminal parts are immutable" rule. Cancelled-by-fallback is **not** a genuine terminal outcome — it is a transient artifact of the abort+reprompt lifecycle split.

### Why not the alternatives

| Approach | Verdict |
|---|---|
| **B**: don't abort, other model-swap mechanism | ❌ `ensureRunning` discards new work while Running; no omos-only way to inject model 2 without abort |
| **C**: accept current-turn cancelled, inject completed next turn | ❌ Board already does this, but orchestrator may act on the cancelled signal (retry/abort) before next turn arrives |
| **D**: disable abort for foreground tasks | ❌ Eliminates the race by removing the feature — foreground task is the most valuable scenario for fallback |
| **E**: lifecycle bridge (await board.wait when cancelled) | ❌ BackgroundJob has no public reattach API; would require opencode core changes |

## Tests

New `fallback-false-cancel.test.ts` (7 cases):
- Rewrites false-cancelled error → completed when board has completed truth
- Rewrites false-cancelled error → error when board has error truth
- Does **not** rewrite when board is cancelled (preserves true user cancel)
- Does **not** rewrite when board is still running (no truth yet)
- Does **not** rewrite genuine errors (non-cancellation messages)
- Idempotent across consecutive transforms
- Does **not** touch non-task tool parts

Full suite: **1732 pass, 0 fail**.

## Related

- **#595** "Fallback is detached from background task lifecycle" — this is the root issue. The false-cancel is the orphaned-success symptom.
- **#822** stopped FG from aborting agents with **no** chain (plain `councillor`). Agents **with** chains still use FG abort — this PR fixes the board/tool-result divergence on that path.
- **#765** fixed the early-registration side for single tasks; this PR extends lifecycle protection to the false-cancel scenario.

## AI assistance

Co-developed-with:
- **oracle** (GLM-5.2) — architecture analysis, source-verified root cause, recommended this approach
- **councillor-reviewer-a** (grok-4.5) — independent review, proposed lifecycle-bridge alternative
- **councillor-reviewer-b** (grok-4.5) — independent review, flagged BackgroundJob fulfill-API question (resolved: no public API)
- **councillor-reviewer-c** (grok-4.5) — independent review, proposed disable-abort alternative

> Note: two of the four review agents (oracle, councillor-reviewer-b) themselves hit this exact bug during the session — they completed successfully but their results were reported as "Task cancelled" to the orchestrator. Their results were recovered via session resume, providing live confirmation of the bug this PR fixes.

### DIFF
--- a/src/hooks/task-session-manager/board-injection.ts
+++ b/src/hooks/task-session-manager/board-injection.ts
@@ -16,6 +16,7 @@ import {
   isInternalInitiatorPart,
   parseTaskStatusOutput,
   renderRunningTaskPlaceholder,
+  renderTaskTerminalFromBoard,
 } from '../../utils';
 import { isRecord } from '../../utils/guards';
 import { log } from '../../utils/logger';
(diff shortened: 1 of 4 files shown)


Jiajun0413:
### Reply to Greptile review

**Point 1 — "Failed board jobs lose their required error text" → Fixed (8de0e2e)**

Valid catch. \`opencode\`'s \`message-v2.ts:325-343\` consumes \`part.state.error\` as \`errorText\` for the UI; deleting it on the error-state rewrite dropped the failure reason. The error field is now preserved, set to the board's \`resultSummary\`, matching the normal error-part shape opencode writes (\`prompt.ts:417\`).

**Point 2 — "A fallback still running leaves the false cancellation visible to the current request" → By design, not a bug**

This is a timing constraint inherent to the omos-only fix path, not an oversight:

1. \`background-job.ts\` has **no public fulfill/reattach API** (only internal \`settle\` + public \`cancel\`). omos cannot externally resolve the BackgroundJob the task tool is awaiting.
2. \`ensureRunning\` (\`runner.ts:120\`) discards new work while a runLoop is Running — omos **must** abort to swap models; there is no omos-only way to avoid the cancel.
3. When the board is still \`running\`, model 2 hasn't completed yet — **the truth does not exist** at that moment. Rewriting to a fabricated "completed" would be worse than showing the (spurious) cancellation, because there's no real result to show.

So: while the board is running, the transform correctly leaves the part as-is. Once the board records the real terminal state (completed/error via idle reconciliation, ~2s after model 2 finishes), the **next** transform rewrites the part to the board's authoritative outcome. The false-cancel is only visible for the window between abort and board settlement — a window where no omos-only mechanism can produce a correct result because the fallback model hasn't finished.

This is documented in the code comments and covered by the "does NOT rewrite when board is still running" test case. The only way to close this window is an opencode core change (expose a \`model.swap\` / \`retry.stop\` primitive that doesn't settle the BackgroundJob as cancelled), which is out of scope for an omos-only fix.

**Point 3 — "Settled completed jobs are reconciled as intended" → Confirmed.**

Jiajun0413:
Superseded by #913, which merges this false-cancel fix with the false-complete path from #912 into a single transform-path PR (shared helpers, no dual-branch conflict).

Please review: https://github.com/alvinunreal/oh-my-opencode-slim/pull/913

PR #882, opened 2026-07-22
feat(agents): support description field on custom agent overrides

### LINKED ISSUE #838: [Feature]: Custom agents cannot define a description, causing misleading “manually by the user” text in the Task tool
### Description

Custom agents defined under `agents.<name>` cannot currently provide a concise `description` for the OpenCode Task tool.

The custom-agent configuration supports fields such as `model`, `prompt`, and `orchestratorPrompt`, but it does not support `description`. Because the configuration schema is strict, users also cannot add this property as a local workaround.

When a custom agent is created, `buildCustomAgentDefinition()` does not set a description on the resulting agent definition. OpenCode therefore displays its generic fallback text in the Task tool catalog:

> This subagent should only be called manually by the user.

This fallback can be misleading for custom agents that are specifically intended to be delegated to by the orchestrator. For example, an agent may contain an `orchestratorPrompt` that explicitly tells the orchestrator to delegate builds, tests, or UI workflows to it, while the Task tool simultaneously describes the same agent as manually invoked only. Models may interpret the latter as a routing restriction and refuse to delegate to the agent.

It would be useful to support an optional `description` property for custom agents:

```jsonc
{
  "agents": {
    "executor": {
      "description": "Command execution, builds, tests, linters, formatters, and compact validation",
      "prompt": "...",
      "orchestratorPrompt": "..."
    }
  }
}
```

The configured value could then be propagated to the SDK agent definition and exposed in the Task tool's `subagent_type` catalog:

```text
executor: Command execution, builds, tests, linters, formatters, and compact validation
```

`description` and `orchestratorPrompt` serve different purposes:

- `description` is a concise summary shown in the Task tool schema.
- `orchestratorPrompt` contains detailed delegation and routing instructions for the orchestrator.

Using `orchestratorPrompt` as the Task tool description would not be ideal because it may be lengthy and contain instructions that are unsuitable for a concise catalog entry.

A possible implementation would be to:

1. Add `description: z.string().min(1).optional()` to `AgentOverrideConfigSchema`.
2. Propagate `override.description` from `buildCustomAgentDefinition()` to the resulting agent definition.
3. Include the property in the generated JSON Schema and custom-agent documentation.
4. Add a test confirming that the configured description reaches the SDK agent configuration returned by `getAgentConfigs()`.

Related OpenCode issue describing the same fallback when an agent description is missing or not propagated:

- https://github.com/anomalyco/opencode/issues/27831

### Extra context

_No response_

mhenke:
Fixes #838

Custom agent overrides had no `description` field. Users could not set a custom description for their agents — the `description` property existed on `AgentDefinition` but was not wired up through the config schema.

## What does this change?

Adds `description: z.string().min(1).optional()` to `AgentOverrideConfigSchema`, propagates it through `buildCustomAgentDefinition` (with `Custom subagent '<name>'` fallback) and `applyOverrides` (for built-in agents), regenerates the JSON schema, and documents the field.

## Why this approach?

Matches the existing `displayName` pattern exactly — one schema field, one propagation line per path. No new abstractions, no new dependencies.

## Related work

No duplicate issues or PRs found.

## AI assistance

Orchestrator: opencode/mimo-v2.5-free. 4 Oracle reviews (correctness, over-engineering, excellence, comprehensive) via task() subagents. Human reviewed the full diff and approved.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -124,6 +124,7 @@ Presets can also be switched at runtime without restarting using the `/preset` c
 | `agents.<customAgent>.orchestratorPrompt` | string | - | Exact `@agent` block injected into the orchestrator prompt; must start with `@<agent-name>` |
 | `agents.<agent>.permission` | object \| string | - | Tool-level permission rules enforced by the SDK. See [Agent Permissions](#agent-permissions) |
 | `agents.<agent>.displayName` | string | - | Custom user-facing alias for the agent in the active config |
+| `agents.<agent>.description` | string | generated | Description shown to OpenCode and the orchestrator; defaults to `Custom subagent '<name>'` for custom agents |
 | `acpAgents.<name>.command` | string | - | Command for an external ACP-compatible agent; creates a wrapper subagent named `<name>` |
 | `acpAgents.<name>.args` | string[] | `[]` | Arguments for the ACP agent command |
 | `acpAgents.<name>.env` | object | `{}` | Extra environment variables for the ACP subprocess |
--- a/oh-my-opencode-slim.schema.json
+++ b/oh-my-opencode-slim.schema.json
@@ -104,6 +104,10 @@
               "type": "string",
               "minLength": 1
             },
+            "description": {
+              "type": "string",
+              "minLength": 1
+            },
             "permission": {
               "anyOf": [
                 {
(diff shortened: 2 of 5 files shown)


mhenke:
@greptile-apps

PR #884, opened 2026-07-22
docs(companion): add Wayland click-through known limitation

### LINKED ISSUE #680: [Bug]: Desktop Companion swallows mouse clicks on Ubuntu + GNOME Wayland (no click-through)

What happened: `oh-my-opencode-slim-companion` window blocks all mouse
clicks inside its 120×120 (or user-configured size) bounding box on
Ubuntu 24.04 + GNOME Wayland. The window is transparent, borderless,
and `always_on_top`, so only the animated GIF is visible — but the
entire region still captures pointer events from Wayland, so clicking
a terminal pane (for example a pane running OpenCode TUI inside
[herdr](https://github.com/ogulcancelik/herdr), or any other terminal
multiplexer) underneath the companion is impossible without either
(a) keyboard focus + alt-tab, or (b) `pkill oh-my-opencode-slim-companion`.

Expected: companion should be a truly click-through overlay (status
notifier style) so clicking a terminal pane underneath always lands
on the pane, never on the GIF. The macOS `ActivationPolicy::Accessory`
behavior in `companion/src/main.rs` already captures this intent for
macOS — the same behavior is needed for Wayland compositors that are
GNOME (default Ubuntu desktop), KDE, Sway, Hyprland, cosmic, labwc,
etc., not just the niri-specific code path in `companion/src/niri.rs`.

### Steps to reproduce

1. Ubuntu 24.04 LTS (or 22.04/24.10), GNOME on Wayland
2. `bunx oh-my-opencode-slim install --companion=yes` (or `yes` to
   the interactive prompt)
3. Open any terminal — for example, `herdr` (a Rust terminal
   multiplexer, https://herdr.dev, supports `opencode` as a
   first-class integration), then split panes with
   `Ctrl+b` → `v`; in one pane run `opencode`
4. The companion window auto-spawns anchored at `bottom-right` (or
   wherever `companion.position` is set)
5. Click any pane in the workspace — including the pane immediately to
   the left of the companion, or any tab the companion overlaps — the
   click is consumed by the companion window. Even with the cursor
   visually on top of a terminal cursor's blink, the click never
   reaches the terminal.

Equivalent observation: hovering over the companion window changes
the mouse cursor to the default arrow (not the text caret), confirming
the window is receiving input events.

### oh-my-opencode.json

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/alvinunreal/oh-my-opencode-slim/master/oh-my-opencode-slim.schema.json",
  "preset": "coding-tuned",
  "presets": { "coding-tuned": { /* standard tuned preset */ } },
  "companion": {
    "enabled": true,
    "position": "bottom-right",
    "size": "medium",
    "gifPack": "default",
    "loopStyle": "classic",
    "speed": 1,
    "debug": false
  }
}
```

### OpenCode version

1.17.13

2.1.0

### Operating system

Ubuntu 24.04.4 LTS (Noble Numbat), GNOME on Wayland

```shell
Root cause appears to be in `companion/src/main.rs`:


.viewport: egui::ViewportBuilder::default()
    .with_title("oh-my-opencode-slim-companion")
    .with_app_id("oh-my-opencode-slim-companion")
    .with_decorations(false)
    .with_transparent(true)
    .with_always_on_top()
    .with_active(false)
    .with_inner_size([120.0, 120.0]),


The `event_loop_builder` only sets `ActivationPolicy::Accessory` /
`with_activate_ignoring_other_apps(false)` on **macOS**:


event_loop_builder: Some(Box::new(|builder| {
    #[cfg(target_os = "macos")]
    {
        use winit::platform::macos::{ActivationPolicy, EventLoopBuilderExtMacOS};
        builder.with_activation_policy(ActivationPolicy::Accessory);
        builder.with_activate_ignoring_other_apps(false);
    }
    #[cfg(not(target_os = "macos"))]
    let _ = builder;
})),


On macOS this combination works as intended: an accessory-style window
that does not steal focus from the terminal underneath. On Wayland
compositors (GNOME Mutter, KWin, Sway/Wlroots, Hyprland, etc.), none
of these hints deliver click-through behavior:

- `with_active(false)` only suppresses the initial focus grab; it
  does **not** mark the surface as `pass_through_pointer`.
- `with_always_on_top()` makes things worse — it guarantees the
  surface sits on top of every other surface, but it gives no
  mechanism for clicks to pass through.
- `with_transparent(true)` only enables alpha compositing; it does
  not change Wayland input semantics.
- The `companion/src/niri.rs` module handles niri-specific
  positioning, but does not address input pass-through either (it
  uses `niri msg` for moving, not for surface roles).
- `companion/src/screen.rs::platform_size()` returns `None` on
  everything except macOS, falling back to `[1440.0, 900.0]`, which
  is a separate but related portability smell — the click-through
  fix should be done on the Wayland surface role, not on `screen.rs`.

Click-through on Wayland requires explicit compositor support. The
two workable paths are:

1. **`zwlr-layer-shell` protocol** — implemented by Sway, Hyprland,
   wlroots-based compositors (labwc, wayfire, river, niri),
   KDE Plasma 5.27+, and (since GNOME 47 with `xdg-expire-permission-v1`
   / `wlr-layer-shell-unstable-v1` picked up via xwayland or libdecor),
   GNOME on Wayland via the Xwayland shim. Set
   `layer = Layer::Overlay`, `exclusive_zone = 0`,
   `keyboard_interactivity = None`, and crucially a `region` with
   `pass_through_pointer` so the compositor does not deliver
   pointer events to the surface.

2. **`xdg-popup` / `xdg-positioner`** anchored to the output
   geometry — works on most Wayland compositors without a special
   shell extension, but requires parent surface and is harder to
   reposition dynamically. Not the right primitive for a status
   overlay, but mentioned for completeness.

The right pick for the companion is `zwlr-layer-shell` with
`pass_through_pointer = true` and `keyboard_interactivity = None`.
Reaching `zwlr-layer-shell` requires talking to `wayland-client`
directly or through a wrapper crate — neither `eframe` 0.29 nor
`winit` 0.30 surface layer-shell today, so a small companion-side
addition would be needed (e.g. via `use wayland-client` or a thin
wrapper around `calloop`).

A lighter-weight alternative: when `zwlr-layer-shell` is
unavailable, fall back to a `wlr-layer-shell` XDG surface configured
with `region` covering the **inverse** rectangle of the visible GIF
cells, and `pass_through_pointer = true` everywhere else. That makes
clicks outside the GIF, even inside the window bounding box, pass
through to the underlying surface.

For GNOME specifically (which does not implement `zwlr-layer-shell`
natively), the practical workaround today is to either (a) run GNOME
with the `mutter-wlr-bridge` extension enabled so `zwlr-layer-shell`
clients are lifted into the GNOME shell surface hierarchy, or (b)
gate the companion spawn on `XDG_CURRENT_DESKTOP` matching a
Wayland compositor that does implement `zwlr-layer-shell`. Without
this work, the only sane setting for the companion on Ubuntu + GNOME
is `companion.enabled: false`, which is what I am doing as a
workaround until this is fixed upstream.

Any feedback, or confirmation that a PR with `zwlr-layer-shell`
support + `pass_through_pointer` + a small `companion.platform`
config knob (`auto` | `gnome-wayland` | `niri` | `macos` | `x11` |
`off`) would be welcome would be appreciated.

---

Reproduction log on this host:

$ export OH_MY_OPENCODE_SLIM_COMPANION_DEBUG=1
$ opencode
# run any session — companion auto-spawns bottom-right
$ pgrep -fa oh-my-opencode-slim-companion
198793 /home/nai/.local/share/opencode/storage/oh-my-opencode-slim/bin/oh-my-opencode-slim-companion
# click anywhere in the 120x120 region — clicks never reach herdr pane underneath
$ tail ~/.local/share/opencode/log/oh-my-opencode-slim-companion.198793.log
# (see companion debug output, no click events because they never get delivered)
```

mhenke:
Fixes #680

The companion window blocks mouse clicks on GNOME Wayland, preventing interaction with windows underneath. Users reported this as a bug, but the root cause is a Wayland compositor limitation — the companion uses standard `always_on_top` + `active(false)` which correctly maps to `xdg-toplevel` with `no-focus`, but click-through on Wayland requires `zwlr-layer-shell` with `pass_through_pointer`, a compositor-level primitive our toolkit (egui/winit) does not expose.

The most common Linux desktop (Ubuntu 24.04 with GNOME 46) does not implement `zwlr-layer-shell` natively, so even adding layer-shell support would not fix GNOME. The issue will be addressed when upstream adds native Wayland input region support.

## What does this change?

Adds a "Known Limitations" section to `docs/companion.md` documenting the Wayland click-through issue, affected environments, and workarounds (disable companion on GNOME/KDE Wayland, use compositor window rules on wlroots-based compositors).

## Why this approach?

A documentation fix is the right call here. The actual fix requires 10-16 days of Rust/Wayland work for an optional accessory that the installer defaults to off, and GNOME — the primary affected platform — cannot be fixed with the proposed approach. The docs workaround (set `companion.enabled: false`) immediately solves the user's problem. We can revisit when egui/winit ships `pass_through_pointer` support.

## Related work

- Checked open AND closed PRs/issues: #680 is the only report of this issue
- Related: #680 (this PR documents the known limitation)

## AI assistance

- Model / harness: OpenCode (mimo-v2.5-free) with @skeptic (mimo-v2.5-free) subagent
- Human reviewed the full diff? Yes

## Checklist

- [x] Docs (docs/) updated — this is a docs-only change

### DIFF
--- a/docs/companion.md
+++ b/docs/companion.md
@@ -1,6 +1,6 @@
 # Desktop Companion App
 
-The desktop companion app provides a floating status overlay showing running and active agents.
+The desktop companion is a floating status overlay that shows running and active agents.
 
 ## How to Enable in Configuration
 
@@ -74,8 +74,7 @@ During interactive installation, the installer asks whether to download and
 enable the native Companion binary. The prompt defaults to `no`, so pressing
 Enter skips it.
 
-On niri, Companion can install normally when enabled now that the native binary
-is fixed.
+On niri, Companion installs and works once enabled.
 
 Companion installation is best-effort. If the binary cannot be downloaded or
 installed, the installer prints a warning and continues installing the core
@@ -119,6 +118,40 @@ oh-my-opencode-slim doctor
 
 ---
 
+## Known Limitations
+
+### Wayland Click-Through (GNOME, KDE)
+
+On Wayland compositors, the companion window captures mouse clicks in its
+bounding box, preventing interaction with windows underneath. This affects
+GNOME, KDE Plasma, and other Wayland desktops that do not implement the
+`zwlr-layer-shell` protocol with `pass_through_pointer` support.
+
+**Affected environments:** Ubuntu 24.04+ (GNOME Wayland), KDE Plasma on Wayland,
+most non-wlroots compositors.
+
+**Workaround:** Disable the companion on Wayland desktops that do not support
+layer-shell input pass-through:
+
+```jsonc
+{
+  "companion": {
+    "enabled": false
+  }
+}
+```
+
+On wlroots-based compositors (Sway, Hyprland, labwc) and Smithay-based
+compositors (niri), use compositor window rules to prevent the companion from
+grabbing focus. See the niri section above for an example.
+
+On macOS and X11, click-through works without issues.
+
+Upstream needs to add native Wayland input region support before this can be
+fixed in the companion.
+
+---
+
 ## Expected Binary Install Path
 
 The runtime looks for the companion binary at:
@@ -158,8 +191,7 @@ Custom binaries configured with `companion.binaryPath` are never overwritten.
 
 ## V2 Release Strategy
 
-For the desktop companion app, the release workflow follows the V2 distribution
-plan:
+The release workflow follows the V2 distribution plan:
 
 1. **GitHub Release Assets**: companion binaries are uploaded to the
    `companion-v0.1.3` GitHub release.


mhenke:
@greptile-apps

PR #889, opened 2026-07-23
fix: keep background job board cache-breakpoint-stable by anchoring to tail message

tsankotsanev:
## Problem
The Background Job Board was injected as its own trailing `user` message. The `@ai-sdk/anthropic` SDK coalesces adjacent same-role messages, so the board merged into the preceding user tool_result turn, and core's `applyCaching` (`slice(-2)` message-level `cache_control`) placed the only readable breakpoint on the **volatile board block**. Since the board moves to a new tail every request, the breakpoint never matches the next request → the deepest reusable breakpoint regresses to the system boundary, re-writing the entire tail (~243KB observed) as cache **every call**.

Verified against real request dumps: consecutive same-session pairs diverged exactly at the board position with ~243KB rewritten per call; measured cache-write of 70–95k tokens/call for hours with reads frozen.

## Fix
- Inject the board as a trailing **part on the tail user message** (message count identical to a board-free render, so `slice(-2)`'s breakpoint lands on the previous byte-stable real message). Fall back to a separate trailing user message only when the tail is an assistant message (different role → no merge).
- `stripTailBoardContent`: strip only the previous request's **tail** board; a board found genuinely mid-history is left untouched (removing it would rewrite already-sent bytes).

## Tests
- New `board-cache-breakpoint.test.ts`: models the SDK merge + `slice(-2)` placement; asserts request N establishes a readable prefix N+1 reproduces (old placement establishes zero), board is a trailing part with board-free message count, and previously-sent bytes never change as the conversation grows.
- Updated cache-safety harness/property tests + snapshot.
- Full suite: 1726 pass / 0 fail. `tsc --noEmit` clean. `biome check` clean.

Follow-up to #871 (that PR's first commit — running-task byte-stability — merged; this board-breakpoint commit landed after the merge and was not included).

### DIFF
--- a/src/hooks/__snapshots__/cache-payload.snapshot.test.ts.snap
+++ b/src/hooks/__snapshots__/cache-payload.snapshot.test.ts.snap
@@ -455,16 +455,6 @@ exports[`cache-impact snapshots (update deliberately — see file header) transf
 ,
         "type": "text",
       },
-    ],
-  },
-  {
-    "info": {
-      "agent": "orchestrator",
-      "id": "m09-background-job-board",
-      "role": "user",
-      "sessionID": "ses_cache_safety_fixture",
-    },
-    "parts": [
       {
         "metadata": {
           "oh-my-opencode-slim.backgroundJobBoard": true,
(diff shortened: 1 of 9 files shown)


mhenke:
@greptile-apps

mhenke:
@greptile-apps

PR #895, opened 2026-07-24
fix: guard disabled_* config fields against non-array values

### LINKED ISSUE #894: Plugin load crash: disabledTools.filter is not a function
## Description

Plugin load crashed with:

```
error="disabledTools.filter is not a function. (In 'disabledTools.filter((toolName) => BASELINE_TOOL_NAMES.has(toolName))', 'disabledTools.filter' is undefined)"
```

## Investigation

Traced to `minimumExpectedToolCount(config.disabled_tools)` in `src/index.ts`, called from the health-check block that runs outside the plugin's init `try/catch` — so this is fatal, not a caught/logged warning.

- `disabled_tools` (and the sibling fields `disabled_agents`/`disabled_mcps`/`disabled_skills`) are all declared as `z.array(z.string()).optional()` in the config schema, so a well-formed on-disk config should never produce a non-array value here.
- My on-disk config (`~/.config/opencode/oh-my-opencode-slim.json`) has no `disabled_tools` key at all, and there's no project-level override — so `config.disabled_tools` should statically only ever be `string[] | undefined` post-validation.
- I was not able to reproduce the exact runtime mechanism that produced a non-array value at the crash site. `mergePluginConfigs()`/`deepMerge()` don't appear to construct a scalar for these fields, and the runtime-preset re-init path only touches `config.agents`/`config.preset`.

Regardless of the precise trigger, several call sites across the codebase read `disabled_tools`/`disabled_agents`/`disabled_mcps`/`disabled_skills` with no defensive guard beyond the declared TypeScript type, so any value reaching them via a path other than schema-validated file loading (e.g. a raw/cast `PluginConfig` object) can crash the same way.

## Environment

- oh-my-opencode-slim 2.2.8 (also present at current `master`)
- OpenCode plugin loaded via `~/.config/opencode/opencode.jsonc` → `"plugin": ["oh-my-opencode-slim"]`

## Proposed fix

I have a fix ready (defensive `Array.isArray()` guards at every affected call site, plus a normalization pass in `loadPluginConfig()`) and will open a PR shortly.

major:
## Problem

- Plugin load crashed with `disabledTools.filter is not a function` when `config.disabled_tools` reached the health-check code as a non-array value.
- The `disabled_tools`/`disabled_agents`/`disabled_mcps`/`disabled_skills` fields are all validated as `z.array(z.string()).optional()`, so a well-formed on-disk config should never produce this — the exact runtime mechanism was investigated but not reproduced locally (see #894).
- Several call sites read these fields with no defensive guard beyond the declared TypeScript type, so any value that reaches them by a path other than schema-validated file loading (e.g. a raw/cast `PluginConfig` object) can crash the same way.

## Changes

- Guard every reader of `disabled_tools`/`disabled_agents`/`disabled_mcps`/`disabled_skills` with `Array.isArray()` before `.filter()`/`.includes()`/iteration, falling back to "nothing disabled" instead of throwing:
  - `minimumExpectedToolCount()` and the tool/MCP health-check block in `src/index.ts`
  - `createBuiltinMcps()` in `src/mcp/index.ts`
  - the orchestrator `wait_for_user` flag and `getDisabledAgents()` in `src/agents/index.ts`
  - `getSkillPermissionsForAgent()` in `src/cli/skills.ts`
  - `validateFinalImageRouting()` in `src/config/loader.ts`
- Add a defensive normalization pass at the end of `loadPluginConfig()` that strips any of these fields if they're present but not an array, warning via the existing `onWarning`/`console.warn` path. Documented in-code as currently unreachable via the normal file-loading path (Zod already rejects the whole file in that case) and retained only as defense-in-depth.
- Add regression tests for malformed (`''`, `null`, `{}`) input at every guarded call site.
- Add/complete docstrings on the functions touched by this fix.

No configuration, commands, or user-facing behavior changed for well-formed configs, so no documentation update is needed.

## Verification

- `bun run typecheck`
- `bun test` — 1729 passed, 0 failed
- `bun run check:ci`

Fixes #894

### DIFF
--- a/src/agents/index.test.ts
+++ b/src/agents/index.test.ts
@@ -1166,3 +1166,64 @@ describe('AgentOverrideConfigSchema permission validation', () => {
     expect(result.success).toBe(false);
   });
 });
+
+describe('getDisabledAgents with malformed config', () => {
+  test('falls back to DEFAULT_DISABLED_AGENTS when disabled_agents is not an array', () => {
+    const config: PluginConfig = {
+      disabled_agents: 'not-an-array' as any,
+    };
+    const disabled = getDisabledAgents(config);
+    const expected = getDisabledAgents(undefined);
+    expect(disabled).toEqual(expected);
+  });
+
+  test('falls back to DEFAULT_DISABLED_AGENTS when disabled_agents is an object', () => {
+    const config: PluginConfig = {
+      disabled_agents: { invalid: 'object' } as any,
+    };
+    const disabled = getDisabledAgents(config);
+    const expected = getDisabledAgents(undefined);
+    expect(disabled).toEqual(expected);
+  });
+
+  test('handles valid array normally', () => {
+    const config: PluginConfig = {
+      disabled_agents: ['explorer'],
+    };
+    const disabled = getDisabledAgents(config);
+    expect(disabled.has('explorer')).toBe(true);
+  });
+});
+
+describe('createAgents with malformed disabled_tools', () => {
+  test('does not throw when disabled_tools is not an array', () => {
+    const config: PluginConfig = {
+      disabled_tools: 'not-an-array' as any,
+    };
+    expect(() => createAgents(config)).not.toThrow();
+  });
+
+  test('does not throw when disabled_tools is an object', () => {
+    const config: PluginConfig = {
+      disabled_tools: {} as any,
+    };
+    expect(() => createAgents(config)).not.toThrow();
+  });
+
+  test('orchestrator is created with wait_for_user enabled when disabled_tools is malformed', () => {
+    const config: PluginConfig = {
+      disabled_tools: 'not-an-array' as any,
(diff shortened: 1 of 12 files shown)


major:
Addressed the triage feedback in #894 (comment): moved `minimumExpectedToolCount` and its supporting constants out of `src/index.ts` into a new internal `src/health-check.ts` module, so it's no longer re-exported from the package root and can't be mis-invoked by OpenCode's legacy plugin loader. The `Array.isArray` guard stays as defense-in-depth. Tests moved to `src/health-check.test.ts`; `bun run typecheck`, `bun run check:ci`, and `bun test` all pass (one unrelated pre-existing env-dependent failure in `src/cli/system.test.ts` on this machine, present on the branch before this commit too).

alvinunreal:
thanks @major

PR #606, opened 2026-06-24
fix: keep fallback out of background task sessions

Zaradacht:
## Summary
- Prevent foreground fallback from aborting and re-prompting background task child sessions.
- Track managed background child sessions as soon as they are observed, before task launch output is registered.
- Clean up managed child-session tracking when child or parent sessions are deleted.

## Validation
- `bun test src/hooks/foreground-fallback/index.test.ts src/hooks/task-session-manager/index.test.ts`
- `bun run typecheck`
- `bun run check:ci`
- `bun test`

Fixes #595

PR #912, opened 2026-07-26
fix(task-session): fill false-completed task parts with child session real output (#863)

### LINKED ISSUE #595: [Bug]: Fallback is detached from background task lifecycle, causing lost failures and orphaned successes

Jiajun0413:
## Problem

When a foreground task's primary model halts on a **non-retryable** error (e.g. 403 quota exhausted), the orchestrator receives a **completed** task tool part with an **empty** `<task_result>`, even though the fallback model continues on an orphan runLoop and produces the real result minutes later. The orchestrator reads the empty output, mis-judges the task as failed/empty, and **self-amplifies by launching redundant replacement tasks** (#863).

### Verified root cause (DB + opencode 1.18.5 source)

Complete chain, anchored in opencode core:
1. Model 1 returns 403 `isRetryable:false` → opencode `halt` (`processor.ts:599`) sets `ctx.assistantMessage.error`, publishes `session.error`, sets status `idle`
2. `process` returns `"stop"` (`processor.ts:680`: `if (ctx.assistantMessage.error) return "stop"`)
3. prompt loop breaks → `ops.prompt` returns the last assistant message, which is **empty** (`finish=undefined`, `parts=[]`) because halt produced no content
4. omos `session.error` handler triggers `tryFallback` (non-abort path) → `promptAsync` re-prompts with model 2
5. opencode `runTask` (`task.ts:213`): `result.parts.findLast(text)?.text ?? ""` → **empty string**
6. settle `Exit.succeed("")` (`background-job.ts:148`) → `status:"completed"`, `output:""`
7. parent task part written as `completed` with empty `<task_result>` — **12s after launch**
8. omos `tryFallback`'s `promptAsync` takes effect only after model 1's runLoop ends → model 2 runs on an orphan runLoop, producing **9346 chars over 855s** — but the parent part already says completed+empty

**Root cause is in opencode `runTask`'s `?? ""` fallback**, which treats a halt-empty assistant message as success. omos cannot patch this as a plugin (no hook to intercept `runTask` settle). This is the **false-complete** counterpart to the false-cancel bug in PR #880.

### DB evidence (reproduced session)

| | Time | State |
|---|---|---|
| B-1 launch | +0s | — |
| parent part `completed` + empty output | +12s | `<task_result></task_result>` |
| B-1 child continues | +12s → +867s | **70 parts, 9346-char audit report** |
| orchestrator mis-judges | +645s | "Councillor B errored/empty — retry once" |
| orchestrator relaunches B-2/B-3 | +692s/+693s | from scratch, missing the 9346 chars |

## Fix

In the `experimental.chat.messages.transform` hook, after `stabilizeRunningTaskParts`, scan task tool parts with `status:"completed"` whose parsed `<task_result>` is **empty**. For each, read the child session's real assistant text via the existing `extractSessionResult` helper and in-place rewrite the part output with `renderTaskCompletedWithText` (mirrors opencode `renderOutput` `task.ts:64-76`).

**Idempotent and non-blocking**: if the child session has not yet produced non-empty text (fallback model still running), `extractSessionResult` returns empty and the part is left unchanged — the next transform turn re-evaluates naturally, so no explicit wait/poll is needed. The transform hook only mutates the in-memory `output.messages` (not the persisted DB), so the rewrite is per-turn but the real output eventually lands in history once the child session completes and opencode writes it itself.

**Gated**: only acts on `completed` task parts whose parsed result is empty. True completions with real text are preserved unchanged. Non-task tool parts are not touched.

## Tests

New `false-complete-fallback.test.ts` (5 cases):
- Fills empty completed output with child session real assistant text
- Does **not** rewrite when completed output already has real text
- Does **not** rewrite when child session has no assistant text yet (still running)
- Idempotent across consecutive transforms
- Does not touch non-task tool parts

Full suite: **1719 pass, 0 fail**. Typecheck: pass.

## Relation to other work

| PR/Issue | Scope | Relation |
|---|---|---|
| **#863** (closed, #870 merged) | self-amplify; #870 added `reconcile_task` tool (operation layer) | **This PR fixes the root cause** (#870 only mitigated the operation layer) |
| **#606** (OPEN, Fixes #595) | **background task**: prevent fallback from aborting background task child sessions (preventive) | **Complementary, not conflicting** — #606 is preventive for background tasks; this PR is corrective for **foreground tasks**. File-level changes don't overlap (#606: foreground-fallback + event/options; this: board-injection + transform/import) |
| **#595** (open) | fallback detached from lifecycle | Same family; #606 targets the background-task sub-case, this targets the foreground-task false-complete sub-case |
| **#880** (my OPEN PR) | false-cancel: `error`+`cancelled` parts rewritten from board record | **Mirror counterpart** — #880 handles false-cancel, this handles false-complete. Shared render template shape |

## AI assistance

Co-developed-with:
- **oracle** (GLM-5.2) — read full omos + opencode source, confirmed precise root cause chain (runTask `?? ""` on halt-empty assistant), recommended this approach over idle-reconcile竞态 alternatives

> Note: the analysis council (councillor-reviewer-a/b/c) could not access the omos source tree from their workspace, so their input was based on the root-cause summary rather than direct source verification. The oracle's analysis is the source-verified basis for this fix.

### DIFF
--- a/src/hooks/task-session-manager/board-injection.ts
+++ b/src/hooks/task-session-manager/board-injection.ts
@@ -7,6 +7,7 @@
  * All injection logic must go through the cache-safe helpers in
  * ../cache-safe-injection.ts to ensure prompt cache safety.
  */
+import type { PluginInput } from '@opencode-ai/plugin';
 import type {
   BackgroundJobRecord,
   BackgroundJobStore,
@@ -16,7 +17,9 @@ import {
   isInternalInitiatorPart,
   parseTaskStatusOutput,
   renderRunningTaskPlaceholder,
+  renderTaskCompletedWithText,
 } from '../../utils';
+import { extractSessionResult } from '../../utils/session';
 import { isRecord } from '../../utils/guards';
 import { log } from '../../utils/logger';
 import {
@@ -159,6 +162,82 @@ export function stabilizeRunningTaskParts(messages: unknown[]): void {
       const placeholder = renderRunningTaskPlaceholder(taskID);
       if (state.output === placeholder) continue;
       state.output = placeholder;
+    }
+  }
+}
+
+/**
+ * Reconcile false-completed foreground task tool parts.
+ *
+ * When a foreground task's primary model halts on a non-retryable error
+ * (e.g. 403 quota exhausted), opencode's `halt` produces an empty assistant
+ * message and `runTask` settles the BackgroundJob as `completed` with an
+ * empty output (`result.parts.findLast(text)?.text ?? ""`). omos's
+ * `tryFallback` then re-prompts with the fallback model on an orphan runLoop
+ * that produces the real result, but the parent task part already says
+ * `completed` with an empty `<task_result>` — the orchestrator reads the
+ * empty output, mis-judges the task as failed/empty, and self-amplifies by
+ * launching redundant replacement tasks (#863).
+ *
+ * This in-place rewrites such false-completed parts with the child session's
+ * real assistant text once the fallback model has produced it. The pass is
+ * idempotent and non-blocking: if the child session has not yet produced
+ * non-empty text, `extractSessionResult` returns empty and the part is left
+ * unchanged — the next transform turn re-evaluates naturally, so no explicit
+ * wait/poll is needed. Because the transform hook only mutates the
+ * in-memory `output.messages` (not the persisted DB), the rewrite is
+ * per-turn but the real output eventually lands in history once the child
+ * session completes and opencode writes it itself.
+ *
+ * Gated: only acts on `status:"completed"` task parts whose parsed
+ * `<task_result>` is empty. True completions with real text are preserved
+ * unchanged.
+ */
+export async function reconcileFalseCompleteFallback(
+  state: InjectionState,
+  messages: unknown[],
+  client: PluginInput['client'],
+  directory?: string,
+): Promise<void> {
+  for (const message of messages) {
+    if (!isMessageWithParts(message)) continue;
+    for (const part of message.parts) {
+      if (part.type !== 'tool' || part.tool !== 'task') continue;
+      const partState = part.state;
+      if (!isRecord(partState)) continue;
+      if (partState.status !== 'completed') continue;
+      if (typeof partState.output !== 'string') continue;
+
+      const status = parseTaskStatusOutput(partState.output);
+      if (!status || status.state !== 'completed') continue;
+      // Only reconcile empty results — real completions are intact.
+      if (status.result && status.result.trim().length > 0) continue;
+
+      const childSessionId = status.taskID;
+      const job = state.backgroundJobBoard.get(childSessionId);
+      if (!job) continue;
+
+      // Read the child session's real assistant text. Non-blocking: returns
+      // empty when the fallback model has not yet produced output, and the
+      // next transform turn re-evaluates naturally (idempotent polling).
+      const extracted = await extractSessionResult(client, childSessionId, {
+        directory,
+        includeReasoning: false,
+      });
+      if (extracted.empty) continue;
+
+      const summary = `Background task completed: ${job.description}`;
+      partState.output = renderTaskCompletedWithText(
+        childSessionId,
+        summary,
+        extracted.text,
+      );
+      log('[task-session-manager] reconciled false-completed task part', {
+        taskID: childSessionId,
+        alias: job.alias,
+        parentSessionID: job.parentSessionID,
+        textLength: extracted.text.length,
+      });
     }
   }
 }
--- a/src/hooks/task-session-manager/board-injection.ts
+++ b/src/hooks/task-session-manager/board-injection.ts
@@ -216,14 +216,32 @@ export async function reconcileFalseCompleteFallback(
       const childSessionId = status.taskID;
       const job = state.backgroundJobBoard.get(childSessionId);
       if (!job) continue;
+      // Wait until the board is no longer actively running so we do not
+      // promote mid-generation fallback text into a "completed" part.
+      // False-complete already settles the board as completed+empty while
+      // the orphan runLoop continues — that terminal board state is enough.
+      if (job.state === 'running') continue;
+      if (job.state === 'cancelled' || job.terminalState === 'cancelled') {
+        continue;
+      }
 
-      // Read the child session's real assistant text. Non-blocking: returns
-      // empty when the fallback model has not yet produced output, and the
-      // next transform turn re-evaluates naturally (idempotent polling).
-      const extracted = await extractSessionResult(client, childSessionId, {
-        directory,
-        includeReasoning: false,
-      });
+      // Fail-open: a transient child-session read must never abort the
+      // parent messages transform (Greptile P1). Empty/error → leave part
+      // unchanged; the next transform turn re-evaluates naturally.
+      let extracted: { text: string; empty: boolean };
+      try {
+        extracted = await extractSessionResult(client, childSessionId, {
+          directory,
+          includeReasoning: false,
+        });
+      } catch (error) {
+        log('[task-session-manager] false-complete extract failed', {
+          taskID: childSessionId,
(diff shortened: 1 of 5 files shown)


Jiajun0413:
Superseded by #913, which merges this false-complete fix (including fail-open / tag sanitize / board-running gate hardenings) with the false-cancel path from #880 into a single transform-path PR.

Please review: https://github.com/alvinunreal/oh-my-opencode-slim/pull/913

PR #913, opened 2026-07-26
fix(task-session): unify FG-fallback false-cancel + false-complete reconcile (#595, #863)

Jiajun0413:
## Summary

Supersedes **#880** and **#912** — same FG-fallback orphan-runLoop family, one PR, one transform path.

When foreground fallback swaps models, opencode can leave the parent task tool part in a **false terminal** state while the fallback model still produces the real result:

| Symptom | Root | Fix |
|---------|------|-----|
| **False-cancel** (#595) | FG abort poisons awaited BackgroundJob → `"Task cancelled"` error part | Rewrite from board terminal truth (`renderTaskTerminalFromBoard`) |
| **False-complete** (#863) | Non-retryable halt → `runTask` settles `completed` + empty (`?? ""`) while `tryFallback` orphan runLoop continues | Fill empty `<task_result>` from child session (`renderTaskCompletedWithText`) |

Both passes run after `stabilizeRunningTaskParts` in `experimental.chat.messages.transform`. Shared `sanitizeTaskResultBody` avoids non-greedy parse truncation on recovered text.

### Hardening (from #912 review / Greptile)

- **Fail-open** try/catch around `extractSessionResult` — never abort parent transform
- **Skip** rewrite while board job is still `running` or `cancelled`
- **Sanitize** embedded `</task_result>` / `</task_error>` in recovered body

### Not in scope

- Does **not** force `task_id` resume (still LLM/prompt Session Reuse)
- Does **not** fix opencode core `runTask` empty settle (omos-only)
- **#606** (keep fallback out of BG sessions) remains complementary and separate
- **#870** (`reconcile_task`) is board state only — not redundant with this

## Tests

- `fallback-false-cancel.test.ts` (from #880)
- `false-complete-fallback.test.ts` (from #912 + hardening cases)
- `task.test.ts` sanitize/round-trip

`bun test` on the three files: **35 pass**. `bun run typecheck`: pass.

## Supersedes

- Closes the need for open **#880** and **#912** — please close those as superseded by this PR.
- Related: #595, #863 (closed; root false-complete still needed), #606 (complementary)

## AI assistance

Co-developed-with: oracle, councillor-reviewer-a/b/c, council synthesis (review of #912 hardenings).

### DIFF
--- a/src/hooks/task-session-manager/board-injection.ts
+++ b/src/hooks/task-session-manager/board-injection.ts
@@ -16,6 +16,7 @@ import {
   isInternalInitiatorPart,
   parseTaskStatusOutput,
   renderRunningTaskPlaceholder,
+  renderTaskTerminalFromBoard,
 } from '../../utils';
 import { isRecord } from '../../utils/guards';
 import { log } from '../../utils/logger';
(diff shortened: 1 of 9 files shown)


mhenke:
Closing as superseded. Both issues this fixes (#595 and #863) were closed by the terminal-evidence work that landed upstream in #1227 and 2d975663: false cancels now promote the foreground waiter to background before an abort, and terminal task publication is gated on runtime evidence instead of self-amplifying. All seven files this PR touches were rewritten in that work, so there is no clean path to merge it. Thanks for the detailed symptom table; it mapped the failure modes cleanly.

PR #920, opened 2026-07-28
feat: add regex patterns for model availability errors

FrancoStino:
### What changed
- Added payload-based detection for provider/model availability issues, such as:
  - `model not available`
  - `model is not available`
  - `unsupported model`
  - `unknown model`

### Why
- Application-level errors such as validation failures or missing fields were incorrectly triggering fallback.
- Some providers may still return recoverable model/provider availability errors wrapped in a `400`, so those cases should still be handled through payload matching instead of status code alone.

### Result
- Normal application-level `400` responses no longer trigger fallback.
- `400` responses that explicitly indicate provider/model unavailability can still trigger fallback as intended.
- Existing failover behavior for `429`, `403`, and outage-related `5xx` responses remains unchanged.


alvinunreal:
thanks @FrancoStino

PR #923, opened 2026-07-30
fix(logger): fall back to stderr when file logging fails (#907)

### LINKED ISSUE #907: Logger fails silently, no stderr fallback
## Description

The logger in `src/utils/logger.ts` silently fails in multiple scenarios:

- If the log directory cannot be created (permissions, disk full), `logFile` stays null and every subsequent `log()` call silently no-ops (line 91).
- If the log file cannot be created/opened, same result (lines 67-74).
- If `appendFile` fails during writing, the error is swallowed (lines 105-109).

The logger is the primary debugging surface for the entire plugin. When it fails silently, there is zero diagnostic information available for any other issue.

## Expected behavior

When file logging fails, the logger should fall back to stderr and warn the user once that file logging is unavailable.

## Actual behavior

Logging silently vanishes. Bug reports come in with no log file and no way to diagnose what happened.

## Fix

On first init failure, warn once to stderr. Future `log()` calls fall back to `console.error`:

```typescript
// In initLogger, after the file creation try/catch:
if (!logFile) {
  console.error('[oh-my-opencode-slim] log dir unwritable, falling back to stderr');
  logFile = '__stderr__';
}

// In log(), check for the sentinel:
if (target === '__stderr__') {
  console.error(logEntry);
  return;
}
```

## Affected file

`src/utils/logger.ts`, lines 63-110

## Skill / Command

N/A

---
This issue was created using the /report-issue skill.

MyGO-Mujica:
Fixes #907

The logger previously swallowed initialization and file-write failures silently.
When the log directory could not be created, the session log file could not be
opened, or an asynchronous append failed, subsequent diagnostic output could
disappear without any stderr signal.

A fallback must also avoid letting a stale queued write failure disable a log
file created by a newer `initLogger()` call.

## What does this change?

- Model the logger sink explicitly as `uninitialized`, `file`, or `stderr`.
- Fall back to stderr for initialization and active-generation file-write
  failures, with one warning per continuous fallback period.
- Guard fallback transitions with a generation counter so stale queued writes
  cannot replace a newer file sink.
- Keep fallback output best-effort even when `console.error` itself throws.
- Add regression coverage for initialization, open, append, queued-write,
  stale-write, and stderr-failure paths.
- Update the utility codemap with the fallback behavior.

## Why this approach?

The explicit sink state keeps the fallback modes unambiguous, while the
generation counter localizes the asynchronous race guard to the logger. This
preserves the existing best-effort logging contract without exposing a new API.

## Verification

- [x] `bun test src/utils/logger.test.ts` — 19 pass, 0 fail
- [x] `bun x biome check src/utils/logger.ts src/utils/logger.test.ts`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] `bun run check:ci` — 0 errors, 1 info (biome deprecation notice, pre-existing; run with LF line endings)

> Windows note: a full `bun test` run in this working tree reports 1583/1720 pass with 137 pre-existing failures in unrelated path, interview, and multiplexer tests — identical to the master baseline (1577/1714 pass, 137 fail). This PR adds 6 new tests (all passing) with zero regressions. The Logger-specific suite, typecheck, and build pass for this change.

### DIFF
--- a/src/utils/codemap.md
+++ b/src/utils/codemap.md
@@ -51,7 +51,7 @@ Centralized utilities and shared abstractions used across the oh-my-opencode-sli
 1. Plugin initializes logger with session ID via initLogger(sessionId)
 2. Logs are appended to `~/.local/share/opencode/log/oh-my-opencode-slim.<sessionId>.log`
 3. Old logs (>7 days) are automatically cleaned up on initialization
-4. Log writes are queued to avoid blocking, with errors silently ignored
+4. Log writes are queued to avoid blocking. File logging falls back to stderr after initialization failure or a write failure in the active generation; stale queued writes cannot replace a newer sink
 
 ### Session Operations
 1. Council dispatch uses `promptWithTimeout()` to send prompts with configurable timeout
--- a/src/utils/logger.test.ts
+++ b/src/utils/logger.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import { afterEach, beforeEach, describe, expect, spyOn, test } from 'bun:test';
 import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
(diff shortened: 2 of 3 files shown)


alvinunreal:
thanks @MyGO-Mujica

PR #934, opened 2026-08-01
fix(smartfetch): suppress css-tree lexer warnings corrupting the TUI (#931)

### LINKED ISSUE #931: [Bug]: WebFetch css-tree warnings leak to stderr and corrupt the OpenCode TUI

Image

Running the built-in `webfetch` tool against CSS-heavy pages can print repeated `css-tree` lexer warnings directly into the OpenCode terminal:

```text
[csstree-match] BREAK after 15000 iterations
[csstree-match] BREAK after 15000 iterations
[csstree-match] BREAK after 15000 iterations
```

The messages bypass the structured tool result and appear over the lower-left input/status area of the OpenCode TUI. Resizing the terminal forces a redraw, but the warning should never be written directly into the host TUI in the first place.

Expected behavior:

- dependency diagnostics are captured, logged, or returned as structured WebFetch metadata;
- `webfetch` must not write unsolicited text to the OpenCode process stderr/stdout;
- a CSS validation limit should not visually corrupt the input area.

This is distinct from #578. That report involved council output and a general rendering issue. This report has a deterministic warning string and a concrete dependency path from WebFetch.

### Steps to reproduce

1. Start OpenCode with `oh-my-opencode-slim` enabled.
2. Run several `webfetch` calls, especially concurrently, against CSS-heavy dynamic pages. For example:

   ```text
   https://artificialanalysis.ai/models/gpt-5-6-sol
   https://artificialanalysis.ai/models/gpt-5-6-terra
   https://artificialanalysis.ai/models/claude-opus-5
   ```

3. Let WebFetch parse the HTML with its default `extract_main: true` behavior.
4. Observe one or more `[csstree-match] BREAK after 15000 iterations` lines drawn over the OpenCode input/status area.
5. Resize the terminal; the stale text disappears after the forced redraw.

The fetch itself may still complete successfully. The defect is the direct warning leak and resulting TUI corruption.

### Root-cause evidence

The current plugin WebFetch implementation creates JSDOM documents in `extractFromHtml()`:

```js
const dom = new JSDOM(html, { url: finalUrl });
// extract_main creates a second JSDOM instance
const readerDom = new JSDOM(html, { url: finalUrl });
```

Current local dependency versions:

```text
oh-my-opencode-slim 2.2.8
jsdom 29.1.1
css-tree 3.2.1
OpenCode 1.18.9
```

`css-tree/lib/lexer/match.js` defines:

```js
const ITERATION_LIMIT = 15000;
```

and, when the limit is hit, directly invokes:

```js
console.warn('[csstree-match] BREAK after ' + ITERATION_LIMIT + ' iterations');
```

That global `console.warn()` runs inside the OpenCode/plugin process, so the text reaches the terminal outside the normal tool-output pipeline.

Upstream css-tree issue with the same warning:

- https://github.com/csstree/csstree/issues/294

Related but different prior TUI report:

- #578

### Possible fixes

Any fix should avoid globally muting unrelated warnings, especially because multiple WebFetch calls can run concurrently. Possible approaches:

1. Upgrade or override `css-tree` when an upstream fix is available.
2. Avoid CSS stylesheet parsing in the WebFetch extraction path, since Readability/text extraction does not require page styles.
3. Isolate HTML extraction in a worker/subprocess whose stderr is captured.
4. Add a narrowly scoped diagnostic sink around the JSDOM/css-tree path without replacing global `console.warn` in the shared process.
5. Add a regression test asserting that CSS-heavy HTML does not write to process stderr during `extractFromHtml()`.

### Test impact

Needs a focused regression test using a saved CSS-heavy HTML fixture that triggers the css-tree iteration guard. Assert that:

- WebFetch still returns text/markdown or a structured extraction error;
- no `[csstree-match]` text reaches process stderr;
- concurrent extractions do not interfere with each other's diagnostics.

2.2.8

### OpenCode version

1.18.9

### Operating system

macOS

The warning is not present in the normal OpenCode or plugin logs because it is emitted directly through `console.warn`. The visible terminal text is exactly:

```text
[csstree-match] BREAK after 15000 iterations
```

No credentials, account identifiers, or private page content are involved.

MyGO-Mujica:
Fixes #931

webfetch against CSS-heavy pages makes css-tree print `[csstree-match] BREAK after 15000 iterations` directly into the OpenCode host terminal. The warning is emitted through a global `console.warn` inside the plugin process, bypassing both jsdom's virtualConsole and the plugin's structured tool output, so it leaks into the host process stderr and corrupts the TUI input/status area until a terminal resize.

## What does this change?

- Wrap both `new JSDOM()` calls in `extractFromHtml()` with a scoped `console.warn` guard (`withCssTreeWarningsSuppressed`) that filters only the `[csstree-match]` prefix.
- Restore the original `console.warn` via `try/finally` so unrelated warnings (e.g. secondary-model retry logs) pass through untouched and the global is never left in a replaced state.
- Add regression coverage: a CSS-heavy fixture (200-segment box-shadow chain, csstree/csstree#294 repro case) that reliably trips the lexer iteration guard, asserting no `[csstree-match]` text leaks, extraction still returns correct content, unrelated warnings pass through, and `console.warn` is restored.
- Update the smartfetch codemap with the wrapping convention.

## Why this approach?

The warning is emitted synchronously by css-tree during JSDOM construction (`css-tree/lib/lexer/match.js` calls the global `console.warn` directly), so interception must happen at the plugin boundary around the construction itself. jsdom's virtualConsole cannot capture it (it only wraps `window.console`), and the upstream fix (csstree/csstree#319) only adds a `maxIterations` option without removing the warning. A scoped, prefix-filtered guard is the minimal boundary that stops unsolicited host-stderr output without muting other diagnostics.

## Verification

- [x] `bun test src/tools/smartfetch/utils.test.ts` — 5 pass, 0 fail
- [x] `bun x biome check src/tools/smartfetch/utils.ts src/tools/smartfetch/utils.test.ts src/tools/smartfetch/codemap.md`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] `bun run check:ci` — 0 errors, 1 info (biome deprecation notice, pre-existing; run with LF line endings)

> Windows note: a full `bun test` run in this working tree reports 1586/1723 pass with 137 pre-existing failures in unrelated path, interview, and multiplexer tests — identical to the master baseline (1577/1714 pass, 137 fail). This PR adds 3 new tests (all passing) with zero regressions. The smartfetch suite, typecheck, and build pass for this change.

### DIFF
--- a/src/tools/smartfetch/codemap.md
+++ b/src/tools/smartfetch/codemap.md
@@ -11,6 +11,7 @@
 - **Transport/policy split from rendering:** `network.ts` focuses on URL normalization, redirect allowlists, charset/body decoding, header extraction, and llms.txt probing, while `utils.ts` focuses on turning fetched content into cleaned text/markdown/html plus frontmatter and user-facing messages.
 - **Cache keyed by fetch shape:** `cache.ts` keys fetches by URL plus behavior-affecting options (`extract_main`, `prefer_llms_txt`, `save_binary`), while render format is derived from the cached fetch result so text/markdown/html do not force redundant network requests.
 - **Graceful degradation:** missing/invalid `llms.txt`, blocked redirects, metadata-only binary responses, and secondary-model failures all return a usable result instead of throwing away the fetched content.
+- **Warning-scoped JSDOM construction:** any new JSDOM construction or css-tree trigger point must be wrapped in `withCssTreeWarningsSuppressed` (see `utils.ts`) so css-tree lexer warnings never leak into the host process stderr.
 
 ## Data & Control Flow
 
--- a/src/tools/smartfetch/utils.test.ts
+++ b/src/tools/smartfetch/utils.test.ts
@@ -1,5 +1,23 @@
 import { describe, expect, test } from 'bun:test';
-import { extractHeadingsFromMarkdown, joinRenderedContent } from './utils';
+import {
+  extractFromHtml,
+  extractHeadingsFromMarkdown,
+  joinRenderedContent,
+  withCssTreeWarningsSuppressed,
+} from './utils';
+
+// 200 段逗号分隔的 box-shadow 链 —— csstree/csstree#294 的复现案例，
+// 稳定触发 css-tree lexer 15000 迭代上限警告（jsdom 29 + css-tree 3.2.1 已验证）。
+// 若上游修复后不再触发，本测试退化为弱断言（无泄漏仍成立），可移除 helper。
+const CSS_TREE_WARNING_HTML = (() => {
+  const shadows: string[] = [];
+  for (let i = 1; i <= 200; i++) {
+    shadows.push(`${i}px 0 0 -${Math.min(i + 3, 200)}px #cfcfcf`);
+  }
+  return `<!DOCTYPE html><html><head><style>
+.range-block__range::-webkit-slider-thumb { box-shadow: ${shadows.join(', ')}; }
+</style></head><body><article><h1>Hello</h1><p>World</p></article></body></html>`;
+})();
 
 describe('smartfetch/utils', () => {
   test('extracts cleaned headings from markdown', () => {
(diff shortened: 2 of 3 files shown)


alvinunreal:
thanks @MyGO-Mujica

joshua-zyy:
Thanks for #934 — it fixed the `console.warn` / `[csstree-match]` leak, but there's a second path still unsuppressed that produces the same visible symptom. Filing this so it doesn't get closed as a duplicate by mistake.

**Symptom**

`Could not parse CSS stylesheet` still appears as red text near the prompt area when `webfetch`/`smartfetch` parses an HTML page whose `<style>` block contains CSS that `css-tree` can't fully parse (nesting, `:has()`, `light-dark()`, etc.). Resizing the terminal window makes the red line disappear.

**Root cause**

The string is emitted by jsdom, not css-tree's warning path:

`node_modules/jsdom/lib/jsdom/living/css/helpers/stylesheets.js:79` (and `:97`):
```js
const error = new Error("Could not parse CSS stylesheet", { cause: err });
error.sheetText = cssText;
error.type = "css-parsing";
globalObject._virtualConsole.emit("jsdomError", error);
```

When no custom `virtualConsole` is passed, jsdom does `new VirtualConsole().forwardTo(console)` (`lib/api.js`), and `forwardTo` routes `jsdomError` → `console.error(e.message)` (`lib/jsdom/virtual-console.js`). OpenCode's TUI (built on `@opentui/core`) overrides the global `console` and renders every `console.error` into an in-TUI console overlay in red, near the input area. On terminal resize the overlay re-renders and the error line scrolls out — which is why resizing "fixes" it.

**Why #934 didn't catch it**

`withCssTreeWarningsSuppressed` (`dist/index.js:38901`) only patches `console.warn` and only filters the `[csstree-match]` prefix. The jsdom path above goes through `console.error` with a different message (`"Could not parse CSS stylesheet"`) and a `type === "css-parsing"` tag, so it passes straight through.

`extractFromHtml` (`dist/index.js:39169`) creates 3 `JSDOM` instances per fetch, none with a custom `virtualConsole`, so every fetch of a page with unparseable CSS leaks one or more of these errors.

**Suggested fix (two options)**

1. Pass a silenced `virtualConsole` to `JSDOM` in `extractFromHtml`, dropping `css-parsing` errors:
```js
import { VirtualConsole } from "jsdom";
const vc = new VirtualConsole();
vc.on("jsdomError", (e) => {
  if (e?.type !== "css-parsing") console.error(e.message); // keep real errors
});
new JSDOM(html, { url: finalUrl, virtualConsole: vc });
```

2. Or extend `withCssTreeWarningsSuppressed` to also swallow `console.error` calls whose message is `"Could not parse CSS stylesheet"` (mirrors the current `console.warn` approach, minimal diff).

Option 1 is cleaner — it suppresses at the source instead of patching global `console`, and it preserves genuine jsdom errors.

**Repro**
- `webfetch` any modern page with a `<style>` block using nesting / `:has()`.
- Observe red `Could not parse CSS stylesheet` near the prompt.
- Resize the terminal → line disappears.

Environment: opencode 1.18.15, oh-my-opencode-slim 2.2.10, Windows / PowerShell 7.

MyGO-Mujica:
@joshua-zyy Thanks for the detailed analysis — I reproduced the leak path against jsdom 29.1.1 and it's exactly as you described: jsdom emits a `jsdomError` with `type: "css-parsing"` ("Could not parse CSS stylesheet"), and without a custom `virtualConsole` it forwards to `console.error`.

Fixed and merged in #987 using your option 1 — a custom `virtualConsole` now filters `css-parsing` errors at the source on both `extractFromHtml` JSDOM paths, while the css-tree `console.warn` guard stays unchanged. Non-`css-parsing` jsdomErrors are still forwarded (as full error objects, preserving stack/cause).

One correction from my reproduction, in case it helps anyone else: the trigger isn't modern CSS syntax — css-tree 3.2.1 natively parses nesting, `:has()`, and `light-dark()`. In a 12-sample matrix, only genuinely malformed CSS (garbage tokens, broken at-rules, deeply broken declarations) fired the error. So "webfetch any modern page with nesting/:has()" won't reproduce it; malformed inline CSS does.

PR #938, opened 2026-08-02
feat: add Atlas Cloud preset

binyangzhu000-sudo:

- Add a generated `atlas-cloud` preset backed by `deepseek-ai/deepseek-v4-pro`.
- Register Atlas Cloud as an OpenAI-compatible provider using `ATLASCLOUD_API_KEY` and `https://api.atlascloud.ai/v1`.
- Merge provider defaults into the user's OpenCode config without persisting credentials or overwriting endpoint, environment, or model overrides.
- Document the installer flag and add coverage for preset generation and config merging.

This makes the generated Atlas preset usable immediately instead of requiring a separate manual provider definition.

Validation:
- `bun test` (1718 passed)
- `bun test src/cli/providers.test.ts src/cli/config-io.test.ts` (50 passed)
- `bun run typecheck`
- `bun run build`
- Biome checks on changed TypeScript files
- isolated CLI install and generated-config verification
- live Atlas Cloud forced tool-call request (HTTP 200)

`bun run check:ci` also reached formatting checks; it reports two existing issues in unchanged `src/hooks/cache-safety.property.test.ts` and `src/utils/env.test.ts`. The changed files pass Biome. The local machine does not have the `opencode` binary, so verification used the repository installer plus a live provider request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,13 +134,13 @@ bun run build
 
 ### Getting Started
 
-The installer generates both OpenAI and OpenCode Go presets, with OpenAI active by default.
+The installer generates OpenAI, OpenCode Go, and Atlas Cloud presets, with OpenAI active by default.
 
 > [!TIP]
 > Tune the models and agents for your own workflow. The defaults are only a
 > starting point; the plugin is designed for deep flexibility and customization.
 
-To make OpenCode Go active during install, run `bunx oh-my-opencode-slim@latest install --preset=opencode-go` or change the default preset name in `~/.config/opencode/oh-my-opencode-slim.json` after installation.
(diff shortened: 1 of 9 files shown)


binyangzhu000-sudo:
Addressed the existing-installation preset gap in `314109c`:

- existing lite configs now receive missing generated presets without overwriting user-defined fields or same-name preset customizations;
- an explicit `--preset=atlas-cloud` now selects the newly merged preset;
- added config-level and installer-path regression coverage.

Validation: `bun test src/cli/config-io.test.ts src/cli/install.test.ts` (51 passed), `bun run typecheck`, `bun run build`, scoped Biome check, and `git diff --check`. The full-repo Biome check still reports two pre-existing formatting issues outside this PR (`src/hooks/cache-safety.property.test.ts` and `src/utils/env.test.ts`).

mhenke:
Thanks for the contribution!  I'll let @alvinreal decide if this should be a preset.  There is a discussion over Agent models and setups this might be best there https://github.com/alvinunreal/oh-my-opencode-slim/discussions/751

mhenke:
Close

PR #944, opened 2026-08-04
fix(foreground-fallback): trigger fallback for unavailable relay channels (#943)

### LINKED ISSUE #943: [Bug]: Fallback never triggers on relay gateway "No available channel for model X" errors

`isFailoverError()` does not recognize the "no available channel" error family emitted by OpenAI-compatible relay gateways (one-api / new-api and forks). When a relay drops or disables a model, every request to it fails permanently — but `isFailoverError()` returns `false`, so a correctly configured fallback chain is never entered.

**Expected:** primary model fails → chain advances to the configured fallback.

**Actual:** the task dies on the primary model while a configured, healthy fallback sits unused. No fallback is attempted at all.

The relay operator had removed my primary model upstream for unrelated technical reasons, so this was a genuine, persistent provider outage — exactly the case the fallback chain exists for.

### Steps to reproduce

1. Configure a two-model chain whose primary is served by an OpenAI-compatible relay gateway (one-api / new-api style), with `fallback.enabled: true`.
2. Have the relay remove or disable the primary model upstream. (Equivalently: point the primary at a model ID the relay has no channel for.)
3. Dispatch a task to that agent — in my case `librarian`, via a background `task` call.
4. Observe the agent fail on the primary model. The fallback is never tried.
5. Check `~/.local/share/opencode/log/oh-my-opencode-slim.*.log` — there are **zero** `[foreground-fallback]` lines for that session, not even a "chain exhausted" record, confirming the handler returned early at the `isFailoverError()` gate rather than trying and failing.

For comparison, fallback works correctly on this same setup when the error *is* whitelisted — an earlier session in the same install shows a clean switch:

    [foreground-fallback] switched to fallback model
    {"sessionID":"ses_03455ba87ffe...","agentName":"librarian",
     "from":"hx-openai/gpt-5.6-luna","to":"deepseek/deepseek-v4-flash"}

So the chain itself is wired correctly. Only the error classification fails.

### oh-my-opencode.json

```jsonc
Config at the time of the bug (file is named `oh-my-opencode-slim.json`). `hx-openai` / `hx-claude` are custom providers in my `opencode.json` pointing at a relay gateway. Trimmed to the relevant agents; the other four use the same two-element chain shape.

{
  "$schema": "https://unpkg.com/oh-my-opencode-slim@2.2.9/oh-my-opencode-slim.schema.json",
  "preset": "custom",
  "disabled_agents": ["observer", "council"],
  "fallback": {
    "enabled": true
  },
  "presets": {
    "custom": {
      "librarian": {
        "model": [
          { "id": "hx-openai/gpt-5.6-luna" },
          { "id": "deepseek/deepseek-v4-flash", "variant": "max" }
        ],
        "skills": [],
        "mcps": ["context7", "gh_grep"]
      },
      "explorer": {
        "model": [
          { "id": "hx-openai/gpt-5.6-luna" },
          { "id": "deepseek/deepseek-v4-flash", "variant": "max" }
        ],
        "skills": [],
        "mcps": []
      }
    }
  }
}
```

### OpenCode version

1.18.12

2.2.9 

### Operating system

Ubuntu 24.04.4 LTS on WSL2 (kernel 6.6.87.2-microsoft-standard-WSL2, x86_64)  in WIndows 11

## Error surfaced to the user

Retry Error
No available channel for model gpt-5.6-luna under group Codex专用 (distributor) (request id: 202608041033237488110678268d9d66Ml0tpxJ)

## Root cause

In `src/hooks/foreground-fallback/index.ts`, the error text matches **none** of the 36 patterns across the three lists — `RETRYABLE_ERROR_PATTERNS` (L36), `TRANSPORT_MESSAGE_PATTERNS` (L69), `PROVIDER_OUTAGE_PATTERNS` (L77). I ran the exact string against all three:

```
RETRYABLE_ERROR_PATTERNS:   0 match
PROVIDER_OUTAGE_PATTERNS:   0 match
TRANSPORT_MESSAGE_PATTERNS: 0 match
```

The near miss is L85:

```
/\bmodel\b.*\bnot available\b/i,
```

This requires `not available` to appear *after* `model`. The relay says `No available channel for model …` — `available` precedes `model`, and the substring `not available` never appears. Semantically identical, lexically unmatched.

Both event paths funnel through this one gate, so neither can recover:

- `session.error` → `isFailoverError(props.error)`
- `session.status` (`type === 'retry'`) → `isFailoverError(props.error) || isFailoverError({ message: props.status.message })`

The `Retry Error` prefix indicates opencode had already exhausted its own internal retries and re-emitted the provider message; omo-slim then re-tests that same text against the same whitelist and again declines.

Status-code recovery doesn't rescue it either: `extractStatusCode()` (L91) reads only `error.statusCode` and `error.data.statusCode`. Relays typically wrap this failure in a JSON error body whose status doesn't land on either path.

Because the pattern lists are module-level `const`s with no config surface, users can't work around this without patching the bundle. My only remedy was to demote the primary and put `deepseek` at the head of the chain.

## Why this is worth fixing centrally

This is the canonical one-api / new-api "model has no upstream channel" message, not a single vendor's quirk. It's already treated as a distinct provider-outage class by several projects:

- [`lobehub/lobehub`](https://github.com/lobehub/lobehub/blob/canary/packages/model-runtime/src/errors/patterns.ts#L635) — a dedicated `AgentRuntimeErrorType.NoAvailableChannel` covering five phrasings
- [`BenedictKing/ccx`](https://github.com/BenedictKing/ccx/blob/main/backend-go/internal/handlers/common/failover.go#L1171) — matches this string in its failover path
- [`langbot-app/LangBot`](https://github.com/langbot-app/LangBot/blob/master/skills/src/log-guard.ts#L510), [`limecloud/lime`](https://github.com/limecloud/lime/blob/main/src/components/agent/chat/utils/agentRuntimeErrorPresentation.ts#L124), [`esengine/DeepSeek-Reasonix`](https://github.com/esengine/DeepSeek-Reasonix/blob/main-v2/internal/control/errmsg_test.go#L91) — same treatment

Relay gateways are the dominant access path for users in regions without direct provider access, which is also the population most dependent on fallback chains working.

## Proposed fix

Add to `PROVIDER_OUTAGE_PATTERNS` (L77):

```
/\bno\s+available\s+channels?\s+for\s+model\b/i,
/\bno\s+channels?\s+available\s+for\s+model\b/i,
/\ball\s+upstream\s+channels\s+are\s+currently\s+unavailable\b/i,
```

Each is specific to "upstream has no route for this model" and shouldn't collide with genuine misconfiguration. Deliberately excluded: a bare `model_not_found` code, which relays also emit for a genuinely misspelled model name — matching that would cause pointless chain-walking on user typos.

Suggested test, in the existing style of `src/hooks/foreground-fallback/index.test.ts`:

```
test('returns true for relay gateway "no available channel" outage', () => {
  expect(
    isFailoverError({
      message:
        'No available channel for model gpt-5.6-luna under group Codex专用 (distributor) (request id: 2026...)',
    }),
  ).toBe(true);
});
```

## Secondary finding — three inert `fallback.*` keys (happy to split into its own issue)

Three documented `fallback.*` keys are validated by the zod schema but never read at runtime. The only config values that reach `ForegroundFallbackManager` are:

```
new ForegroundFallbackManager(
  ctx.client,
  runtimeChains,
  config.fallback?.enabled !== false,
  config.fallback?.maxRetries ?? 3,
  sessionLifecycle,
);
```

- **`timeoutMs`** — 39 occurrences in the bundle, all within `session.abort` and `webfetch`; none in the fallback path. Its description promises "time before aborting … or swapping to the next fallback model", but fallback is purely event-driven, so a primary that hangs without ever erroring will never fail over. This one is actively misleading: it reads like a hang-detection timeout and isn't.
- **`retryDelayMs`** — appears only in its own schema declaration.
- **`retry_on_empty`** — appears only in its own schema declaration, despite documenting active behavior ("empty provider responses are treated as failures, triggering fallback/retry").

Either wire these up or mark them deprecated the way `runtimeOverride` already is, so the schema stops advertising inert knobs.

MyGO-Mujica:
Fixes #943

When a one-api/new-api relay returns `No available channel for model ...`, `isFailoverError()` does not classify the error as a provider outage.

The error is commonly carried as text or inside an HTTP 400 response body, so it does not reach the existing 5xx status-code path. It also matches none of the current retryable, transport, or provider-outage patterns. As a result, the foreground fallback gate returns false and a configured fallback chain never advances beyond the unavailable primary model.

## What does this change?

- Add a narrow `no available channel` provider-outage pattern.
- Add regression coverage for the reported one-api/new-api error string as a plain string, `message`, and HTTP 400 `responseBody`.
- Preserve `channel not found` as a non-retryable configuration error.
- Add a manager-level regression test showing that `session.error` advances to the next configured fallback model for the reported error family.

## Why this approach?

`No available channel` describes transient relay capacity or routing availability, so it belongs with existing provider-outage detection rather than permanent model or configuration errors.

The pattern is deliberately limited to the full phrase rather than a broad `channel` match. It covers singular and plural relay wording without treating `channel not found`, invalid channel identifiers, or unrelated model errors as fallback candidates. The existing fallback gate, retry budget, and chain selection logic remain unchanged.

## Verification

- [x] `bun test src/hooks/foreground-fallback/index.test.ts` — 65 pass, 0 fail
- [x] `bun run typecheck`

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -171,6 +171,27 @@ describe('isFailoverError', () => {
   test('returns true for "forbidden" (lowercase) in message', () => {
     expect(isRetryableError({ message: 'forbidden' })).toBe(true);
   });
+
+  test('returns true for NewAPI "no available channel" error shapes', () => {
+    const message =
+      'No available channel for model gpt-5.6-luna under group Codex专用 (distributor) (request id: abc123)';
+
+    expect(isRetryableError(message)).toBe(true);
+    expect(isRetryableError({ message })).toBe(true);
+    expect(
+      isRetryableError({
+        data: { statusCode: 400, responseBody: message },
+      }),
+    ).toBe(true);
+  });
+
+  test('returns false for permanent channel-not-found errors', () => {
+    expect(
+      isRetryableError({
+        message: 'channel not found for model gpt-5.6-luna',
+      }),
+    ).toBe(false);
+  });
 });
 
 // ---------------------------------------------------------------------------
@@ -242,6 +263,42 @@ describe('ForegroundFallbackManager session.error', () => {
     ];
     expect(call[0].path.id).toBe('sess-1');
     // Should have picked the next model after anthropic/claude-opus-4-5
+    expect(call[0].body.model.providerID).toBe('openai');
+    expect(call[0].body.model.modelID).toBe('gpt-4o');
+  });
+
+  test('triggers fallback on unavailable provider channel session.error', async () => {
+    await mgr.handleEvent({
+      type: 'message.updated',
+      properties: {
+        info: {
+          sessionID: 'sess-1',
+          providerID: 'anthropic',
+          modelID: 'claude-opus-4-5',
+          role: 'assistant',
+        },
+      },
+    });
+
+    await mgr.handleEvent({
+      type: 'session.error',
+      properties: {
+        sessionID: 'sess-1',
+        error: {
+          message:
+            'No available channel for model gpt-5.6-luna under group Codex专用 (distributor)',
+        },
+      },
+    });
+
+    expect(mocks.abort).not.toHaveBeenCalled();
+    expect(mocks.promptAsync).toHaveBeenCalledTimes(1);
+
+    const call = mocks.promptAsync.mock.calls[0] as [
+      {
+        body: { model: { providerID: string; modelID: string } };
+      },
+    ];
     expect(call[0].body.model.providerID).toBe('openai');
     expect(call[0].body.model.modelID).toBe('gpt-4o');
   });
--- a/src/hooks/foreground-fallback/index.ts
+++ b/src/hooks/foreground-fallback/index.ts
@@ -82,6 +82,7 @@ const PROVIDER_OUTAGE_PATTERNS = [
   /\bupstream outage\b/i,
   /\bprovider outage\b/i,
   /\bprovider unavailable\b/i,
+  /\bno available channel/i,
   /\bmodel\b.*\bnot available\b/i,
   /\bmodel is not available\b/i,
   /\bunsupported model\b/i,


alvinunreal:
thanks @MyGO-Mujica

PR #951, opened 2026-08-05
fix(foreground-fallback): retry same-model on streaming 5xx errors for agents without chains

### LINKED ISSUE #947: Retry on streaming response 502/503 errors from upstream providers
## Problem

Streaming response errors from upstream providers (Nvidia NIM) are not retried. The session goes straight to idle without any retry or fallback.

### Error examples

```
Streaming response failed: [502] Upstream error from Nvidia: ResourceExhausted: Worker local total request limit reached (33/32)
Streaming response failed: [503] The request queue is full.
```

### Impact

- Councillor and explorer subagent sessions fail silently, producing empty results
- The errors are transient (worker overload, queue full) — retrying would likely succeed
- Affects all sessions using free-tier providers backed by Nvidia NIM

### Root cause

`isFailoverError` already detects these errors (502/503 status codes + text patterns). But `tryFallback` returns early for agents without a fallback chain (councillors, explorer, etc.), so no retry occurs.

### Proposed fix

1. Add `/\bupstream error\b/i`, `/\bstreaming response failed\b/i`, `/\brequest queue is full\b/i` to `PROVIDER_OUTAGE_PATTERNS` (defense in depth)
2. Modify `tryFallback` / `tryFallbackWithAbort` to retry with the current model (up to maxRetries) when no fallback chain is configured
3. Limit same-model retries before escalating to chain-based fallback

### Log evidence

```log
timestamp=2026-08-04T23:08:08.192Z level=ERROR message="stream error" agent=councillor-beta mode=subagent error.error="Streaming response failed: [502] Upstream error from Nvidia: ResourceExhausted: Worker local total request limit reached (33/32)"
timestamp=2026-08-04T23:08:08.203Z level=ERROR message=process error="Streaming response failed: [502] Upstream error from Nvidia: ResourceExhausted: Worker local total request limit reached (33/32)"
timestamp=2026-08-04T23:08:08.205Z [task-session-manager] idle/status idle observed ... session went to completed with empty result
```

No foreground-fallback log entry was emitted — the error was detected but skipped because councillor has no chain.

mhenke:
Fixes #947

Streaming response errors from upstream Nvidia NIM (502 ResourceExhausted, 503 queue full, 504 error code 524) fail silently with no retry for agents without a fallback chain (councillors, explorer, etc.). The orchestrator also gets no retry when the session goes from stream error straight to idle without a retry loop. These are transient worker-overload errors — a retry hitting a different worker usually succeeds.

Log evidence:
```
stream error agent=councillor-beta mode=subagent error="Streaming response failed: [502] Upstream error from Nvidia: ResourceExhausted..."
stream error agent=orchestrator mode=primary error="Streaming response failed: [504] error code: 524"
```
No foreground-fallback retry occurred — isFailoverError caught the error but tryFallback bailed because councillor has no chain.

## What does this change?

When a retryable streaming error hits a session with no fallback chain, the foreground-fallback manager now retries the current model up to maxRetries (default 3) with a 5-second dedup window. Respects disableChain — explicitly disabled agents skip same-model retry. Extracted rePromptWithModel() helper from the existing chain-fallback code.

## Why this approach?

Two alternatives considered: (a) add fallback chains to every agent — but councillors would inherit orchestrator chains, switching to the wrong agent. (b) retry at the OpenCode SDK level — but we can not control that, and isRetryable is false for 502/503/504. Same-model retry is the minimal intervention — transient NIM worker overloads resolve quickly by hitting a different worker.

## Related work

- Related: #947

## AI assistance

- Model / harness: Pi (Claude Opus 4.5) with OpenCode (gpt-5.2)
- Human reviewed the full diff? [x]

## Checklist

- [x] bun run check:ci, bun run typecheck, and bun test pass (pre-existing biome lint only, 1852 tests)
- [x] No behavior/command docs changed
- [x] PR targets the master branch

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -1452,8 +1452,14 @@ describe('ForegroundFallbackManager resolveChain cross-agent isolation', () => {
       },
     });
 
-    // oracle has no chain → should not fall back at all
-    expect(mocks.promptAsync).not.toHaveBeenCalled();
+    // oracle has no chain → retries with its own model, never
+    // cross-bleeds into orchestrator's chain.
+    expect(mocks.promptAsync).toHaveBeenCalledTimes(1);
+    const call = mocks.promptAsync.mock.calls[0] as [
+      { model: { providerID: string; modelID: string } },
+    ];
+    expect(call[0].model.providerID).toBe('anthropic');
+    expect(call[0].model.modelID).toBe('claude-opus-4-5');
   });
 
   test('uses cross-agent last-resort only when agent name is unknown', async () => {
@@ -1508,8 +1514,14 @@ describe('ForegroundFallbackManager resolveChain cross-agent isolation', () => {
       },
     });
 
-    // build has no configured chain and must not inherit orchestrator's
-    expect(mocks.promptAsync).not.toHaveBeenCalled();
+    // build has no configured chain → retries with its own model, never
+    // inherits orchestrator's chain.
+    expect(mocks.promptAsync).toHaveBeenCalledTimes(1);
+    const call = mocks.promptAsync.mock.calls[0] as [
+      { model: { providerID: string; modelID: string } },
+    ];
+    expect(call[0].model.providerID).toBe('openai');
+    expect(call[0].model.modelID).toBe('gpt-5.6');
   });
 });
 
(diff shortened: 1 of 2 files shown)


mhenke:
Closing in favor of #1112. That PR implements this same retry as the retry_same_model action in the classifyError rework, with the same exponential backoff, and it also covers #966 and #1113. This branch now conflicts with master, and #1112 supersedes it. If #1112 stalls, this narrower fix is easy to reopen.

PR #958, opened 2026-08-05
fix(foreground-fallback): detect CliProxyAPI auth-unavailable errors (#957)

### LINKED ISSUE #957: [Bug]: CliProxyAPI 503 auth_unavailable not recognized as retryable, fallback chain never consulted

**What happened:** When a model has exhausted its quota in CliProxyAPI (CPA), CPA returns HTTP 503 with body `{"error":{"message":"auth_unavailable: no auth available (providers=..., model=...)","type":"server_error","code":"internal_server_error"}}`. The foreground-fallback manager does not recognize this as a retryable/failover error, so the session retries the same dead model indefinitely and gets stuck — no fallback to the next model in the chain ever occurs.

This affects both foreground sessions and background subagent sessions (e.g. designer dispatched via `task(background: true)`).

**Expected:** The `auth_unavailable` / `no auth available` error should be recognized as a retryable error, triggering the agent's configured fallback chain to advance to the next model.

### Steps to reproduce

1. Configure an agent (e.g. designer) with a multi-model fallback chain where the primary model is routed through CliProxyAPI.
2. Exhaust the primary model's quota on CPA so that CPA disables it temporarily and starts returning 503 `auth_unavailable`.
3. Dispatch the agent (foreground or background).
4. Observe: the session retries the same dead model indefinitely; the fallback chain is never consulted.

### oh-my-opencode.json

```jsonc
{
  "preset": "cpa",
  "presets": {
    "cpa": {
      "designer": {
        "model": [
          "cli-proxy-api/gemini-3.6-flash",
          "cli-proxy-api/gemini-3.5-flash",
          "cli-proxy-api/gemini-3.1-pro",
        ],
        // ...
      },
      "orchestrator": {
        "model": [
          "cli-proxy-api/kimi-k3",
          "cli-proxy-api/glm-5.2",
        ],
        // ...
      },
    },
  },
}
```

### OpenCode version

1.18.13

2.2.10

### Operating system

macOS 27.0

```shell
Root cause analysis:
The isFailoverError() function in dist/index.js checks two things:
1. Status code extraction (error.statusCode ?? error.data?.statusCode): The OUTAGE_STATUS_CODES set includes 503, so if the status code were extracted, it would match. However, OpenCode core wraps CPA's error into a stringified AI_APICallError, and the structured statusCode may not propagate through to the plugin event object.
2. Text pattern matching (when status code extraction fails): Neither RETRYABLE_ERROR_PATTERNS nor PROVIDER_OUTAGE_PATTERNS contains a pattern that matches auth_unavailable or no auth available. The closest pattern is no available channel which does not match no auth available.
Relevant source (dist/index.js:26217-26267):
RETRYABLE_ERROR_PATTERNS = [
  /\b429\b/, /rate.?limit/i, /too many requests/i,
  /quota.?exceeded/i, /usage.?exceeded/i, /ExceededBudget/i,
  /over.?budget/i, /usage limit/i, /overloaded/i,
  /resource.?exhausted/i, /insufficient.?(quota|balance)/i,
  /high concurrency/i, /reduce concurrency/i,
  /monthly usage limit/i, /5-hour usage limit/i,
  /weekly usage limit/i, /\b403\b/, /forbidden/i,
  /blocked by gateway/i
]

PROVIDER_OUTAGE_PATTERNS = [
  /\binternal server error\b/i, /\bbad gateway\b/i,
  /\bgateway timeout\b/i, /\bservice unavailable\b/i,
  /\bupstream outage\b/i, /\bprovider outage\b/i,
  /\bprovider unavailable\b/i, /\bno available channel/i,
  /\bmodel\b.*\bnot available\b/i, /\bmodel is not available\b/i,
  /\bunsupported model\b/i, /\bunknown model\b/i
]
Neither list includes /no auth available/i or /auth_unavailable/i.

Related issues:
- #954 — Foreground fallback never switches model (no user message found during replay). This is a different bug (replay step fails even when the error is recognized), but it compounds the problem: even if auth_unavailable were recognized, #954's no user message found issue would still prevent the actual model switch.
- #955 — Lists several foreground-fallback bugs including a "Chain gap" section noting that extractStatusCode misses cause.statusCode, error.status, and response.status. Fixing that gap might help CPA's 503 status code propagate, but it's not guaranteed since the issue is at the OpenCode core → plugin event boundary.
- #956 — Background subagent sessions don't trigger fallback at all. Even if auth_unavailable were recognized, background sessions would still be stuck.

Suggested fix:
Add the following patterns to RETRYABLE_ERROR_PATTERNS in dist/index.js:
/no auth available/i,
/auth_unavailable/i,
This is a minimal, low-risk change. A more thorough fix would also investigate why CPA's 503 status code doesn't propagate through to the plugin event's error.statusCode, but that requires changes in OpenCode core's error wrapping.
```

MyGO-Mujica:
Fixes #957

CliProxyAPI disables an exhausted upstream and returns HTTP 503 with `auth_unavailable: no auth available (providers=..., model=...)`. The foreground-fallback classifier does not recognize this error family, so the session retries the same dead model indefinitely and never advances to the next model in the configured fallback chain. This affects both foreground and background subagent sessions.

## What does this change?

- Add `no auth available` and `auth_unavailable` to `RETRYABLE_ERROR_PATTERNS`, alongside the existing auth/forbidden signals.
- Cover the reported error string as a plain string, `message`, HTTP 400 `responseBody`, and the full CliProxyAPI JSON response body.
- Add a manager-level regression test showing that `session.error` with this error family advances to the next configured fallback model.

## Why this approach?

`auth_unavailable` means the provider is temporarily unable to serve the model, so it belongs with the existing retryable/outage detection rather than permanent configuration errors. The patterns are deliberately narrow phrases so unrelated errors (`channel not found`, `invalid API key`, etc.) are not treated as fallback candidates. The existing fallback gate, retry budget, and chain selection logic remain unchanged.

## Verification

- [x] `bun test src/hooks/foreground-fallback/index.test.ts` — 67 pass, 0 fail
- [x] `bun run typecheck`

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -208,6 +208,27 @@ describe('isFailoverError', () => {
     ).toBe(true);
   });
 
+  test('returns true for CliProxyAPI "auth unavailable" error shapes', () => {
+    const message =
+      'auth_unavailable: no auth available (providers=cli-proxy-api, model=gemini-3.6-flash)';
+
+    expect(isRetryableError(message)).toBe(true);
+    expect(isRetryableError({ message })).toBe(true);
+    expect(
+      isRetryableError({
+        data: { statusCode: 400, responseBody: message },
+      }),
+    ).toBe(true);
+    expect(
+      isRetryableError({
+        data: {
+          responseBody:
+            '{"error":{"message":"auth_unavailable: no auth available","type":"server_error","code":"internal_server_error"}}',
+        },
+      }),
+    ).toBe(true);
+  });
+
   test('returns false for permanent channel-not-found errors', () => {
     expect(
       isRetryableError({
@@ -313,6 +334,42 @@ describe('ForegroundFallbackManager session.error', () => {
         error: {
           message:
             'No available channel for model gpt-5.6-luna under group Codex专用 (distributor)',
+        },
+      },
+    });
+
+    expect(mocks.abort).not.toHaveBeenCalled();
+    expect(mocks.promptAsync).toHaveBeenCalledTimes(1);
+
+    const call = mocks.promptAsync.mock.calls[0] as [
+      {
+        model: { providerID: string; modelID: string };
+      },
+    ];
+    expect(call[0].model.providerID).toBe('openai');
+    expect(call[0].model.modelID).toBe('gpt-4o');
+  });
+
+  test('triggers fallback on CliProxyAPI auth-unavailable session.error', async () => {
+    await mgr.handleEvent({
+      type: 'message.updated',
+      properties: {
+        info: {
+          sessionID: 'sess-1',
+          providerID: 'anthropic',
+          modelID: 'claude-opus-4-5',
+          role: 'assistant',
+        },
+      },
+    });
+
+    await mgr.handleEvent({
+      type: 'session.error',
+      properties: {
+        sessionID: 'sess-1',
+        error: {
+          message:
+            'auth_unavailable: no auth available (providers=cli-proxy-api, model=gemini-3.6-flash)',
         },
       },
     });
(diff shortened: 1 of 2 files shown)


alvinunreal:
thanks @MyGO-Mujica

PR #962, opened 2026-08-06
remove(delegate-task-retry): delete stale task retry hook

adikpb:
## Summary

Removes `src/hooks/delegate-task-retry/` (~252 lines) and all its wiring. The hook targets a task-tool API surface that no longer exists in OpenCode core.

## Why

Verified against the current native task tool:

- `patterns.ts:7-52` matches 8 error strings absent from the native tool (`Must provide either category or subagent_type`, `Unknown category`, `Skills not found`, `is not allowed. Allowed agents:`). Native errors now read `Unknown agent type: ...` / `Subagent depth limit reached (...)`.
- The hook therefore **never fires on real native errors** — it only ever fires on the `Unknown agent` substring overlap and appends advice based on nonexistent error formats.
- `hook.ts:42` injects a retry example with `category="unspecified-low"`, `run_in_background=false`, `load_skills=[]` — none of these args exist in the native task schema (only `description`, `prompt`, `background`, `subagent_type`, `task_id`). Whenever it fires, it actively teaches the model invalid calls.

This matches the repo's established cleanup pattern of removing plugin features once upstream ships them natively.

## Changes

- Delete `src/hooks/delegate-task-retry/` (hook.ts, patterns.ts, index.test.ts, codemap.md)
- Remove wiring in `src/index.ts` (import, factory call, wrapper, `tool.execute.after` dispatch)
- Remove re-export in `src/hooks/index.ts` and category entry in `src/hooks/codemap.md`
- Biome format fix in `src/hooks/task-session-manager/board-injection.ts` (pre-existing lint regression on base, unrelated; needed for CI green)

## Verification

- `rg` for `delegate-task-retry|delegateTaskRetry|DelegateTaskRetry` in `src/`: zero matches
- `bun run check:ci`: pass
- `bun run typecheck`: pass
- `bun test`: 1849 pass, 0 fail


adikpb:
@greptileai please re-review — the stale codemap references you flagged (codemap.md row + .slim/codemap.json hashes) are fixed in 16a4819.

alvinunreal:
thanks @adikpb

PR #967, opened 2026-08-06
fix(foreground-fallback): detect cannot-connect transport errors

### LINKED ISSUE #956: [Bug]: Background/delegated subagent sessions ignore the configured model fallback chain

**What happened:** When the Orchestrator dispatches a specialist as a **background subagent** (the default background-orchestration workflow), and that specialist's primary model provider is unreachable (connection refused), the session **retries the dead provider indefinitely instead of sliding to the next model in its configured fallback chain**. In our observed run: a `fixer` background session made **75 attempts against a dead `b70` provider with zero failover**, blocking the whole orchestration for 20+ minutes.

**Expected:** The foreground-fallback behavior (which works — see below) should apply to background/delegated sessions too: after `fallback.maxRetries` failures, the session should advance to the next model in the agent's chain (`b70 → opencode-go/deepseek-v4-flash → openrouter/... → deepseek/...` in our config).

**Evidence that the chain itself is correct:** A *foreground* fixer session created directly (same agent, same chain, same dead primary) slid after **1** failed `b70` attempt and completed on `opencode-go`:

```
# foreground session (fallback works):
1 × providerID=b70
2 × providerID=opencode-go   ← slid + completed

# background subagent session (fallback does NOT work):
75 × providerID=b70
3 × stream error
0 × any other provider
```

The `src/hooks/foreground-fallback/index.ts` header says it is "Runtime model fallback for foreground (interactive) agent sessions" and that it "mirrors the same fallback loop used for delegated sessions" — but no equivalent loop appears to exist for delegated/background sessions (grep for fallback handling only finds `foreground-fallback`; the background machinery in `src/utils/background-job-*` handles task tracking, not model failover).

## Steps to reproduce

1. Install slim 2.2.10 on opencode 1.18.13 with a preset where a specialist has a multi-model chain, e.g.:
   ```jsonc
   "agents": {
     "fixer": { "model": ["b70/Qwen36-27B-MTP-TC-Q6_K", "opencode-go/deepseek-v4-flash"] }
   }
   ```
2. Take the primary provider down (stop the local llama.cpp server on `b70`).
3. In the Orchestrator, dispatch the specialist as a background task (the default `task(..., background: true)` path): "have fixer add a comment to README".
4. Observe the opencode log: repeated `stream error ... Cannot connect to API` on `b70`, zero streams on `opencode-go`, task never completes.
5. Compare with a foreground session: `opencode run --agent fixer` (or API `POST /session {agent: fixer}` + message) with the same dead primary → slides to `opencode-go` after 1 attempt and completes.

## oh-my-opencode-slim.json

```jsonc
{
  "preset": "local",
  "presets": {
    "local": {
      "orchestrator": { "model": ["opencode-go/deepseek-v4-flash", "openrouter/nvidia/nemotron-3-ultra-550b-a55b:free", "deepseek/deepseek-v4-flash"] },
      "fixer": { "model": ["b70/Qwen36-27B-MTP-TC-Q6_K", "opencode-go/deepseek-v4-flash", "openrouter/nvidia/nemotron-3-ultra-550b-a55b:free", "deepseek/deepseek-v4-flash"] },
      "explorer": { "model": ["lms/gemma-4-e4b-it-qat", "opencode-go/deepseek-v4-flash"] },
      "librarian": { "model": ["b50/Gemma4-12B-IT-QAT-MTP-Q4_K_XL", "opencode-go/deepseek-v4-flash"] }
    }
  },
  "agents": {
    "fixer": { "model": ["b70/Qwen36-27B-MTP-TC-Q6_K", "opencode-go/deepseek-v4-flash", "openrouter/nvidia/nemotron-3-ultra-550b-a55b:free", "deepseek/deepseek-v4-flash"] },
    "explorer": { "model": ["lms/gemma-4-e4b-it-qat", "opencode-go/deepseek-v4-flash"] },
    "librarian": { "model": ["b50/Gemma4-12B-IT-QAT-MTP-Q4_K_XL", "opencode-go/deepseek-v4-flash"] }
  },
  "fallback": { "enabled": true, "timeoutMs": 30000, "maxRetries": 3 }
}
```

## Environment

- OpenCode version: 1.18.13
- oh-my-opencode-slim version: 2.2.10
- Server: `opencode serve --port 4096` (systemd), `OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true`
- Providers: local llama.cpp (`b70`, `b50`), LM Studio (`lms`), opencode-go, openrouter, deepseek

## Suggested fix

Extend the foreground-fallback manager (or add a parallel loop for delegated sessions) so `session.error` / task-failure events on background subagent sessions also advance the agent's `_modelArray` chain after `maxRetries`. The manager already has all the machinery (chain lookup, tried-model tracking, re-prompt) — the gap appears to be that background sessions are not fed through it. If background failover is intentionally out of scope, documenting the limitation (and ideally exposing a hook/config so orchestrators can react) would still be valuable — right now a dead specialist provider silently stalls the whole orchestration graph.

MyGO-Mujica:

Provider SDKs report connection failures with natural-language messages such as `stream error: Cannot connect to API` that carry no transport code. The `TRANSPORT_MESSAGE_PATTERNS` classifier only matches anchored transport phrases (`connect ECONNREFUSED`, `fetch failed`, ...), so these errors are not recognized as failover-worthy and the session retries the dead provider instead of advancing to the next model in the fallback chain.

## What does this change?

- Add a narrow `cannot connect to api` pattern to `TRANSPORT_MESSAGE_PATTERNS`.
- Cover the plain string and `{ message }` error shapes with the `stream error:` prefix.
- Preserve non-API connection errors (`Cannot connect to database`) as non-retryable.
- Add a manager-level regression test showing `session.error` advances to the next configured fallback model.

## Why this approach?

The pattern is deliberately limited to the API connection phrase so unrelated connection errors (database, etc.) are not treated as fallback candidates. The existing anchored transport patterns and the rest of the classifier remain unchanged.

## Verification

- [x] `bun test src/hooks/foreground-fallback/index.test.ts` — 70 pass, 0 fail
- [x] `bun run typecheck`

> Related: issue #956 reports background subagent sessions never advancing on the same error family (`stream error: Cannot connect to API`). This PR fixes the shared error-classification gap; whether it fully resolves #956 depends on the separate background event-delivery path.

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -229,6 +229,18 @@ describe('isFailoverError', () => {
     ).toBe(true);
   });
 
+  test('returns true for "cannot connect to API" transport errors', () => {
+    expect(isRetryableError('Cannot connect to API')).toBe(true);
+    expect(isRetryableError('stream error: Cannot connect to API')).toBe(true);
+    expect(
+      isRetryableError({ message: 'stream error: Cannot connect to API' }),
+    ).toBe(true);
+  });
+
+  test('returns false for non-API connection errors', () => {
+    expect(isRetryableError('Cannot connect to database')).toBe(false);
+  });
+
   test('returns false for permanent channel-not-found errors', () => {
     expect(
       isRetryableError({
@@ -370,6 +382,41 @@ describe('ForegroundFallbackManager session.error', () => {
         error: {
           message:
             'auth_unavailable: no auth available (providers=cli-proxy-api, model=gemini-3.6-flash)',
+        },
+      },
+    });
+
+    expect(mocks.abort).not.toHaveBeenCalled();
+    expect(mocks.promptAsync).toHaveBeenCalledTimes(1);
+
+    const call = mocks.promptAsync.mock.calls[0] as [
+      {
+        model: { providerID: string; modelID: string };
+      },
+    ];
+    expect(call[0].model.providerID).toBe('openai');
+    expect(call[0].model.modelID).toBe('gpt-4o');
+  });
+
+  test('triggers fallback on cannot-connect session.error', async () => {
+    await mgr.handleEvent({
+      type: 'message.updated',
+      properties: {
+        info: {
+          sessionID: 'sess-1',
+          providerID: 'anthropic',
+          modelID: 'claude-opus-4-5',
+          role: 'assistant',
+        },
+      },
+    });
+
+    await mgr.handleEvent({
+      type: 'session.error',
+      properties: {
+        sessionID: 'sess-1',
+        error: {
+          message: 'stream error: Cannot connect to API',
         },
       },
     });
--- a/src/hooks/foreground-fallback/index.ts
+++ b/src/hooks/foreground-fallback/index.ts
@@ -78,6 +78,10 @@ const TRANSPORT_MESSAGE_PATTERNS = [
   /^request timeout$/i,
   /^connect ECONNREFUSED\b/i,
   /^getaddrinfo ENOTFOUND\b/i,
+  // Provider SDKs also report connection failures with natural-language
+  // messages (e.g. "stream error: Cannot connect to API") that carry no
+  // transport code. Match the narrow phrase only.
+  /cannot connect to api/i,
 ];
 const PROVIDER_OUTAGE_PATTERNS = [
   /\binternal server error\b/i,


MyGO-Mujica:
The `test (latest)` failure is a GitHub Actions infrastructure issue — the runner failed to resolve action download info (`Error: Service Unavailable`) before any test ran. Could you re-run the test job? Greptile Review already passed.

alvinunreal:
thanks @MyGO-Mujica

Issue #966, opened 2026-08-06
foreground-fallback infinite loop when quota-exhausted error is misclassified as retryable


joshua-zyy:
### Summary

When a provider's **quota/usage is exhausted** (a non-transient error), the `foreground-fallback` hook enters an **infinite re-fallback loop**: the fixer agent session repeatedly aborts and re-prompts every ~10 seconds, and the session never completes or fails. The fallback chain is never exhausted; the manager just resets the tried-set and retries forever.

### Environment

- Plugin version: **2.2.10**
- opencode version: 1.18.14
- OS: Windows
- Config: fallback chain `["openai/gpt-5.6-luna", "ollama-cloud/glm-5.2"]` for the `fixer` agent (array form of `presets.<name>.<agent>.model`)

### Root cause

1. **Quota-exhausted errors are classified as retryable.** `RETRYABLE_ERROR_PATTERNS` in `src/hooks/foreground-fallback/index.ts` includes `/usage limit/i` (and similar patterns like `/quota.?exceeded/i`, `/usage.?exceeded/i`, `/monthly usage limit/i`). A "usage limit has been reached" error is therefore treated exactly like a transient 429, even though it will not recover for hours.

   Actual error observed:
   ```
   stream error providerID=openai modelID=gpt-5.6-luna agent=fixer
   error.error="AI_APICallError: The usage limit has been reached"
   ```

2. **The chain-exhaustion path loops forever instead of aborting.** In `execFallback` (dist/index.js ~26527-26554):

   ```js
   let nextModel = chain.find((m) => !tried.has(m));
   if (!nextModel) {
     if (chain.length > 1) {
       // resetting tried set for re-fallback ...
       tried = new Set;
       tried.add(primary);
       if (currentModel && currentModel !== primary) tried.add(currentModel);
       this.sessionTried.set(sessionID, tried);
       nextModel = stickyFallback;   // chain[chain.length - 1]
     } else { ... abort ... }
   }
   ```

   After both models in the chain have been tried, instead of aborting, it resets the tried-set and re-prompts with the sticky fallback model — **unconditionally and without any counter/limit**. Meanwhile the re-prompt is issued with the agent's configured model (the chain head `gpt-5.6-luna`), which fails again, re-triggering fallback. This produces the observed infinite loop.

   Log evidence (oh-my-opencode-slim log, repeated every ~10s forever):
   ```
[...]

PR #976, opened 2026-08-07
fix(foreground-fallback): replay v2 user messages and detect quota-threshold errors

### LINKED ISSUE #954: [Bug]: Foreground fallback never switches model — `no user message found` during replay while core retries 5xx forever

When a provider returned HTTP 500 (`Internal Server Error: No available Claude accounts support the requested model`), I expected the orchestrator's configured **model fallback chain** to kick in and retry the request on the next model in the chain.

Instead, the fallback **never switched models**. The session stayed stuck retrying the **same** model indefinitely — the TUI showed `Retrying (attempt 8) · Internal Server Error: …` and kept climbing (attempt 9, 10, …) for ~25 minutes across 20+ attempts. No fallback model was ever used.

Two root causes confirmed from logs + source:

1. **The foreground-fallback *did* trigger** (it saw the retryable error), but its replay step failed every time with `no user message found`, so it aborted without switching models.
2. **OpenCode core retries 5xx on the same model forever** (`session/retry.ts`: *"5xx errors are transient server failures and should always be retried"*, exponential backoff 2s→30s, **no attempt cap**), so the loop never terminates and the fallback is retried against the same broken model endlessly.

I expected at least one of: (a) the fallback successfully re-prompts on the next chain model, or (b) core to cap 5xx retries so a terminal error surfaces.

### Steps to reproduce

1. Configure `fallback` (enabled) and an agent with a model chain (see config below).
2. Route the agent's primary model to a relay/provider that intermittently returns **HTTP 500**.
3. Run a prompt that reaches that agent (e.g. delegate to `designer`).
4. Observe: the session retries the **same** model forever (`Retrying (attempt N)`); the plugin logs `no user message found` and never switches to the next fallback model.

### oh-my-opencode.json

```jsonc
"fallback": {
  "enabled": true,
  "maxRetries": 2
}
```

### OpenCode version

1.18.10

2.2.10

### Operating system

  macOS (darwin) 26.3 (25D125)

```shell
**Plugin log — fallback triggers but replay fails every time** (designer subagent, primary `17zy/kimi-k3`, 7-model chain configured):

[2026-08-05T08:45:50.286Z] [foreground-fallback] no user message found {"sessionID":"ses_02ee7af4…"}
[2026-08-05T08:46:16.441Z] [foreground-fallback] no user message found {"sessionID":"ses_02ee7af4…"}
…(repeats for every retry, ~25 min)…
[2026-08-05T08:54:53.205Z] [foreground-fallback] resetting tried set for re-fallback {"sessionID":"ses_02ee7af4…","agentName":"designer","currentModel":"17zy/kimi-k3","prevTried":["17zy/kimi-k3","17zy/qwen3.8-max","openai_my/gpt-5.6-sol","17zy_openai/gpt-5.6-sol","openai_my/gpt-5.6-luna","17zy_openai/gpt-5.6-luna","17zy_openai_compat/qwen3.7-max"],"nextModel":"17zy_openai_compat/qwen3.7-max"}

Note it burned through the **entire** chain (`prevTried` lists all 7 models) yet never actually switched — every `execFallback` died at the "find last user message" step.

**OpenCode core log — same model retried forever:**

08:45:50 level=ERROR message="stream error" providerID=17zy modelID=kimi-k3 agent=designer error.error="AI_APICallError: Internal Server Error"
08:45:54 …  08:46:16 …  08:46:32 …  08:47:04 …  08:48:08 …  08:50:17 …  (continues to 09:07)


**Proof the mechanism *can* work** (a different process did switch successfully):

[2026-08-05T08:44:42.933Z] [foreground-fallback] switched to fallback model {"sessionID":"ses_02ee8e2b…","agentName":"fixer","from":"17zy/deepseek-v4-flash","to":"17zy_openai_compat/deepseek-v4-flash"}


**Verified against a live server:** `GET /session/{id}/message` returns the user message correctly (both idle and busy), and a standalone `createOpencodeClient` call fetches it fine. So the server/messages endpoint is healthy.

**Suspected root cause (please confirm):** in `getClient()` (foreground-fallback), the plugin builds a **fresh** client:

createOpencodeClient({ baseUrl: input.serverUrl.origin, directory: input.directory })

instead of using the already-configured `input.client` that OpenCode passes into the plugin (which carries OpenCode-configured transport (auth headers where required, in-process fetch fallback when the server has no external listener / in-process `fetch`). If `serverUrl.origin` is stale or points at a different concurrent instance (multiple `opencode` processes were running), `session.messages()` returns empty/errors → `result.data ?? []` → `[]` → `no user message found` → fallback aborts. Two suggested directions:

1. Use `input.client` (or otherwise reuse OpenCode's configured client) for `session.messages()` / `promptAsync()` in the foreground-fallback instead of constructing a bare client.
2. (Core, FYI) OpenCode's unbounded 5xx retry means even a working fallback races a loop that never terminates; capping 5xx retries would make failures surface.
```

adevwithpurpose:

Fixes the plugin-side root cause of #954: foreground fallback never switches models because replay of the last user message always fails.

### Bug 1 — `no user message found` on every fallback (closes #954)

`execFallback` located the last user message with `isUserMessageWithParts`, which requires the v1 shape `{ info: { role: 'user' }, parts }`. Since OpenCode 1.18 the plugin SDK's `session.messages()` returns v2 `SessionMessage` objects (`{ type: 'user', text, time, id }` — verified in `@opencode-ai/sdk@1.18.13` `SessionMessageUser`), so the finder never matched and replay aborted with `no user message found` every time. This is why the reporter saw 20+ identical failures over ~25 minutes against a long-persisted message — the fetch succeeded, but the shape never matched.

- Add `isReplayableUserMessage` / `partsFromReplayMessage` in `src/hooks/types.ts` accepting both the v1 transform shape and the v2 `session.messages()` shape.
- Use them in `execFallback`; v2 user text is replayed as a `text` part alongside the existing internal-initiator marker.
- When no replayable message exists, log `messageCount` and `requestError` (from `result.error`) so a failed request is no longer conflated with an empty session (per the request in #954).

### Bug 2 — codex quota-threshold errors never trigger fallback

The codex provider reports quota exhaustion as `All codex accounts reached configured quota threshold (reset after 20h ...)`. That matched no `RETRYABLE_ERROR_PATTERNS` entry, so the chain was never consulted and the same dead model was retried. Add `/quota.?threshold/i` to the pattern list.

Note on #966: the quota-exhaustion *infinite re-fallback loop* reported there is a separate defect in the chain-exhaustion path (sticky re-prompt without a limit), not in error classification. Falling back to the next model when one model's quota is exhausted is the desired behavior; #966's fix belongs in the reset/re-prompt branch.

## Verification

- `bun run typecheck` — passes
- `bun run check:ci` — passes (only a pre-existing biome-config-migration info)
- `bun test src/hooks/foreground-fallback/index.test.ts` — 73/73 pass, including 3 new tests:
  - `returns true for codex quota-threshold errors`
  - `replays the last user message from v2-shaped session.messages data`
  - `prefers the latest user message across mixed v1/v2 shapes`
- Full suite: 1598 pass / 78 fail — identical failure set to the pristine base commit (pre-existing Windows/platform-specific failures in multiplexer/snapshot tests, verified via `git stash` comparison), so this change adds 3 passing tests and zero new failures.

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -121,6 +121,20 @@ describe('isFailoverError', () => {
     );
   });
 
+  test('returns true for codex quota-threshold errors', () => {
+    expect(
+      isFailoverError({
+        message:
+          'AI_APICallError: [codex/gpt-5.6-sol-medium] All codex accounts reached configured quota threshold (reset after 20h 41m 59s)',
+      }),
+    ).toBe(true);
+    expect(
+      isFailoverError(
+        'AI_APICallError: [codex/gpt-5.6-sol-medium] All codex accounts reached configured quota threshold (reset after 20h 41m 59s)',
+      ),
+    ).toBe(true);
+  });
+
   test('returns true for "usage exceeded"', () => {
     expect(isRetryableError({ message: 'usage exceeded' })).toBe(true);
   });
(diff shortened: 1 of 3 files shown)


alvinunreal:
thanks @adevwithpurpose

PR #980, opened 2026-08-08
fix(zellij): target panes by ID with fallback and serialized creation (#978)

### LINKED ISSUE #978: [Bug]: Zellij pane targeting actions fail silently in agent-tab mode

What happened: In the Zellij `agent-tab` path, oh-my-opencode-slim v2.2.10 invokes CLI actions that Zellij 0.44.3 does not accept:

```text
zellij action focus-pane --pane-id <paneId>
zellij action rename-pane --name <name>
```

In a real attached Zellij terminal pane, both commands exit with code 2:

```text
error: The subcommand 'focus-pane' wasn't recognized
error: Found argument '--name' which wasn't expected
```

`crossSpawn()` resolves ordinary non-zero process exits, while `runInPane()` does not check the exit codes of these actions before reporting success. As a result, the first reused pane is not reliably targeted or renamed, but the multiplexer can still report `{ success: true }`.

I also observed that `zellij action current-tab-info --json` returned `No active tab found for current client` from the OpenCode/Zellij pane-child context. This makes the agent-tab path unable to reliably determine which tab to restore after creating subsequent panes.

Expected: Pane actions should use Zellij-supported, pane-ID-targeted commands; a failed required attach write must not be reported as a successful pane spawn; and tab routing should not rely on the transient CLI client's active-tab context.

### Steps to reproduce

1. Start OpenCode inside Zellij 0.44.3 with the configuration below.
2. Trigger delegated work so OMO creates or reuses the `opencode-agents` tab.
3. From a terminal pane in the same Zellij session, run:

   ```sh
   zellij action focus-pane --pane-id "$ZELLIJ_PANE_ID"
   zellij action rename-pane --name attached-probe
   zellij action current-tab-info --json
   ```

4. Observe that the first two actions are rejected by Zellij's CLI parser. In the attached-pane reproduction, `current-tab-info` also returned `No active tab found for current client`.

For comparison, Zellij 0.44.3 accepts pane-targeted forms such as:

```sh
zellij action rename-pane attached-probe -p "$ZELLIJ_PANE_ID"
zellij action write-chars 'echo test' -p "$ZELLIJ_PANE_ID"
```

### oh-my-opencode.json

```jsonc
{
  "multiplexer": {
    "type": "zellij",
    "layout": "main-vertical",
    "zellij_pane_mode": "agent-tab"
  }
}
```

### OpenCode version

1.18.15

2.2.10

### Operating system

Fedora 44 container with a real attached Zellij terminal pane

The invalid `focus-pane --pane-id` and `rename-pane --name` argv are present in the released `v2.2.10` source at `src/multiplexer/zellij/index.ts`.

MyGO-Mujica:
Fixes #978

In the Zellij multiplexer, oh-my-opencode-slim invoked CLI actions that every published Zellij release rejects:

- `zellij action focus-pane --pane-id <id>` — subcommand doesn't exist
- `zellij action rename-pane --name <name>` — `--name` isn't a valid argument

Both exit with code 2 on Zellij 0.44.x. `crossSpawn()` resolves ordinary non-zero exits, and `runInPane()` ignored the exit codes, so the first reused pane was not reliably targeted or renamed while the multiplexer still reported success.

`current-tab-info --json` is also client-bound and fails from pane child processes ("No active tab found for current client"), so the agent-tab path could not reliably determine which tab to restore after creating subsequent panes.

## What does this change?

- Require Zellij >= 0.44.1 via `zellij --version` parsing (`new-pane --tab-id` only exists in 0.44.1+); older releases make the multiplexer unavailable instead of silently misbehaving.
- Drop the invalid `focus-pane` call and target panes directly by ID: `rename-pane <name> -p <paneId>` and `write-chars <chars> -p <paneId>`; a failing attach write no longer reports `{ success: true }`.
- Replace `current-tab-info` with `list-panes --json --tab --all`, resolving the parent tab from the pane's `ZELLIJ_PANE_ID`.
- Retry a directed `new-pane` once without `--direction` when Zellij silently drops it (a tab is crowded after ~4 splits), so pane creation keeps working instead of failing.
- Serialize pane-creation sequences through a queue so concurrent sub-agent spawns cannot interleave focus-switching (verified: concurrent cross-tab creates drift the attached client's focus).
- Restore the parent tab after the initial `new-tab`, and only cache successful parent-tab lookups so a transient failure is retried.

## Why this approach?

`focus-pane --pane-id` and `rename-pane --name` are invalid in every published Zellij version, so there is no legacy syntax to fall back to — pane-ID-targeted forms are the only working API. `current-tab-info` is client-bound and unusable from pane children; the pane list JSON carries stable `tab_id`s, which lets the adapter derive the parent tab from the parent pane ID. The directed-create retry and the creation queue were both driven by reproduced behavior in a real Zellij 0.44.3 session (silent drops after 4 directed splits; focus drift under concurrent cross-tab creates).

## Verification

- [x] `bun test src/multiplexer/zellij/index.test.ts` — 31 pass, 0 fail
- [x] `bun x biome check src/multiplexer/zellij/index.ts src/multiplexer/zellij/index.test.ts src/multiplexer/zellij/codemap.md docs/multiplexer-integration.md`
- [x] `bun run typecheck`
- [x] Real Zellij 0.44.3 reproduction in a Fedora 44 container: serial creation of 5 panes keeps the attached client on the parent tab; panes 5-6 fall back to directionless creation after directed splits are dropped.

Note: Zellij cannot run on Windows, so the runtime verification was done in a Docker Fedora 44 container with a real attached Zellij session.

### DIFF
--- a/docs/multiplexer-integration.md
+++ b/docs/multiplexer-integration.md
@@ -241,7 +241,7 @@ Please analyze this codebase and create a documentation structure.
 | Multiplexer | Status | Notes |
 |-------------|--------|-------|
 | **Tmux** | ✅ Supported | Full layout control with `main-vertical`, `main-horizontal`, `tiled`, and more |
-| **Zellij** | ✅ Supported | Creates a dedicated `opencode-agents` tab by default; can open panes in the parent OpenCode tab with `zellij_pane_mode: "current-tab"`; maps `main-*` layouts to pane directions |
+| **Zellij** | ✅ Supported | Requires Zellij **0.44.1 or newer** (the adapter gates on `zellij --version` and skips itself on older releases; 0.44.0 lacks `new-pane --tab-id`). Creates a dedicated `opencode-agents` tab by default; can open panes in the parent OpenCode tab with `zellij_pane_mode: "current-tab"`; maps `main-*` layouts to pane directions |
 | **Herdr** | ✅ Supported | Splits panes in the current Herdr workspace; maps `main-vertical`/`even-horizontal`/`tiled` layouts to right splits and `main-horizontal`/`even-vertical` to down splits; no layout rebalancing (like Zellij) |
 | **cmux** | ✅ Supported | Requires cmux 0.64.14+ (0.64.17+ recommended); creates the agent column to the right and stacks subsequent agents downward without moving focus |
 | **Kitty** | ✅ Supported | Uses `kitten @ launch` to open new windows; requires `allow_remote_control` **and** `listen_on` in kitty.conf (OpenCode must run inside a kitty window; kitty exports `KITTY_LISTEN_ON` which the plugin passes through to reach kitty from detached subagent processes). No layout rebalancing (like Zellij/Herdr) |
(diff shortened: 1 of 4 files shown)


MyGO-Mujica:
@greptile-apps please recheck — the availability race is fixed in d938a56: \isAvailable()\ now caches the in-flight probe promise, so an overlapping availability check awaits the same probe instead of observing a premature false.

alvinunreal:
thanks @MyGO-Mujica

PR #987, opened 2026-08-09
fix(smartfetch): suppress jsdom css-parsing errors leaking into the TUI

MyGO-Mujica:
Related: #934

PR #934 suppressed css-tree lexer warnings leaking through `console.warn`, but there is a second, independent leak path with the same visible symptom: jsdom emits a `jsdomError` with `type: "css-parsing"` ("Could not parse CSS stylesheet") when a page's CSS is malformed. Without a custom `virtualConsole`, jsdom forwards that error to `console.error`, which the OpenCode TUI renders as red text near the prompt.

The trigger is not modern CSS syntax: css-tree 3.2.1 natively parses nesting, `:has()`, and `light-dark()`. The error fires only for genuinely malformed CSS (garbage tokens, broken at-rules, deeply broken declarations) that any browser would reject.

## What does this change?

- Adds `withJsdomCssParsingErrorsSuppressed()`, which installs a `VirtualConsole` that drops `css-parsing` errors and forwards all other `jsdomError`s to `console.error`.
- Both `JSDOM` construction paths in `extractFromHtml` now pass that `virtualConsole`, alongside the existing css-tree `console.warn` guard (kept unchanged).
- The lazy `getJSDOM()` dynamic import was replaced by a static import; the plugin already probes jsdom at init time, so startup behavior is unchanged.

## Why this approach?

Filtering at the source via `virtualConsole` is cleaner than extending the global `console.warn` patch to also swallow `console.error` calls: it keeps genuine jsdom errors visible and avoids growing the global-patch approach. Non-`css-parsing` jsdomErrors (e.g. `resource-loading`) still reach `console.error`.

## Verification

- [x] `bun test src/tools/smartfetch/utils.test.ts` — 10 pass, 0 fail
- [x] `bun test src/tools/smartfetch/` — 22 pass, 0 fail
- [x] `bun run typecheck`
- [x] `bun x biome check src/tools/smartfetch/utils.ts src/tools/smartfetch/utils.test.ts`
- [x] Independent runtime check: pre-fix path leaks 1 `console.error` for malformed CSS; post-fix `extractFromHtml` (both paths) emits 0 and still extracts content; clean CSS emits 0.

### DIFF
--- a/src/tools/smartfetch/utils.test.ts
+++ b/src/tools/smartfetch/utils.test.ts
@@ -4,6 +4,7 @@ import {
   extractHeadingsFromMarkdown,
   joinRenderedContent,
   withCssTreeWarningsSuppressed,
+  withJsdomCssParsingErrorsSuppressed,
 } from './utils';
 
 // 200 段逗号分隔的 box-shadow 链 —— csstree/csstree#294 的复现案例，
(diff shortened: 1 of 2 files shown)


MyGO-Mujica:
@greptile-apps please recheck — the P2 metadata comment is fixed in c419ca5: non css-parsing jsdomErrors are now forwarded as full error objects (stack, cause, URL preserved) instead of message-only.

PR #993, opened 2026-08-10
fix(smartfetch): dedupe cache size accounting for llms.txt results (#992)

### LINKED ISSUE #992: [Bug]: smartfetch cache size accounting charges llms.txt and plain-text results 4x
﻿## What happened, and what did you expect?

### What happened

`smartfetch`'s webfetch cache (`src/tools/smartfetch/cache.ts`) overcharges cache size for llms.txt and plain-text results by 4x.

For those result shapes, `rawContent` / `html` / `markdown` / `text` all point at the same content (see `src/tools/smartfetch/tool.ts`: the llms.txt path sets all four fields to the same `text`, and the non-HTML branch runs `cleanFetchedText(rawText)` four times). But the `sizeCalculation` sums `Buffer.byteLength` of all four fields independently, so a 128KiB page is charged 512KiB.

Consequence: the 50MiB `CACHE` only holds ~12.5MiB of real content for these result shapes. Entries are evicted early and the same page is re-fetched over the network repeatedly within the 15-minute TTL window.

Measured with a real `LRUCache` (50MiB `maxSize`, 500 × 128KiB llms.txt-shaped entries):

- Old accounting: 99 entries retained (~12.5MiB effective)
- Correct accounting: 399 entries retained (~50MiB effective)

HTML results are unaffected — their four fields genuinely differ and are charged once each.

### What did you expect?

Cache accounting should charge bytes once per distinct string value, so the 50MiB budget reflects actual retained content. llms.txt and plain-text pages should not be charged 4x for the same content.

### Steps to reproduce

1. Run OpenCode with oh-my-opencode-slim's enhanced `webfetch` enabled.
2. Fetch a page that resolves via `llms.txt` (or any plain-text page), with enough content to be cached.
3. Inspect the cache retention: entries are evicted at ~1/4 of the intended capacity (e.g. 128KiB pages only ~99 fit in the 50MiB cache instead of ~400).

N/A — deterministic accounting bug, verified with a standalone LRUCache reproduction.

### Suggested fix

Dedupe by string value in `sizeCalculation` (Set-based, SameValueZero), and reuse a single `cleanFetchedText` pass in the non-HTML branch of `src/tools/smartfetch/tool.ts`.

MyGO-Mujica:
﻿Fixes #992

`smartfetch`'s cache `sizeCalculation` charges bytes four times for llms.txt and plain-text results. Those results point all four fields (`rawContent`/`html`/`markdown`/`text`) at the same content, so the 50MiB `CACHE` only holds ~12.5MiB of real content and evicts entries early — the same page gets re-fetched over the network repeatedly.

Measured: 500 × 128KiB llms.txt-shaped entries keep only **99** in the 50MiB cache with the old accounting, versus **399** with deduped accounting (~4x effective capacity for these result shapes).

## What does this change?

- [x] Extract `calculateCacheSize` in `src/tools/smartfetch/cache.ts` and charge bytes once per distinct string value (Set dedupe, SameValueZero) instead of summing all four fields.
- [x] Reuse a single `cleanFetchedText` pass in the non-HTML branch of `src/tools/smartfetch/tool.ts` so the four fields share one reference (behavior unchanged, cleaning cost 4x → 1x).
- [x] Add regression tests for llms.txt / plain-text / HTML / binary shapes plus a real LRUCache capacity check.

HTML results (four genuinely different fields) are still charged for all four — only duplicated content is deduped.

## Why this approach?

The cache's job is to estimate retained memory; identical content is held once, so charging it four times overstates usage by 4x and defeats the 50MiB budget. Value-based dedupe is exact for both the same-reference (llms.txt) and same-content-different-reference (pre-fix plain-text) shapes, and leaves HTML accounting untouched.

## Verification

- [x] `bun test src/tools/smartfetch/cache.test.ts` — 7 pass
- [x] `bun test src/tools/smartfetch/` — 38 pass (full dir regression)
- [x] `bun run typecheck`
- [x] scoped Biome check on the 3 changed files
- [x] Independent reproduction: old 524288B vs new 131072B per 128KiB entry; real 50MiB LRUCache 99 → 399 entries
- [ ] `bun run check:ci` / full `bun test` — not run on Windows (known CRLF noise; scoped checks passed)

### DIFF
--- a/src/tools/smartfetch/cache.test.ts
+++ b/src/tools/smartfetch/cache.test.ts
@@ -1,5 +1,7 @@
 import { describe, expect, test } from 'bun:test';
-import { buildCacheKey } from './cache';
+import { LRUCache } from 'lru-cache';
+import { buildCacheKey, calculateCacheSize } from './cache';
+import type { BinaryFetch, CachedFetch, FetchResult } from './types';
 
 describe('smartfetch/cache', () => {
   test('includes save_binary but not format in the cache key', () => {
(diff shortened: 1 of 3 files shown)


alvinunreal:
thanks @MyGO-Mujica

PR #1009, opened 2026-08-13
feat(tasks): add task_status and task_nudge supervision tools

adevwithpurpose:

Background specialists run detached: the orchestrator has no read-only way to inspect a live child's state, and the only "instruction" channel (`task(..., task_id: ...)`) resumes the child and starts new model work. This adds two safe supervision tools:

- `task_status` — read-only status of a tracked child task (bounded live `/session/status` read, with explicit uncertainty when the host read fails, times out, or returns a malformed entry), with an activity-based `possibly_stuck` flag for `busy`/`retry` children that showed no `message.updated` / `step-finish` / busy-status activity for 120s. Never prompts or mutates the child.
- `task_nudge` — admits a single concise follow-up to a running child via `session.prompt(..., noReply: true)` without resuming, aborting, or starting a new model run. Rate-limited to one nudge per 30s per task, with the slot reserved atomically and rolled back on refusal/failure.

The orchestrator prompt now instructs calling `task_status` for read-only inspection and `task_nudge` once (only when `possibly_stuck`) instead of re-issuing or resuming tasks.

Supporting changes:

- `TaskActivityTracker` records session activity from `message.updated` / `step-finish` / busy `session.status` events keyed by **session id** (never the message id, which `message.updated`/`step-finish` carry in `info.id`) and forgets sessions on `session.deleted`; status timestamps flow into the `possibly_stuck` computation.
- `promptSafe` normalizes Windows backslash paths so Background Job Board renders are stable across platforms.

Both tools are strictly read-only or admission-bounded; completed and reconciled tasks are unaffected.

## Review remediation

Resolves Oracle review findings and Greptile status/nudge notes:

1. **Activity keyed by session ID** — `message.updated` / `step-finish` events carry the message id in `info.id` and the session id in `info.sessionID`; the tracker now keys by session id (extracted `resolveEventSessionID`), so child activity actually refreshes the stuck timer. Event-policy integration tests cover the wiring.
2. **Bounded, explicit status uncertainty** — `task_status` reads live status through the bounded snapshot helper (5s default timeout). A failed, timed-out, or malformed host response is reported as `status_uncertain: true` with `last_status_error` and `state: <board state> (unconfirmed)` — never presented as a confident live status — and never yields `possibly_stuck: true`.
3. **Shared live status/activity policy** — `task_nudge` admission goes through the same policy module as `task_status` (`task-policy.ts`): the child must be board-`running`, live-confirmed `busy`/`retry`, and reported `possibly_stuck`. A stale board record alone never admits a nudge.
4. **Atomic rate-limit reservation** — the per-task nudge slot is reserved synchronously before the first `await`, so concurrent nudges cannot both pass the check; any refusal or transport failure rolls the reservation back so the task is not locked out.
5. **New tests** — cross-parent ownership rejection, status failure/malformed/hang-bounded uncertainty, concurrent-nudge exclusivity, rate-limit rollback, noReply behavioral proof, and event-integration tests for the activity tracker.

## Verification

- `bun test src/tools/task-status.test.ts src/tools/task-nudge.test.ts src/tools/task-activity.test.ts src/hooks/cache-payload.snapshot.test.ts`: 28 pass, 0 fail (3 snapshots)
- `bun test src/hooks/task-session-manager/index.test.ts`: 124 pass, 0 fail (confirms cross-platform board rendering with the `promptSafe` normalization)
- `bun run typecheck`: pass
- `bun run lint`: pass
- `git diff --check`: pass

### DIFF
--- a/src/agents/orchestrator.ts
+++ b/src/agents/orchestrator.ts
@@ -221,6 +221,8 @@ Balance: respect dependencies, avoid parallelizing what must be sequential, and 
 - Before dispatching a specialist, check the Background Job Board and current conversation for an existing task that already covers the objective.
 - \`task_result\` returns only a completed specialist's final assistant message, and can be called by any parent session that owns the task. Never use \`task(..., task_id: ...)\` to fetch output: that resumes the child and starts new model work.
 - Before retrying completed work whose result appears missing or incomplete, retrieve it with \`task_result\`. Dispatch again only when the retrieved result does not satisfy the objective.
+- For a live child task, call \`task_status\` for read-only state inspection. There is no safe live-prompt channel: never use \`task(..., task_id: ...)\` as a progress check or instruction because it resumes model work.
+- If \`task_status\` reports \`possibly_stuck: true\`, use \`task_nudge\` once to admit a concise follow-up without resuming or aborting the child; never use it as a polling loop.
 - Prefer \`task(..., background: true)\` for delegated work that can run independently.
 - For work already chosen for delegation, launch independent specialist lanes in the background so the orchestrator stays unblocked and can reconcile results when they return.
 - Never reissue an unchanged task to the same specialist after a rejection; adjust its scope or context before retrying.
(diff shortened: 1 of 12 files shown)


alvinunreal:
@greptileai

PR #1010, opened 2026-08-13
fix(multiplexer): wait for child session readiness before attach

adevwithpurpose:

Refs code-yeongyu/oh-my-openagent#3505 and anomalyco/opencode#9099 (attach-before-ready races).

When a child session was created, `MultiplexerSessionManager` launched `opencode attach --session <id>` immediately. If the child had not yet been published on the server's `/session/status` endpoint, the attach TUI fell back to the new-session picker — the pane attached to the wrong session or nothing. The same race existed on respawn after an idle close.

This change gates both the initial spawn and the respawn on `waitForSessionReady`:

- probes `GET /session/status` for the child session (`idle`/`running`/`busy`/`retry`) with a per-attempt abort timeout (default 1s) so a hanging probe cannot stall the event handler;
- **absolute bounded deadline (default 2s):** the wait is bounded by a wall-clock deadline from the first probe, not by 8 independent 1s attempts — a hanging status endpoint can no longer stall the sequential event stream for ~10s. Per-attempt timeouts and backoff delays are capped by the remaining budget so the total never exceeds the deadline;
- **lifecycle cancellation/disposal fencing:** cleanup and per-instance disposal abort any in-flight readiness wait via an AbortSignal, so teardown is never blocked by a hanging probe, and a late-ready result after cleanup cannot spawn a stale pane;
- **safe recovery after initial readiness timeout:** a timed-out session stays known and is recovered by a later `busy` status event with a fresh readiness pass; a `busy` observed *during* the in-flight wait is deferred and triggers an immediate recovery pass when the wait fails, instead of being silently orphaned;
- **explicit cmux exclusion contract:** the cmux adapter is excluded from this gate — cmux sessions go through `CmuxSessionLifecycle`, which has its own spawn machinery (server health check plus deferred spawn retry with a TTL). Contract is documented on `waitForSessionReady` and pinned by an adapter-contract test;
- deduplicates concurrent `session.created` events so duplicate creates neither double-wait nor double-spawn;
- exposes `waitForSessionReady` / `SessionReadinessOptions` from the multiplexer index for reuse; probe, delay, deadline clock, and abort signal are injectable for tests.

## Verification

- `bun test src/multiplexer/session-manager.test.ts src/multiplexer/shared.test.ts`: 107 pass, 0 fail (adds deadline-bound, abort-signal, cleanup-during-wait, disposal-during-wait, initial-timeout recovery, busy-during-wait recovery, and cmux adapter-contract tests; 30 pre-existing herdr/kitty/tmux adapter failures confirmed base-red on the base commit)
- `bun run typecheck`: pass
- `bunx @biomejs/biome check` on changed files: pass
- `git diff --check`: pass

### DIFF
--- a/src/multiplexer/index.ts
+++ b/src/multiplexer/index.ts
@@ -14,6 +14,8 @@ export {
   MultiplexerSessionManager,
   TmuxSessionManager,
 } from './session-manager';
+export type { SessionReadinessOptions } from './shared';
+export { waitForSessionReady } from './shared';
 export { TmuxMultiplexer } from './tmux';
 export type { Multiplexer, PaneResult } from './types';
 export { isServerRunning } from './types';
--- a/src/multiplexer/session-manager.test.ts
+++ b/src/multiplexer/session-manager.test.ts
@@ -320,6 +320,99 @@ describe('MultiplexerSessionManager', () => {
       );
     });
 
+    test('recovers a session whose initial readiness timed out on a later busy event', async () => {
+      const ctx = createMockContext();
+      const manager = new MultiplexerSessionManager(
+        ctx,
+        defaultMultiplexerConfig,
+      );
+
+      mockWaitForSessionReady.mockResolvedValueOnce(false);
+
+      await manager.onSessionCreated({
+        type: 'session.created',
+        properties: {
+          info: {
+            id: 'child-recover-timeout',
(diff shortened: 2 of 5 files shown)


alvinunreal:
thanks @adevwithpurpose

PR #1011, opened 2026-08-13
fix(foreground-fallback): protect managed child sessions and skip replay after streamed output

adevwithpurpose:

Relates to #595 and #560 (orphaned/resumed background tasks). Two failure modes remain when a managed child session errors:

1. **Detachment:** the foreground-fallback manager aborted and re-prompted *every* session on a retryable error, including background-task child sessions whose completion delivery is owned by their parent's task awaiter. Re-prompting a managed child outside its task lifecycle detaches it from that awaiter.
2. **Duplicate output / re-executed tools:** when a session streamed a partial answer or issued tool calls and then hit a mid-stream retryable error, the fallback replayed the user turn, duplicating the already-emitted assistant output or re-executing the tools with duplicate side effects.

This change narrows the fallback's blast radius. It does **not** claim to fully fix orphan-result handling — that remains owned by the background-task lifecycle (#595/#560):

- `ForegroundFallbackManager` now accepts a `shouldHandleSession` predicate. Sessions owned by the background-task lifecycle (registered on `session.created` with a `parentID`, present on the background job board, or seeded from `session.list()` at init) are skipped by both the abort path and the replay path, so fallback no longer detaches managed child sessions. After a plugin reload, pre-existing child sessions do not re-emit `session.created`, so ownership is also seeded from the session list snapshot at init (`collectManagedChildSessionIDs`) — a restart can no longer leave an in-flight child unprotected.
- A replay guard (`hasAssistantActivity`) blocks re-queueing the user turn when the failed attempt already produced activity after the last user message: meaningful text (v1 and v2 shapes) **or tool-call parts**. Reasoning/retry parts do not count as activity. Fallback still applies when the attempt failed before producing any activity (first-token rate limits, pre-stream errors).
- The fallback model and retry budget are only consumed **when the replay actually happens**. A skipped replay no longer records the next model as tried or resets the retry count, so a later retryable failure can still use the fallback model (regression-covered by a sequential-event test).
- Stream-settlement hardening: the error event can race the persistence of the streamed output, so when the failed attempt's message is still open (`time.created` without `time.completed`), the guard waits once (150 ms) and re-checks `session.messages()` before deciding to replay. No fixed delay is added when nothing is in flight.

## Verification

- `bun test src/hooks/foreground-fallback/index.test.ts`: 109 pass, 0 fail (includes managed-child detachment guards, replay-guard text/tool/settlement tests, sequential-event state-consumption regression, and predicate wiring/order tests)
- `bun test -t "restart recovery|tracks background child sessions" src/hooks/task-session-manager/index.test.ts`: 3 pass, 0 fail
- `bun run typecheck`: pass
- `bun run check:ci`: only pre-existing failures (see below) — changed files are Biome-clean
- `git diff --check`: pass

Known pre-existing failures (present at base, not introduced here; both are Windows-only path-separator issues and pass on Linux CI):
- `task-session-manager` "reads before and after launch attach with unique-line counts and caps" hard-codes a forward-slash path (`src/large.ts` vs `src\large.ts`).
- `cache-safety` "every hook module defining a message transform is covered here" compares `Bun.Glob` results against forward-slash expectations and sees backslash paths on Windows.

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -3,9 +3,10 @@ import { isInternalInitiatorPart } from '../../utils';
 import { SessionLifecycle } from '../session-lifecycle';
 import {
   ForegroundFallbackManager,
-  hasMeaningfulAssistantOutput,
+  hasAssistantActivity,
   isFailoverError,
   isRetryableError,
+  isUnsettledAssistantMessage,
 } from './index';
 
 // ACCEPTANCE GAP: config() hook behaviour is not covered by CI — verify live.
@@ -332,13 +333,13 @@ describe('isFailoverError', () => {
 });
 
 // ---------------------------------------------------------------------------
-// hasMeaningfulAssistantOutput
+// hasAssistantActivity
 // ---------------------------------------------------------------------------
 
-describe('hasMeaningfulAssistantOutput', () => {
+describe('hasAssistantActivity', () => {
   test('true for v2 assistant message with text', () => {
     expect(
-      hasMeaningfulAssistantOutput({
+      hasAssistantActivity({
         type: 'assistant',
         text: 'Here is my partial answer',
       }),
@@ -347,7 +348,7 @@ describe('hasMeaningfulAssistantOutput', () => {
 
   test('true for v1 assistant message with text part', () => {
     expect(
-      hasMeaningfulAssistantOutput({
+      hasAssistantActivity({
         info: { role: 'assistant' },
         parts: [{ type: 'text', text: 'partial' }],
       }),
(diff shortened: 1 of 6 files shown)


mhenke:
> *This was generated by AI during triage.#606 has been closed as superseded by this PR. This PR implements the ownership guard to prevent foreground fallback from aborting/re-prompting managed child sessions (the foreground fallback replay guard) and includes the replay-skip-after-streamed-output mitigation.

mhenke:
@adevwithpurpose can you please resolve the conflicts and we can merge this in   cc @alvinunreal

adevwithpurpose:
Done — rebased both commits onto master (f47b1c2) and resolved the conflicts:

- src/index.ts: kept the new revivedRunTracker wiring alongside the managed child-session set
- task-session-manager: combined the child-session cleanup with the newer isFallbackInProgress / hard-timeout guards in session.deleted
- foreground-fallback: ownership gates and replay guard merged cleanly with the current ForegroundFallbackManager

Verified after rebase: tsc clean, 113 foreground-fallback + 144 task-session-manager tests passing, biome check clean, CI green on this branch. Ready to merge from our side.

mhenke:
Closing this one, with thanks. The detachment half is covered by a different approach that landed upstream: #595 and #560 are closed via the terminal-evidence work in #1227. Foreground waiters now get promoted to background before an abort (457776e8), and terminal bookkeeping is gated on runtime evidence (46022fe3), so managed children are no longer aborted blind. Your other finding still stands: master has no guard against replaying a prompt after the session already streamed partial output, so a mid-stream error can still duplicate tool execution. A small PR limited to that replay guard would be welcome. The fallback code is being reworked in #1112 right now, so filing fresh against current master afterward is the easiest path.

PR #1018, opened 2026-08-15
fix(tasks): prevent premature council task shutdown

zrothien-cyber:
Prevent background council and councillor jobs from being stopped before their terminal results are captured.

Runtime status snapshots can temporarily omit a child session while it is still running or while its terminal result is being reconciled. The previous behavior treated a missing entry as idle and stopped the job immediately. This change marks missing entries uncertain, waits for four consecutive idle observations before stopping, and extends the idle reconciliation window to 15 seconds.

Increasing only the polling delay was insufficient because a missing session was still interpreted as idle. The new behavior preserves the existing cleanup path after repeated confirmed idle states while giving late terminal results time to arrive.

### DIFF
--- a/src/hooks/task-session-manager/index.ts
+++ b/src/hooks/task-session-manager/index.ts
@@ -43,7 +43,7 @@ export { BACKGROUND_JOB_BOARD_METADATA_KEY } from './board-injection';
  * ponytail: fixed timeout — event-driven confirmation would fully close the race but adds
  * significant complexity for a case that rarely exceeds this window in practice.
  */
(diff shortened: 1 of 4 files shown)


zrothien-cyber:
Closing this — upstream v2.2.14 already implements an equivalent (and cleaner) fix in its rewritten runtime-status reconciler: missing/error runtime status is marked `uncertain` and never treated as idle, so a temporarily-omitted child session is not stopped prematurely. That covers the root cause this PR targets. Closing as superseded; thanks for the review.

PR #1029, opened 2026-08-16
fix(config): normalize disabled_* keys before schema validation (#1027)

### LINKED ISSUE #1027: [Bug]: Legacy string in any disabled_* config key silently drops the entire OMO config

What happened: When any `disabled_*` config key (`disabled_agents`, `disabled_tools`, `disabled_mcps`, `disabled_skills`) is written in the legacy string form instead of an array, the entire oh-my-opencode-slim config file is rejected and silently discarded. The plugin falls back to an empty config, so all user settings (agents, presets, models, skills) are lost and defaults are used instead.

For example, this config:

```jsonc
{
  "disabled_agents": "explorer",
  "agents": {
    "fixer": {
      "model": "openai/gpt-5.6-luna"
    }
  }
}
```

`disabled_agents` is a string while the schema requires `string[]`. `PluginConfigSchema.safeParse()` rejects the whole object because the `disabled_*` fields are defined as `z.array(z.string()).optional()` with no `.catch()` fallback. `loadConfigFromPath` returns `null`, and `loadPluginConfig` falls back to `{}` — so the valid `agents` section is lost too. The only signal is a `console.warn` about "Invalid config", with no indication that a single legacy field caused it.

The loader even contains a normalization loop intended to coerce these fields to arrays (`src/config/loader.ts`), but its own comment states it is unreachable: safeParse rejects the file before the loop ever runs.

Expected: only the invalid field should be ignored (or normalized to an array), and the rest of the config should load normally.

### Steps to reproduce

1. Create `oh-my-opencode-slim.jsonc` with a legacy string for any `disabled_*` key plus other valid settings (e.g. an `agents` section).
2. Start OpenCode with the plugin.
3. Observe that the agent models fall back to defaults — the entire config was rejected.

Verified with the real published `PluginConfigSchema` (2.2.14): `safeParse` fails with issues on `disabled_agents` and the valid `agents` object is collateral damage; array form passes.

### oh-my-opencode.json

```jsonc
{
  "disabled_agents": "explorer",
  "agents": {
    "fixer": {
      "model": "openai/gpt-5.6-luna"
    }
  }
}
```

### OpenCode version

1.18.18

2.2.14

MyGO-Mujica:
Fixes #1027

When any `disabled_*` config key (`disabled_agents`, `disabled_tools`, `disabled_mcps`, `disabled_skills`) is written as a legacy string instead of an array, `PluginConfigSchema.safeParse()` rejects the entire config object and `loadPluginConfig` silently falls back to `{}` — dropping every other user setting (agents, presets, models) with only a generic "Invalid config" warning.

## What does this change?

- Normalize `disabled_*` keys inside `loadConfigFromPath` before schema validation:
  - String value → normalized to a single-element array (e.g. `"explorer"` → `["explorer"]`), preserving the user's disable intent, with a warning.
  - Any other non-array value (number, boolean, object) → dropped with a warning.
  - Array / undefined → untouched.
- Warnings go through `options.onWarning` (`invalid-schema`, already consumed by doctor/TUI) and `console.warn`, and respect `options.silent`.
- Removed the now-superseded unreachable normalization loop at the end of `loadPluginConfig` (its own comment stated safeParse rejected the file before it could run).
- Regression tests: string normalization, array passthrough, garbage-type drop, and untouched-config control cases.

## Why this approach?

Normalizing before `safeParse` keeps a single invalid field from discarding the user's whole config, while a string still preserves the disable intent (writing `"explorer"` clearly means "disable explorer"). This matches the loader's original normalization intent that was never reachable.

## Verification

- [x] `bun test src/config/loader.test.ts` — 83 pass
- [x] `bun test src/config/` — 179 pass (10 files, no regression)
- [x] `bun x biome check` on both changed files — clean
- [ ] `bun run check:ci` — not run (Windows CRLF workspace makes full-repo format check noisy)
- Windows note: changed files converted to LF; verified locally on Windows with Bun 1.3.14.

### DIFF
--- a/src/cli/doctor.test.ts
+++ b/src/cli/doctor.test.ts
@@ -105,6 +105,63 @@ describe('runDoctorCheck', () => {
     expect(result.configs[1].path).toContain('.jsonc');
   });
 
+  test('string disabled_agents normalizes instead of failing schema validation', () => {
+    const projectDir = path.join(tempDir, 'project');
+    const configDir = path.join(projectDir, '.opencode');
+    fs.mkdirSync(configDir, { recursive: true });
+    fs.writeFileSync(
+      path.join(configDir, 'oh-my-opencode-slim.json'),
+      JSON.stringify({
+        disabled_agents: 'explorer',
+        agents: { oracle: { model: 'test/model' } },
+      }),
+    );
+
+    const warnSpy = spyOn(console, 'warn').mockImplementation(() => {});
+    try {
+      const result = runDoctorCheck(projectDir);
+
+      // The string is normalized (matching the loader), so the config is
+      // valid rather than a false invalid-schema diagnosis
+      expect(result.ok).toBe(true);
+      expect(result.configs[1].ok).toBe(true);
+      expect(result.configs[1].error).toBeUndefined();
+      expect(result.configs[1].config?.disabled_agents).toEqual(['explorer']);
+      expect(result.configs[1].config?.agents?.oracle?.model).toBe(
+        'test/model',
+      );
+
+      // The normalization is reported to the user
+      const calls = warnSpy.mock.calls as string[][];
+      const message = calls.find((call) =>
+        (call[0] as string).includes('disabled_agents'),
+      )?.[0];
+      expect(message).toContain('should be an array; normalized');
+    } finally {
+      warnSpy.mockRestore();
+    }
+  });
+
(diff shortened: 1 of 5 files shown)


MyGO-Mujica:
@greptile-apps please recheck

alvinunreal:
thanks @MyGO-Mujica

Issue #1031, opened 2026-08-16
[Bug]: Orchestrator cannot see live task timing and falsely cancels active background tasks


ProphetOfDoom-PoD:
### What happened, and what did you expect?

What happened:
With oh-my-opencode-slim 2.2.14, an orchestrator canceled an active background writer as “45+ minutes without liveness.” The actual task lifetime was 10m 29s. Slim’s task-session-manager log recorded 176 consecutive busy observations and lastLiveBusyAt only 10 seconds before the parent-initiated cancellation.

The parent-facing Background Job Board and orchestrator-wake reminder did not expose elapsed time, last_live_busy_at, or idle duration. The wake text was static, so the parent inferred a stall incorrectly and issued cancel_task. The child session was then deleted after abort, removing its normal inspectable history.

Expected:
The parent should receive actual per-task elapsed and liveness metadata. A task with recent busy activity should not be presented as plausibly stalled without an explicit warning. Canceled child-session history should remain inspectable or be archived.

This was not an OpenCode automatic timeout: cancellation was explicitly parent-initiated.

### Steps to reproduce

1. Use oh-my-opencode-slim 2.2.14 with default orchestrator-wake settings and no wallClockTimeoutMs override.
2. Start a background task that spends 10–12 minutes reading/searching files or invoking tools without editing files.
3. Let the orchestrator-wake reminder fire.
4. Observe that parent-visible task information contains no elapsed time, last activity timestamp, or idle duration.
5. Cancel the task while the task-session-manager still records recent busy activity.
6. Observe that cancellation deletes the child session history.

Observed incident:
- task created: 2026-08-15T09:42:17Z
- last busy observation: 2026-08-15T09:52:36Z
- parent cancellation: 2026-08-15T09:52:46Z
- claimed stall duration: “45+ minutes”

### oh-my-opencode.json

```jsonc
{
  "preset": "omniroute-hybrid"
  // No orchestratorWake or wallClockTimeoutMs override.
}
```

### OpenCode version

1.18.13

### oh-my-opencode-slim version

2.2.14

### Operating system

Windows

### Logs, screenshots, or extra context

```shell
Sanitized task-session-manager evidence:

2026-08-15T09:52:36.474Z:
busy/status busy observed
sessionID=ses_ffb330710...
updatedLastLiveBusyAt=1786787556474

2026-08-15T09:52:46.480Z:
cancel-task request received
requested=ses_ffb330710...
state=running
[...]

PR #1032, opened 2026-08-16
fix(tasks): Guard recent live-busy cancellations

zrothien-cyber:
Prevent `cancel_task` from aborting a tracked job that has reported a live busy status in the prior two reconciliation cycles. Intentional cancellation now requires `force: true`; stale, timed-out, and status-uncertain jobs retain their existing paths.

This addresses the parent-initiated cancellation in #1031. Timing display and child-history archival remain out of scope.

Repository-wide `bun run check:ci` still reports pre-existing violations in `src/utils/session.ts` and `src/agents/orchestrator.ts`; targeted Biome, full tests, typecheck, and build pass.

Refs #1031

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -51,6 +51,8 @@ Fast, structural code search and refactoring - more powerful than plain text gre
 the current orchestrator session, and it does not roll back partial edits. After
 cancelling a write-capable task, inspect and reconcile file changes before
 launching replacement work.
+Recent live-busy tasks are protected for 10 seconds; use `force: true` only for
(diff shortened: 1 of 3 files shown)


mhenke:
Closing for now. Master rewrote cancel-task.ts and its tests since this was opened (923 insertions, 622 deletions on our side of the diff), so the branch conflicts deeply in both files and the 39-line guard can't be re-applied cleanly. The underlying problem in #1031 is still open and still real. If you're up for it, a fresh PR against the current cancel-task implementation would be very welcome.

PR #1035, opened 2026-08-16
fix(config): strip UTF-8 BOM before parsing config files (#1028)

### LINKED ISSUE #1028: Config with UTF-8 BOM silently ignored — presets/subagent models never load
## Describe the bug

If `oh-my-opencode-slim.json` starts with a UTF-8 BOM (`EF BB BF`, i.e. `U+FEFF`), the plugin's config loader fails and the **entire config is silently discarded**.

`loadConfigFromPath` in `dist/index.js` reads the file with `fs.readFileSync(configPath, "utf-8")` — Node keeps the BOM as a `\uFEFF` character in the string — then passes the content to `JSON.parse` after comment/env interpolation. `JSON.parse` throws:

```
[oh-my-opencode-slim] Invalid JSON in .../oh-my-opencode-slim.json: JSON Parse error: Unrecognized token ''
```

`loadConfigFromPath` returns `null`, so no preset, no per-agent models, no skills/MCPs configuration is applied. The plugin falls back to built-in defaults, and every subagent inherits the session/orchestrator model instead of its configured model (e.g. a lightweight explorer/librarian runs the heavy orchestrator model).

The only user-visible output is a `console.warn` at plugin load time; there is no error dialog or other indication that the config was rejected. Users only discover the problem indirectly (subagents running the wrong model).

## To reproduce

1. Save `oh-my-opencode-slim.json` as **UTF-8 with BOM** (common on Windows: Notepad "UTF-8 with BOM" save, PowerShell 5.1 `Set-Content` default encoding, VS Code with `files.encoding: "utf8bom"`, some git/editor round-trips).
2. Start opencode with the plugin loaded.
3. Spawn any subagent (e.g. `fixer`, `explorer`).
4. Observed: the subagent runs with the session/orchestrator model instead of the model configured for it in the preset.

## Expected behavior

- A BOM is invisible to JSON consumers per RFC 8259 (the file is valid JSON if the BOM is ignored), so the plugin should tolerate it: strip a leading `\uFEFF` before parsing.
- Even if parsing fails, the failure should be loud and specific — ideally a distinct warning (e.g. a `bom-detected` or `invalid-json` warning with a clear message naming the cause), not a silent fallback that makes every subagent run the wrong model.

## Suggested fix

In `loadConfigFromPath`, strip a leading BOM before comment stripping/interpolation/parse:

```js
const content = fs.readFileSync(configPath, "utf-8").replace(/^\uFEFF/, "");
```

and/or emit a dedicated `bom-detected` warning kind so users see exactly what happened.

## Impact

Silent misconfiguration affecting all agent model assignments (and skills/MCP config). Only discoverable by inspecting `opencode debug agent <name>` output or `opencode.db` session rows.

MyGO-Mujica:
Fixes #1028

Config files saved as UTF-8 with BOM (`EF BB BF`) are silently discarded: `fs.readFileSync(..., "utf-8")` keeps the BOM as a leading `\uFEFF` character, `stripJsonComments` leaves it in place, and `JSON.parse` throws `Unrecognized token`. `loadConfigFromPath` then returns `null` and the plugin falls back to an empty config — no preset, no per-agent models, no skills/MCPs configuration is applied, and subagents run the session/orchestrator model instead of their configured model. The only signal is a single `console.warn` at load time.

## What does this change?

- Strip a leading UTF-8 BOM (`/^\uFEFF/`) from file content right after `readFileSync`, before comment stripping / parsing, in all three config-reading paths:
  - `src/config/loader.ts` (`loadConfigFromPath`) — the main runtime path.
  - `src/cli/doctor.ts` (`checkConfigFile`) — so doctor no longer misdiagnoses a BOM-prefixed config as `invalid-json` (keeps doctor consistent with the runtime loader).
  - `src/cli/config-io.ts` (`parseConfigFile`) — shared by auto-update-checker and other consumers.
- Regression tests for all three paths: a BOM-prefixed config loads identically to the same config without BOM.

## Why this approach?

RFC 8259 permits a BOM and treats it as invisible to JSON consumers, so tolerating it is the correct behavior — no warning needed. Stripping at each read site keeps the change minimal and mirrors the existing normalization-before-validation pattern used for `disabled_*` keys (#1029).

## Verification

- [x] `bun test src/config/loader.test.ts` — 84 pass
- [x] `bun test src/cli/doctor.test.ts` — 30 pass
- [x] `bun test src/cli/config-io.test.ts` — 37 pass
- [x] `bun test src/config/` — 180 pass (10 files, no regression)
- [x] `bun x biome check` on all 6 changed files — clean

### DIFF
--- a/src/cli/config-io.test.ts
+++ b/src/cli/config-io.test.ts
@@ -76,6 +76,14 @@ describe('config-io', () => {
     expect(result.error).toBeUndefined();
   });
 
+  test('parseConfigFile strips a UTF-8 BOM before parsing', () => {
+    const path = join(tmpDir, 'bom.json');
+    writeFileSync(path, `\uFEFF${'{"a": 1}'}`);
+    const result = parseConfigFile(path);
+    expect(result.config).toEqual({ a: 1 } as any);
+    expect(result.error).toBeUndefined();
+  });
+
   test('parseConfigFile returns null for non-existent file', () => {
     const result = parseConfigFile(join(tmpDir, 'nonexistent.json'));
     expect(result.config).toBeNull();
--- a/src/cli/config-io.ts
+++ b/src/cli/config-io.ts
@@ -404,7 +404,8 @@ export function parseConfigFile(path: string): {
     if (!existsSync(path)) return { config: null };
     const stat = statSync(path);
     if (stat.size === 0) return { config: null };
-    const content = readFileSync(path, 'utf-8');
+    // Strip a UTF-8 BOM (RFC 8259 permits one) so JSON.parse does not choke.
+    const content = readFileSync(path, 'utf-8').replace(/^\uFEFF/, '');
     if (content.trim().length === 0) return { config: null };
     return { config: JSON.parse(stripJsonComments(content)) as OpenCodeConfig };
   } catch (err) {
--- a/src/cli/doctor.test.ts
+++ b/src/cli/doctor.test.ts
@@ -162,6 +162,27 @@ describe('runDoctorCheck', () => {
     expect(result.configs[1].config?.disabled_tools).toEqual(['webfetch']);
   });
 
+  test('config with UTF-8 BOM is accepted as valid', () => {
+    const projectDir = path.join(tempDir, 'project');
+    const configDir = path.join(projectDir, '.opencode');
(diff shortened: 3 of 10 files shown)


MyGO-Mujica:
@greptile-apps please recheck

PR #1042, opened 2026-08-17
fix(foreground-fallback): recognize ProviderModelNotFoundError as failover (#1034)

### LINKED ISSUE #1034: [Bug]: ProviderModelNotFoundError does not advance configured fallback chain

When the first model in an agent's configured model chain is absent from OpenCode's provider registry, OpenCode raises:

```text
ProviderModelNotFoundError: Model not found: custom/missing-model.
```

The agent session immediately becomes idle/terminal and oh-my-opencode-slim does not advance to the next configured model. There are no `[foreground-fallback]` entries for that session.

Expected: because the agent has an explicit ordered model chain and fallback is enabled, skip the unresolvable primary and try the next model.

Actual: the task fails on the primary; launching the task again selects the same missing primary and fails identically.

This is distinct from #943: that issue covered a provider response saying `No available channel`. Here OpenCode rejects the model locally, before any provider request. In the same OpenCode run, a separate `No available channel` error did successfully produce a `[foreground-fallback] switched to fallback model` entry, confirming that the chain and fallback manager were active.

Current `main` (2.2.14) still does not classify the exact `Model not found` error. `PROVIDER_OUTAGE_PATTERNS` includes `model ... not available`, `unsupported model`, and `unknown model`, but not OpenCode's `ProviderModelNotFoundError` / `Model not found` wording.

### Steps to reproduce

1. Configure a custom OpenCode provider with only `valid-model` registered.
2. Configure an agent with `model: ["custom/missing-model", "custom/valid-model"]`.
3. Set `fallback.enabled` to `true`.
4. Start OpenCode and dispatch a task to that agent.
5. Observe `ProviderModelNotFoundError` and immediate task termination; `custom/valid-model` is never attempted.

### oh-my-opencode.json

```jsonc
{
  "fallback": { "enabled": true, "maxRetries": 1 },
  "agents": {
    "compare": {
      "model": [
        "custom/missing-model",
        "custom/valid-model"
      ]
    }
  }
}
```

Relevant OpenCode provider shape:

```jsonc
{
  "provider": {
    "custom": {
      "models": {
        "valid-model": { "name": "valid-model" }
      }
    }
  }
}
```

### OpenCode version

1.18.18

Observed on 2.2.13; verified the classification gap is still present on current `main` / 2.2.14.

### Operating system

macOS 27.0 (arm64)

```text
created ... agent=compare model=undefined
share subscriber failed type=message.updated cause="Cause([Fail(ProviderModelNotFoundError: Model not found: custom/missing-model.)])"
loop ... step=0
```

A narrow fix could recognize the structured OpenCode `ProviderModelNotFoundError` (or its exact `Model not found: provider/model` message) and advance only when that session has another configured chain entry. This avoids broadly treating arbitrary provider `model_not_found` responses as transient outages.

MyGO-Mujica:
Fixes #1034

When the first model in an agent's configured model chain is absent from OpenCode's provider registry, OpenCode raises `ProviderModelNotFoundError: Model not found: <model>`. `isFailoverError()` does not classify this wording — `PROVIDER_OUTAGE_PATTERNS` covers `model not available`, `unsupported model`, and `unknown model`, but not `Model not found` — so the fallback chain never advances and the task fails on the missing primary model. Relaunching the task selects the same missing primary and fails identically.

## What does this change?

- `PROVIDER_OUTAGE_PATTERNS` gains `/\bmodel not found\b/i`, so `ProviderModelNotFoundError` advances the fallback chain.
- Tests: classification, regression guard, negative cases, and a gate-level `session.error` fallback-switch test.

## Why this approach?

The model may exist on a later entry in the configured chain, so a missing primary should be treated as a provider outage and advance the chain rather than fail the task. The pattern is deliberately narrow (`model not found`) to avoid matching unrelated wording such as configuration errors.

## Verification

- `bun test src/hooks/foreground-fallback/index.test.ts` — 93 pass (89 pre-existing + 4 new)
- `bun run typecheck` — clean
- `bun x biome check` on both changed files — clean

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -298,6 +298,36 @@ describe('isFailoverError', () => {
       }),
     ).toBe(false);
   });
+
+  test('returns true for OpenCode ProviderModelNotFoundError "Model not found" errors', () => {
+    // Issue #1034: OpenCode's ProviderModelNotFoundError ("Model not found:
+    // <model>") was not classified as a failover error, so a missing primary
+    // model failed the task outright instead of advancing the fallback chain.
+    // The reporter's error string always carries the message; the bare
+    // camelCase class name "ProviderModelNotFoundError" (no spaces) does not
+    // match /\bmodel not found\b/i and is intentionally not covered here.
+    expect(
+      isFailoverError(
+        'ProviderModelNotFoundError: Model not found: custom/missing-model.',
+      ),
+    ).toBe(true);
+    expect(
+      isFailoverError({ message: 'Model not found: custom/missing-model' }),
+    ).toBe(true);
+  });
+
+  test('returns true for existing model-outage patterns (regression guard)', () => {
+    expect(isFailoverError('model not available')).toBe(true);
+    expect(isFailoverError('unsupported model')).toBe(true);
+    expect(isFailoverError('unknown model')).toBe(true);
+  });
+
+  test('returns false for normal errors and model mentions without outage wording', () => {
+    expect(isFailoverError('Cannot connect to database')).toBe(false);
+    expect(isFailoverError({ message: 'invalid model configuration' })).toBe(
+      false,
+    );
+  });
 });
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


alvinunreal:
thanks @MyGO-Mujica

PR #1047, opened 2026-08-18
fix(tmux): route subagent panes to attached parent session

### LINKED ISSUE #1016: [Bug]: tmux subagent panes target the server's initial pane for attached sessions

When multiple OpenCode TUI clients attach to one server from separate tmux sessions, subagent panes always open in the tmux session containing the original OpenCode server.

Expected behavior: a subagent pane should open beside the attached TUI session whose root session started that subagent.

Actual behavior: subagent panes from every attached root session open beside the server's initial TUI.

This prevents tmux pane isolation between concurrently attached root sessions.

## Steps to reproduce

1. Start OpenCode with oh-my-opencode-slim inside tmux session A:

   ```sh
   opencode --hostname 127.0.0.1 --port 4096
   ```

2. Start another tmux session B and attach a different root session:

   ```sh
   opencode attach http://127.0.0.1:4096 \
     --session <root-session-id> \
     --dir <project-directory>
   ```

3. Start a background subagent from the root session displayed in tmux session B.
4. Observe that the subagent pane is created in tmux session A rather than B.

Clearing `~/.cache/opencode` and restarting does not change the behavior.

## oh-my-opencode.json

```jsonc
{
  "multiplexer": {
    "type": "auto",
    "layout": "main-horizontal",
    "main_pane_size": 70
  }
}
```

## OpenCode version

1.18.18

## oh-my-opencode-slim version

2.2.13

## Operating system

macOS 26.5, tmux 3.6a

## Logs, screenshots, or extra context

The apparent cause is static tmux target capture in `src/multiplexer/tmux/index.ts`.

`TmuxMultiplexer` captures the server process's pane:

```ts
targetPane = process.env.TMUX_PANE
```

`spawnPane()` later adds that fixed pane through `targetArgs()`:

```ts
return this.targetPane ? ["-t", this.targetPane] : []
```

The plugin handling subagent lifecycle runs in the shared OpenCode server process. An `opencode attach` client runs inside its own tmux session, but its `TMUX_PANE` is not communicated to that server process. Consequently, every subagent split targets the server's original pane.

A possible solution would be per-root-session pane registration:

- Associate each attached root session ID with its tmux pane or tmux session.
- Pass the parent root session identity into multiplexer pane creation.
- Resolve the split target from that mapping.
- Retain the server's initial pane as the fallback for non-attached/local TUI usage.

ermanhavuc:
## Summary

- register each local TUI's active OpenCode session and `TMUX_PANE` with a 10-second heartbeat and 30-second expiry
- forward child `parentID` into tmux pane creation, targeting splits and layout updates beside the attached parent TUI
- fall back to the server process's startup pane when registration is missing, stale, or rejected by tmux
- document attached-session targeting and add regression coverage for registration, routing, layout isolation, expiry, ownership, and fallback

Closes #1016

## Verification

- `bun test`
- `bun run typecheck`
- `bun run build`
- `bun run lint`
- `bun test src/multiplexer/tmux/index.test.ts src/multiplexer/tmux-pane-registry.test.ts src/tui.test.ts src/multiplexer/session-manager.test.ts`

`bun run check:ci` still reports three pre-existing formatting violations in `src/hooks/task-session-manager/idle-reconciliation.test.ts`, `src/hooks/task-session-manager/runtime-status-reconciliation.test.ts`, and `src/hooks/task-session-manager/runtime-status-reconciliation.ts`; changed files pass Biome checks.

### DIFF
--- a/docs/multiplexer-integration.md
+++ b/docs/multiplexer-integration.md
@@ -326,6 +326,19 @@ OpenCode pane (resolved from the parent pane's `ZELLIJ_PANE_ID` via
 If the parent pane cannot be resolved, the tab target is omitted and Zellij
 places the pane in whatever tab it has focused — no tab id is guessed.
 
+### Tmux attached-session targeting
+
+When multiple local TUI clients attach to one OpenCode server from different
+tmux sessions, each TUI records its active OpenCode session and `TMUX_PANE`.
+Child panes and layout updates target the tmux pane registered by their parent
+session, so each attached root session keeps its subagents beside itself.
+
+Registrations are session-scoped, refreshed while the TUI is active, and expire
+after 30 seconds without a heartbeat. If no fresh registration exists,
+or tmux rejects a registered target, pane creation falls back to the server
+process's original `TMUX_PANE`. This preserves direct/local TUI behavior and
+avoids losing subagent visibility after an attached pane closes unexpectedly.
+
 ### Zellij details
 
 The Zellij adapter requires **Zellij 0.44.1 or newer**. Older releases are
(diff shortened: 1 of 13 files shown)


alvinunreal:
thanks @ermanhavuc

PR #1048, opened 2026-08-18
fix(smartfetch): set parentID on secondary sessions to suppress TUI notifications (#1046)

### LINKED ISSUE #1046: smartfetch secondary session triggers TUI "Session done" notification (noise, not actionable)
## Summary

Every `webfetch` call that uses the secondary-model summarization pops a macOS notification **"Session done · smartfetch-secondary"** from the opencode TUI whenever the terminal is blurred. The notification is technically accurate but not actionable — it fires for an ephemeral helper session the user never sees.

## Environment

- opencode 1.18.18 (TUI)
- oh-my-opencode-slim 2.2.15
- macOS

## Root cause

1. The smartfetch tool creates a plain SDK session for secondary-model summarization:

   ```ts
   const sessionResponse = await client.session.create({
     query: { directory },
     body: { title: "smartfetch-secondary" },
     throwOnError: true
   });
   ```

   No `parentID` is set.

2. The opencode TUI ships a built-in `internal:notifications` plugin that raises a "Session done" notification whenever any session transitions busy → idle. It suppresses notifications only for sessions that have a `parentID` (subagent sessions):

   ```js
   // opencode TUI (packages/tui):
   function EJ(B, Z, X, Q) {
     let J = Z ? B.state.session.get(Z) : void 0;
     let Y = J?.parentID !== void 0;
     B.attention.notify({
       title: J?.title, message: X,
       notification: Y ? false : { when: "blurred" },
       sound: { name: Q, when: "always" }
     });
   }
   ```

   Because `smartfetch-secondary` has no `parentID`, the TUI treats it like a first-class session and notifies — even though it lives for a single summarization call and is deleted immediately after.

## Suggested fix

The tool's `execute(args, ctx)` context already exposes the calling session (`ToolContext.sessionID`). Create the secondary session with that as its `parentID`:

```ts
body: { title: "smartfetch-secondary", parentID: ctx.sessionID }
```

The TUI then classifies it as a subagent session and suppresses the "Session done" notification (same path as `task`-spawned subagents), and opencode's session tracking treats it as child work of the real session.

## Repro

1. Run opencode in the TUI.
2. Switch focus to another app (so the TUI is blurred).
3. Trigger a `webfetch` with a `prompt` argument (e.g. `webfetch(url="https://example.com", prompt="summarize")`).
4. Observe a "Session done · smartfetch-secondary" notification pop shortly after.

Expected: no notification for the helper session; the main session's own "Session done" (and question/permission notifications) remain unchanged.

MyGO-Mujica:
Fixes #1046

Every `webfetch` call that uses secondary-model summarization pops a macOS notification "Session done — smartfetch-secondary" from the opencode TUI whenever the terminal is blurred. The secondary session is an ephemeral helper session the user never sees, but because it is created without a `parentID`, the TUI's `internal:notifications` plugin treats it like a first-class session and raises a "Session done" notification (it only suppresses notifications for sessions that have a `parentID`).

## What does this change?

- `runSecondaryModel` and `runSecondaryModelWithFallback` accept an optional `parentSessionID`.
- The smartfetch tool passes `ctx.sessionID` (the calling session) as `parentSessionID`.
- The secondary session is created with `parentID: parentSessionID`, so the TUI classifies it as a subagent session and suppresses the "Session done" notification — the same path used by `task`-spawned subagents.
- When `parentSessionID` is absent, the create body is byte-identical to before (safe degradation).
- Tests: `parentID` present in the create body when passed, absent when not; execute-level test asserting `ctx.sessionID` flows into the create body.

## Why this approach?

The secondary session is child work of the real session, so tagging it with the caller's `parentID` is semantically correct and matches how subagent sessions are already tracked. The TUI's notification suppression is keyed on `parentID`, so this is the minimal change that stops the noise without touching the TUI.

## Verification

- [x] `bun test src/tools/smartfetch/` — 50 pass (3 new, no regression)
- [x] `bun x biome check` on all 4 changed files — clean

### DIFF
--- a/src/tools/smartfetch/secondary-model.test.ts
+++ b/src/tools/smartfetch/secondary-model.test.ts
@@ -362,4 +362,44 @@ describe('smartfetch/secondary-model', () => {
       _testConfig.secondaryModelTimeoutMs = originalTimeout;
     }
   });
+
+  test('passes parentID to session.create when parentSessionID is provided', async () => {
+    mockV2Client = createV2ClientMock([{ text: 'Answer' }]);
+
+    const result = await runSecondaryModelWithFallback(
+      testInput,
+      [models[0]],
+      'Summarize',
+      'This is enough fetched content to clear the short-content guard.',
+      'parent-session-id',
+    );
+
+    expect(result.text).toBe('Answer');
+    expect(mockV2Session.create).toHaveBeenCalledWith(
+      expect.objectContaining({
+        body: expect.objectContaining({
+          title: 'smartfetch-secondary',
+          parentID: 'parent-session-id',
+        }),
+      }),
+    );
+  });
+
+  test('omits parentID from session.create when parentSessionID is undefined', async () => {
+    mockV2Client = createV2ClientMock([{ text: 'Answer' }]);
+
+    const result = await runSecondaryModelWithFallback(
+      testInput,
+      [models[0]],
+      'Summarize',
+      'This is enough fetched content to clear the short-content guard.',
+    );
+
+    expect(result.text).toBe('Answer');
+    expect(mockV2Session.create).toHaveBeenCalledWith(
+      expect.objectContaining({
+        body: { title: 'smartfetch-secondary' },
+      }),
+    );
+  });
 });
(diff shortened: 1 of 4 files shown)


alvinunreal:
thanks @MyGO-Mujica

PR #1049, opened 2026-08-18
fix(task-result): return status for running background tasks

### LINKED ISSUE #1039: Orchestrator prematurely calls task_result for running background tasks
### Environment
- **oh-my-opencode-slim**: `2.2.14`
- **OpenCode Version**: `0.0.0-next-16378`
- **OS**: macOS (`darwin`)

### Symptom
When the Orchestrator agent spawns background subagents (e.g., `fixer`, `designer`), red tool error blocks appear in the OpenCode TUI / UI for `task_result` calls shortly after launch, even though the subagent is running fine.

### Error Text (Sanitized)
```
Task <task-id> is still running. Wait for its terminal result instead of retrieving or duplicating it.
```

### Steps to Reproduce / Observations
1. The Orchestrator delegates a task to a background subagent (e.g., `fixer` or `designer`) using the `task` tool.
2. The Background Job Board registers the subagent task with state `running`.
3. In the same or immediate next turn, Orchestrator prematurely invokes `task_result` with the running task's ID (`<task-id>`).
4. `task_result` tool execution runs `assertRetrievableState()`, sees `tracked.state === "running"` (or `status.type === "busy"`), and throws `Error: Task <task-id> is still running...`.
5. OpenCode UI renders this thrown tool exception as a red error block.
6. Meanwhile, the subagent process continues executing in the background and eventually completes successfully, reconciling its final output into the parent session.

### Actual Behavior
- `task_result` throws an unhandled tool error when queried for a `running` background task, resulting in misleading red error UI elements and an unnecessary wasted tool turn.

### Expected Behavior
- Either the Orchestrator prompt should explicitly instruct the model NOT to call `task_result` on subagents whose state is `running`, OR `task_result` should return a clean non-error status message (e.g., structured XML indicating the task is still in progress) rather than throwing an exception.

### Impact
- **Non-blocking for background execution**: The subagent continues running in the background and successfully finishes its work.
- **User Experience / TUI misleading**: Users seeing red `task_result` error blocks in the TUI get the impression that the background subagent crashed or failed.
- **Wasted Tool Call**: Orchestrator wastes a model turn attempting an invalid retrieval.

### Root Cause Evidence
- In `createTaskResultTool` (`oh-my-opencode-slim/dist/server.js`):
  ```ts
  function assertRetrievableState(requested, tracked) {
    const terminalState = tracked.state;
    if (terminalState === "running") {
      throw new Error(`Task ${requested} is still running. Wait for its terminal result instead of retrieving or duplicating it.`);
    }
  }
  ```
- In plugin system logs (`oh-my-opencode-slim.*.log`):
  - Log records `background task launch registered` with `state: "running"`.
  - Immediately followed by `tool.execute.after task_result` returning `status: "error"`.
  - Subsequent `busy/status busy observed` logs confirm the subagent session remains alive and active until `state: "completed"` and `processed injected background completion`.

### Suggested Direction
1. Clarify Orchestrator prompt guidelines so Orchestrator understands `task_result` is strictly for retrieving already-completed tasks, avoiding premature calls on `running` tasks.
2. Alternatively (or additionally), have `task_result` handle `running` state gracefully by returning a non-throwing status summary string (e.g. `<task id="..." state="running">Background task is currently in progress...</task>`).

HeZ2z:
Fixes #1039

When the Orchestrator calls `task_result` for a background task that is still running, the tool throws an exception. OpenCode renders that exception as a red TUI error block even though the child task continues running successfully.

## What does this change?

- Return a clean status message for tracked `running` and `retry` tasks before reading child messages.
- Recommend `task_status` only for tasks registered in the Background Job Board.
- Tell untracked live sessions to retry `task_result` after the task finishes.
- Keep completed tasks returning their final text.
- Preserve error, cancelled, partial-result, and generation-integrity protections.
- Document the active-state response in the tool description.
- Add regression coverage for actionable tracked and untracked guidance, including tracked retry state.

## Why this approach?

The Job Board and runtime status endpoint already distinguish active tasks from terminal results. Returning that existing state removes the misleading UI error without weakening the checks that prevent partial output from being presented as a completed result. The change stays scoped to `task-result` and does not alter Orchestrator prompts or generated files.

## Verification

- [x] `bun test src/tools` — 130 passed
- [x] `bun run lint` — clean
- [x] `bunx biome check src/tools/task-result.ts src/tools/task-result.test.ts` — clean
- [x] `git diff --check` — clean
- `bun test` — 2133 passed; 4 unrelated interview dashboard integration tests currently fail

### DIFF
--- a/src/tools/task-result.test.ts
+++ b/src/tools/task-result.test.ts
@@ -85,7 +85,7 @@ describe('task_result', () => {
     ).resolves.toBe('final findings');
   });
 
-  test('rejects a still-running tracked task', async () => {
+  test('returns a non-error status for a still-running tracked task', async () => {
     const { board, tool, messages } = createTool();
     board.registerLaunch({
       taskID: 'ses_child1',
@@ -94,12 +94,14 @@ describe('task_result', () => {
       description: 'trace bug',
     });
 
-    await expect(
-      tool.execute({ task_id: 'exp-1' }, {
-        sessionID: 'parent-1',
-        agent: 'orchestrator',
-      } as any),
-    ).rejects.toThrow('still running');
+    const output = await tool.execute({ task_id: 'exp-1' }, {
+      sessionID: 'parent-1',
+      agent: 'orchestrator',
+    } as any);
+
+    expect(output).toContain('task_id: ses_child1');
+    expect(output).toContain('state: running');
+    expect(output).toContain('task_status');
     expect(messages).not.toHaveBeenCalled();
   });
 
@@ -122,13 +124,13 @@ describe('task_result', () => {
       data: { ses_child1: { type: 'busy' } },
     });
 
-    await expect(
-      tool.execute({ task_id: 'exp-1' }, {
-        sessionID: 'parent-1',
-        agent: 'orchestrator',
-      } as any),
-    ).rejects.toThrow('still running');
+    const output = await tool.execute({ task_id: 'exp-1' }, {
+      sessionID: 'parent-1',
+      agent: 'orchestrator',
+    } as any);
 
+    expect(output).toContain('state: running');
+    expect(output).toContain('task_status');
     expect(board.get('ses_child1')).toMatchObject({
       state: 'running',
       statusUncertain: false,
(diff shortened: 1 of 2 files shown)


alvinunreal:
thanks @HeZ2z

Issue #1002, opened 2026-08-11
Support for OpenCode v2 (v2.opencode.ai)?


InCerryGit:
Hello,

Thank you for building and maintaining `oh-my-opencode-slim`.

Does the plugin currently support OpenCode v2 (`v2.opencode.ai`)? The OpenCode migration guide notes that V1 plugins are not compatible with V2 and that the V2 plugin API is still being finalized.

If V2 is not supported yet, are there plans or a tracking issue for the migration? Any expected limitations or guidance for migrating existing presets, agent configuration, and workflows would also be appreciated.

Thank you for your time.

HowBoring:
Yes, partially. You can find more details in the document [OpenCode v2 (opencode2) Compatibility](https://github.com/alvinunreal/oh-my-opencode-slim/blob/master/docs/opencode-v2-compatibility.md).

80avin:
It currently breaks on opencode2

I doubt if the claimed v2 support is for the npm `opencode-ai@next` release.
If yes, then that is severely outdated. It was made 2 months ago ( 27 June ) and the latest v2 master has been breaking since quite long.

forstjiri:
Concrete error i got: "Invalid V2 TUI plugin module: oh-my-opencode-slim"

mhenke:
> *This was generated by AI during triage.*

Labeled enhancement. This is human-led work: the v2 plugin API is still moving, and the compatibility doc overstates what currently works. The concrete breakage reports in this thread ("Invalid V2 TUI plugin module") should go into docs/opencode-v2-compatibility.md so the doc matches reality. An agent can draft that doc update; the migration itself needs a maintainer call on scope and timing.

mhenke:
> *This was generated by AI during triage.*

Adding a concrete acceptance item so this doesn't get lost: an agent can draft the docs/opencode-v2-compatibility.md update that records the current breakage reports from this thread (for example "Invalid V2 TUI plugin module"). The doc currently overstates what works. The v2 migration itself stays a maintainer decision on scope and timing; this issue tracks the doc accuracy fix independently.

Crackde:
> Concrete error i got: "Invalid V2 TUI plugin module: oh-my-opencode-slim"

You probably have `oh-my-opencode-slim` listed in your plugins.

Try changing it to `oh-my-opencode-slim@latest` in your `opencode.json` file (for example, `.config/opencode/opencode.json`):

```json
"plugin": [
[...]

PR #1058, opened 2026-08-21
fix(tui): support opencode2 beta-17793 { id, setup } TUI plugin contract

arpankanwer:
## Problem

opencode2 (`@opencode-ai/cli@0.0.0-beta-17793`) fails to load the plugin's TUI module:

```
Plugin failed: oh-my-opencode-slim
Invalid V2 TUI plugin module: oh-my-opencode-slim
```

Tracked in #1002 (multiple users confirmed this exact error).

## Root cause

The beta's V2 TUI plugin loader imports each registered package's `./tui` subpath and validates the default export as `{ id: string, setup: function }` (mirroring `@opencode-ai/plugin@0.0.0-beta-17793` `dist/tui/plugin.d.ts`), then calls `setup(context)` with a single context argument and treats the return value as cleanup.

The shipped module only exported the v1 contract `{ id, tui(api, options, meta) }`, so validation failed. OpenCode v1 hosts validate `{ id, tui }` and ignore extra keys — both contracts must be satisfied simultaneously from one module.

## Fix

Dual-contract default export `{ id, tui, setup }`:

- **`tui`** — v1 hook, byte-for-byte unchanged behavior (sidebar slot, `/preset` command, tmux pane sync). Only mechanical change: two call sites now pass `api.route.current` after generalizing `syncTmuxPaneRegistration`.
- **`setup`** — new v2 path reusing the same internals:
  - sidebar via `ctx.ui.slot({ append: 'sidebar.content', render })`
  - v2 `ResolvedTheme` tokens mapped onto the flat shape `renderSidebar` already consumes (`text.default→text`, `text.subdued→textMuted`, `border.default→borderActive`, `background.default→background`; v2 has no `accent` token so the badge renders without a chip background)
  - directory from `ctx.location?.directory`, render loop via `ctx.renderer.requestRender()`
  - returns cleanup disposing slot + timer + tmux registration
  - no `/preset` registration (legacy `api.command` doesn't exist on v2) — matches the documented compatibility matrix
- `syncTmuxPaneRegistration` accepts both route shapes (v1 `{name, params:{sessionID}}`, v2 `{type:'session', sessionID}`) with runtime guards.
- v2 context types declared locally, mirroring the beta's `context.d.ts` (pinned `@opencode-ai/plugin@1.18.13` still ships old types).

## Verification

- `tsc --noEmit`: clean
- `bun test`: 2164 pass / 0 fail (full suite), incl. new dual-contract tests (module shape, slot claim + dispose, disabled-env short-circuit)
- Live against `@opencode-ai/cli@0.0.0-beta-17793`: TUI starts with **no** plugin-failed toast; server-side v2 adapter registers all bridges (health check passed)

## Notes

- No dependency changes; only `src/tui.ts` + `src/tui.test.ts` touched.
- Docs unchanged: no user-facing command/config/output differences on v1; v2 sidebar behavior matches the existing compatibility doc.

### DIFF
--- a/src/tui.test.ts
+++ b/src/tui.test.ts
@@ -272,13 +272,12 @@ describe('tmux pane registration', () => {
       ownerPid: 100,
       lastRecordedAt: 0,
     };
-    const api = {
-      route: {
-        current: { name: 'session', params: { sessionID: 'root-session-b' } },
-      },
-    };
 
-    syncTmuxPaneRegistration(api as never, registration, 1_000);
+    syncTmuxPaneRegistration(
+      { name: 'session', params: { sessionID: 'root-session-b' } },
+      registration,
+      1_000,
+    );
 
     expect(readTmuxPane('root-session-b', 1_000)).toBe('%42');
   });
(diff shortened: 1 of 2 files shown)


arpankanwer:
Addressed the Greptile P2 in 7daa361: cleanup now sets a disposal flag that's checked before each refresh tick and again after the async snapshot read, so an in-flight callback can no longer call `requestRender()` after the sidebar slot is disposed. Also condensed the added doc comments to match the surrounding style.

alvinunreal:
Thanks @arpankanwer

PR #1062, opened 2026-08-22
fix(config): preserve legacy continueOnIdle opt-out

### LINKED ISSUE #1050: Removed `backgroundJobs.continueOnIdle` is silently stripped; `orchestratorWake` defaults to enabled
`backgroundJobs.continueOnIdle` is not present in the 2.2.15 schema or runtime. The `backgroundJobs` schema object is not `.strict()`, so the key is silently discarded during validation. No warning or error is emitted.

`backgroundJobs.orchestratorWake` defaults to `{ "enabled": true, "intervalMs": 300000 }`. A configuration carrying `continueOnIdle: false` therefore runs with periodic wakes enabled, with no indication that the setting was dropped.

Two further observations in 2.2.15:

- The release notes and commit `2c69132` state the default wake interval was raised from 5 to 10 minutes. The shipped schema and runtime both use `300000` ms.
- `orchestratorWake` can start a parent turn via `session.promptAsync` while child tasks are still running.

Requested:

- Emit a warning for the removed key rather than stripping it silently, or map `continueOnIdle: false` to `orchestratorWake.enabled: false` for one release.
- Correct either the documented or the shipped wake interval default.

Version: 2.2.15

Iams4kura:

`backgroundJobs.continueOnIdle: false` was silently stripped after that key was
removed, allowing the replacement wake scheduler to default to enabled. This
change migrates boolean legacy values at the config-file loading boundary,
emits a deprecation warning, and preserves the opt-out across user/project
config layers. An explicit `orchestratorWake.enabled` in the same file wins.

The current five-minute wake interval was intentionally restored on `master`,
so this change leaves that default untouched.

Closes #1050

## Summary

- Fixes: A config containing backgroundJobs.continueOnIdle=false is accepted but the key is stripped, so orchestratorWake defaults to enabled and periodic wake prompts run against the user's explicit opt-out.
- Root cause: The background-jobs schema strips the removed key before the loader can preserve its boolean meaning, and per-layer nested defaults can overwrite a migrated user-level opt-out when a project-level backgroundJobs override is partial.

## Regression evidence

- Before: `bun test src/config/loader.test.ts -t "migrates deprecated backgroundJobs.continueOnIdle and warns"` exited `1`

- After: `bun test src/config/loader.test.ts -t "migrates deprecated backgroundJobs.continueOnIdle and warns"` exited `0`

## Verification

- `bun test src/config`
- `bun test`
- `bun run lint`
- `bun run typecheck`
- `bun run build`
- `./node_modules/.bin/biome check docs/configuration.md src/config/loader.ts src/config/loader.test.ts`

`bun run check:ci` was also run. It currently reports formatting drift in five
untouched files whose worktree hashes match `master`; the changed TypeScript
files pass the same pinned Biome check, and the repository's CI lint, typecheck,
test, and build commands all pass.

## Scope

- 3 files changed, +242 / -1 lines

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -315,6 +315,12 @@ Set `enabled: false` to keep idle reconciliation and background-job orchestratio
 without periodic wake prompts. See the
 [Background Orchestration](background-orchestration.md) guide for the concept,
 defaults, and examples.
+
+Configurations that still use the removed `backgroundJobs.continueOnIdle` key
+emit a deprecation warning and migrate its boolean value to
+`orchestratorWake.enabled`. An `orchestratorWake.enabled` value in the same
+config file takes precedence; replace the legacy key with that setting.
+
 `wallClockTimeoutMs` is a hard deadline that only supervises explicitly
 background native task calls; foreground calls or calls with `background`
 omitted are not supervised. It is independent from OpenCode's external
--- a/src/config/loader.test.ts
+++ b/src/config/loader.test.ts
@@ -605,6 +605,94 @@ describe('onWarning callback', () => {
     );
   });
 
+  test('migrates deprecated backgroundJobs.continueOnIdle and warns', () => {
+    const projectDir = path.join(tempDir, 'project');
+    const projectConfigDir = path.join(projectDir, '.opencode');
+    fs.mkdirSync(projectConfigDir, { recursive: true });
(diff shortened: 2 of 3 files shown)


alvinunreal:
thanks @Iams4kura

PR #1066, opened 2026-08-23
fix(background-jobs): grade stop evidence and deliver revival corrections

### LINKED ISSUE #1063: RC-B: Redesign background-task stop detection so idle or absent status is not treated as termination
> *This was generated by AI during triage.*

## Background

The background-task lifecycle cluster has one shared root cause that this issue tracks on its own: how `task-session-manager` decides a job has stopped. Keeping it separate from the foreground-fallback guard (RC-A, carried in #1011) lets each fix land and review independently.

## Problem

The stop policy treats a missing or idle session as proof that a background job ended. `getRuntimeSessionStatusSnapshot` says in its own docstring that an absent session "does not prove that a background task has terminated," but `runtime-status-reconciliation.ts` still feeds both `status === 'idle'` and `status === undefined` into `observeNonBusyRuntime` as equivalent stop evidence.

The policy then runs like this:

1. First non-busy observation: `noteStopConfirmation` plus `markStatusUncertain`.
2. Second observation 5 seconds or more later (`STOP_CONFIRMATION_GRACE_MS`, which matches the poll interval): `markStopped(...)`.

A job can be declared stopped about 10 to 12 seconds after launch if its session does not look busy in two snapshots in a row. Reasoning models and providers with variable time-to-first-token often pause for 5 to 15 seconds mid tool call, which trips this.

The reverse happens too. `markRunningFromLiveSession` revives a `stopped` job back to `running` on any later busy signal, after a recovery wake has already fired at the parent. That leaves the parent with contradictory instructions and repeats full-context wakes.

## Symptoms this explains

- #1033: background Fixer tasks stop with no terminal result, no tool call, no child process.
- #1044: false stopped-job recovery wake from a transient idle status, then a revive.
- #1053: subagents marked STOPPED_WITHOUT_TERMINAL_RESULT after the hardcoded 5s grace.
- #1054: the `status_uncertain` trail when a fallback model loses its terminal result.
- #1000: wake and token churn from the repeated recovery wakes.

## Proposed fix

Replace the ad-hoc heuristics (2s idle delay, 5s grace, 5s poll, dedup windows) with one liveness rule:

- The host status map is authoritative for positive busy and retry signals only.
- Absence is never negative evidence.
- A stop conclusion requires host-declared idle events, N consecutive event-level idles over at least 30 seconds, or a wall-clock deadline.
- All thresholds derive from a single `stopConfirmationMs` config value.
- Revival emits a corrective trailing notice instead of a bare contradiction.

Anything that injects content into the outgoing payload must go through `src/hooks/cache-safe-injection.ts` and stay in the tail zone. The revival notice and any new wake text fall under the prompt cache safety rules in AGENTS.md.

## Acceptance criteria

- [ ] An absent or idle session marks a job uncertain, never stopped.
- [ ] Confirmed stops require host idle events or a wall-clock deadline.
- [ ] The grace period covers TTFT pauses, at least 30 to 60 seconds, adaptive where possible.
- [ ] Revival emits a corrective trailing notice.
- [ ] `bun test` cache safety suites pass with no snapshot updates.
- [ ] `bun run check:ci`, `bun run typecheck`, `bun test` are green.

## Links

- Cluster parent: #595
- Foreground fallback guard (RC-A): #1011
- Closed as duplicate of this work: #1044
- Related, still open (verify each against this fix after merge): #1031, #1033, #1051, #1053
- Related tail behavior: #1054
- Wake churn: #1000

mhenke:
Fixes #1063

Background tasks were reported **stopped** without real evidence. Two concrete failures hit in practice:

1. A child session going idle — or being merely *absent* from a runtime status snapshot — was treated as termination. Live jobs were marked stopped mid-flight; parents received "stopped without a terminal result" notices for jobs that were still running or had just produced a result.
2. When a falsely-stopped job was revived, no correction reached the parent context — and under the `checkpoint-compatible` injection strategy the corrective notice was silently dropped entirely.

## What does this change?

Stop evidence is now graded instead of binary: host-declared idle starts a configurable confirmation clock (`backgroundJobs.stopConfirmationMs`, default 30s) that requires sustained non-busy observations before `markStopped`; absence from the runtime status map routes to uncertainty-only and never stops a job; busy/retry/live-busy evidence resets the clock. Reviving a genuinely-stopped job enqueues exactly one cache-safe corrective notice ("…was incorrectly reported stopped and is running again…"), delivered as a trailing volatile message under every injection strategy, and orphaned pending corrections are pruned when the parent session ends.

## Why this approach?

Alternatives considered: (a) keep treating any non-busy status as stop — this *is* the bug; (b) require explicit terminal events only — misses genuinely dead children after restarts/crashes, which operators do need reconciled (8 such confirmed stops occurred naturally during validation); (c) heuristic silence timeouts — strictly worse false-positive surface. Graded evidence with a confirmation grace keeps death detection while making idle/absence provably insufficient. Correction delivery reuses the existing cache-safe trailing-message channel (`src/hooks/cache-safe-injection.ts`) rather than mutating conversation history, preserving prompt-prefix stability.

## Related work

- Related: #1063 (this fix). Builds on #1026, #1007, #1008 (task lifecycle/recovery) and #837/#912 (idle/false-complete races); none address stop-evidence grading or revival corrections.

## AI assistance

- Model / harness: OpenCode orchestrator (`opencode/x-preview-f-free`) with @oracle reviews on `opencode/big-pickle` (4 independent passes: correctness, over-engineering, code-excellence, comprehensive) and @skeptic triage on `opencode/x-preview-f-free`. All review findings were fixed before commit.
- Human reviewed the full diff? [x] — reviewed interactively in-session prior to submission

## Checklist

- [x] `bun run check:ci`, `bun run typecheck`, and `bun test` pass — 2172 tests / 128 files, 0 fail
- [x] Docs updated: `docs/background-orchestration.md`, `docs/configuration.md`, codemaps for `src/utils/` and `src/hooks/task-session-manager/`

## Test results

**Unit gates:** `check:ci` ✅ · `typecheck` ✅ · `bun test` 2172 pass / 0 fail — includes new regression tests for revival-correction trailing-message delivery (and redelivery after a prior placement pass), orphan pruning on parent delete, absence→uncertainty-only routing, and checkpoint-compatible correction delivery.

**Live acceptance** (`acceptance-rc-b.md` executed cold-start in a separate CLI against `dist/` built from this commit — full spec posted as a comment below):

| Scenario | Verdict | Evidence |
|---|---|---|
| B — no premature stop on idle | PASS | idle 01:15:17 → grace clock 01:15:19 → completed via terminal result 01:16:05; zero stop lines |
| C1 — no spurious correction | PASS | revived never-stopped job → zero `Correction:` lines |
| D — cache-safe revival | PASS | zero `zero cached tokens` warnings across revival turn |
| A — absence ≠ stop evidence | PASS | zero stops fired off absence across entire log |

C2 (corrective notice on genuine stopped→running) and E (checkpoint-compatible strategy) are not reachable live without simulation/restart; both paths are covered by the unit suite above.

**Field log audit:** across ~2.5h of production logs spanning two sessions, 8/8 `confirmed runtime-stopped` events corresponded to genuinely dead sessions (process restarts/timeouts) — never idle or absence.


mhenke:
# RC-B Acceptance Tests — Background Stop-Evidence Policy

Live acceptance evaluations for the RC-B changes (issue #1063). Execute them
yourself, now, in this session. Do not ask what to do with this file — this
section IS the answer.

PROVENANCE: v1 of this spec produced B/C/D = INCONCLUSIVE on a real cold-start
run (the spawn instruction returned a terminal result, so the job completed
instead of stopping). v2 below makes every scenario decidable. Scenario A is
unchanged (it passed).

## Who you are and what you run

- You are an orchestrator agent inside an OpenCode session that has the
  oh-my-opencode-slim plugin loaded **from this worktree**. The plugin under
  test is the one managing YOUR background jobs right now.
- There is **no external CLI, no TUI, no second OpenCode**. Do not launch or
  drive any other OpenCode instance. You exercise the plugin by spawning real
  background subagents in your own session and reading the plugin log.

## How to perform each step

| Action | Meaning |
|---|---|
| SPAWN | Launch a background subagent (e.g. `task(..., background=true)`). Save the returned session/task ID — it looks like `ses_…`. |
| WAIT n | `bash`: `sleep n`. Real time. Do NOT message, reconcile, or check the job during the wait. |
| STATUS | Look up the job on your Background Job Board / `task_status(<ses_…>)`. |
| REVIVE | Resume the SAME session with a new instruction (e.g. `task_revive`). |
| LOG CHECK | `bash` grep against `$L` from the setup below. |

Rules:

1. **Verdicts come ONLY from LOG CHECK output.** Never from what a subagent
   claims, never from what you expect. Record evidence as log line numbers.
2. **No simulations.** Real subagents, real sleeps, real runtime events. If a
   condition cannot occur naturally in your environment, mark the scenario
   INCONCLUSIVE with one line of reason — do not fake it.
3. Run scenarios in order **B → C+D → A**. E is optional (needs a restart).
4. If genuinely blocked, mark INCONCLUSIVE and continue. Never ask questions.
5. Finish with a results table: `Scenario | PASS/FAIL/INCONCLUSIVE | evidence`.

## Expected lifecycle (verified against source and one live run)

- A spawned subagent that ends its turn ALWAYS produces a terminal result; the
  plugin parses it (`parsed task output status`, `hasResult:true`) and marks
  the job **completed**. Completion preempting a stop is CORRECT behavior.
- On the child's first host-declared idle, the plugin starts a 30s grace clock
  (log: `observed quiescent job from idle`).
- `markStopped` / `confirmed runtime-stopped` fires ONLY if the runtime keeps
  the session live-and-idle for ≥30s with no terminal result.
- A session ABSENT from the runtime status map routes to uncertainty-only and
  must NEVER be marked stopped. This is the core RC-B guarantee.
- Genuinely stopped jobs arise when a session dies without a result (process
  loss, kill) — you cannot manufacture one via a normal spawn. Do not try.

## Preconditions

- Plugin loaded from this worktree (`initialized with multiplexer config`
  log line directory matches) and `dist/` built from commit `6bc60202`+,
  with OpenCode restarted AFTER the build.
- `backgroundJobs.stopConfirmationMs` = default `30000`; strategy `latest`.

## One-time setup

```bash
L=$(ls -t ~/.local/share/opencode/log/oh-my-opencode-slim.*.log | head -1)
echo "$L"
```

---

## Scenario B — Idle never causes a PREMATURE stop (core RC-B guarantee)

Tests the anti-regression property: on pre-fix code, idle/absence produced
instant stops. Post-fix, a stop may only be confirmed ≥30s after sustained
idle — and a terminal result preempts stopping entirely.

1. SPAWN with exactly this instruction:

   > Acceptance test: do NO work, use no tools. End your turn immediately with
   > exactly one line: "standing by for go-ahead".

   Record `<TID>` (the `ses_…` ID).

2. WAIT 45.

3. STATUS: expect `completed` (terminal result preempts stop). Record actual.

4. LOG CHECK:

   ```bash
   grep -n "idle/status idle observed\|observed quiescent job from idle\|confirmed runtime-stopped\|markStopped\|parsed task output status" "$L" | grep "<TID>"
   ```

**PASS:** first idle is observed, then EITHER the job completes via terminal
result OR a confirmed stop appears ≥30s AFTER the first idle timestamp — and
NO stop line exists earlier than 30s after first idle.
**FAIL:** any confirmed-stop/`markStopped` line timestamped <30s after the
first idle observation. That is precisely the pre-fix bug.
**Note:** "completed instead of stopped" is a PASS here, not INCONCLUSIVE.

## Scenario C — Revival correctness (two branches)

**Branch 1 — negative guarantee (always runnable):**

1. REVIVE the Scenario B session with:

   > Go-ahead: reply with the single word "ack".

2. Inspect YOUR OWN context for any injected line starting
   `Correction: background job`, then LOG CHECK:

   ```bash
   grep -n "Correction: background job" "$L" | tail -10
   ```

**PASS:** revive succeeds, job runs, ZERO Correction lines anywhere (this job
was never stopped, so no correction may fire). **FAIL:** any Correction line.

**Branch 2 — positive path (ONLY if your board currently shows a job in
`stopped, unreconciled` state):**

1. REVIVE that stopped job with any short instruction.
2. Expect EXACTLY ONE `Correction: background job <alias> (<taskID>) was
   incorrectly reported stopped and is running again; disregard the earlier
   stopped notice.` line on the revival turn, GONE on your next turn.
3. Same LOG CHECK as Branch 1.

**PASS:** single occurrence on revival turn only, stripped next turn.
**FAIL:** missing, duplicated, or persistent. If no stopped job exists, record
Branch 2 as NOT RUN (cannot manufacture a stopped job without simulation).

## Scenario D — Revival traffic does not bust the prompt cache

Runs with Scenario C; no extra actions.

```bash
grep -n "zero cached tokens" "$L" | tail -5
```

**PASS:** no watchdog warning for your session across the revival turn(s).
**FAIL:** warning appears starting at a revival turn. If neither C branch ran,
INCONCLUSIVE.

## Scenario A — Absence is uncertainty-only (never stop evidence)

Pure log audit over everything run so far. No new jobs needed.

```bash
grep -n "markStopped\|confirmed runtime-stopped\|did not contain a live session" "$L" | tail -40
```

**PASS:** zero stop events, OR every stop is preceded earlier in the log by
idle observations spanning the 30s grace.
**FAIL:** any stop fired off absence alone (`did not contain a live session
state` immediately before a stop).

## Scenario E (optional, needs restart) — checkpoint-compatible strategy

Set `"strategy": "checkpoint-compatible"` under `backgroundJobs`, fully restart
OpenCode (this ends your session — do last or skip), redo setup, rerun B+C.

**PASS:** same criteria as B and C Branch 1 hold under this strategy.

---

## Coverage map

| Scenario | Behavior under test | Discriminates |
|----------|---------------------|---------------|
| B | no stop <30s after idle; completion preempts stop | FAILS on pre-fix code (instant stops) |
| C1 | no spurious correction for never-stopped jobs | H2 regression guard |
| C2 | one corrective notice for genuinely stopped→running | H2 (queue aliasing) |
| D | correction rides trailing volatile message, cache-safe | H1 (cache-prefix bust) |
| A | absence → uncertainty-only, never stop | RC-B core policy |
| E | checkpoint-compatible delivers corrections | M3 (strategy drop) |

Full `markStopped` confirmation-path coverage lives in the unit suite
(`bun test`, 253 passing); this file smoke-tests the guarantees reachable from
a live harness.

## Reference

- Policy: `observeNonBusyRuntime` (src/hooks/task-session-manager/
  stop-confirmation.ts) runs only for host-declared idle; first idle starts the
  grace clock, second after grace confirms `markStopped`. Absence routes to
  uncertainty-only and never starts confirmation.
- Notice text source: `board-injection.ts:1143`; enqueue condition:
  `event-router.ts:632` (stopped+terminalUnreconciled → running).

mhenke:
Closing this. Master landed the stop-evidence redesign that #1063 asked for: 2d975663 gates terminal publications on runtime evidence, stopConfirmationMs is in the config schema and wired in src/index.ts, and the old stop-confirmation.ts was deleted upstream. #1063 is closed. This branch now conflicts with all of that, so there is nothing left to merge.

PR #1086, opened 2026-08-24
fix(revived-run-tracker): gate background task notifications behind shouldManageSession

### LINKED ISSUE #1079: Background task lifecycle forces Orchestrator after switching parent to Plan or Build
## Environment

- oh-my-opencode-slim: 2.2.15
- OpenCode: 1.18.21

## Problem

Plan, Build, and Orchestrator are independently selectable primary modes. A background task lifecycle event can inject a new Orchestrator turn into the parent session after the user has switched that session to Plan or Build. The visible result is a normal Plan/Build response followed by an unsolicited Orchestrator response.

The exact stopped-job reminder is:

```text
A background job stopped without a terminal result. Consult the Background Job Board, recover or reroute the work as needed, and do not wait for that job as if it were still running.
```

## Reproduction

1. Use Orchestrator in a parent session.
2. Start a background specialist task.
3. Switch the same parent session to Plan or Build before the child reaches a terminal state.
4. Let the child stop without a terminal result.
5. Send or complete a normal prompt in the newly selected mode.
6. Observe an additional synthetic turn running as Orchestrator.

Completed/error background notifications appear to have the same cross-mode risk.

## Source evidence

In the published 2.2.15 artifact:

- `ORCHESTRATOR_STOPPED_JOB_WAKE_TEXT` is declared in `dist/index.js:28054-28056`.
- Stopped unreconciled jobs trigger recovery at `dist/index.js:45873-45876`.
- Recovery injects `session.promptAsync()` with hardcoded `agent: "orchestrator"` at `dist/index.js:28372-28383`.
- Completed/error parent notification also hardcodes `agent: "orchestrator"` at `dist/index.js:31140-31180`.
- Selected mode is tracked as one mutable session-level agent value at `dist/index.js:44901-44925` and updated by `chat.message` at `dist/index.js:46288-46304`.

A queued lifecycle prompt can therefore outlive a manual mode switch. Synthetic internal prompts can also overwrite the tracked agent back to Orchestrator.

## Expected behavior

Background task lifecycle events should not change or bypass the user-selected primary mode. Completion should update the Background Job Board passively, and the next manually selected Plan, Build, or Orchestrator turn should consume it.

## Suggested fix

- Deliver terminal notifications passively instead of starting a model turn with `agent: "orchestrator"`.
- Do not update selected-agent metadata for synthetic internal prompts.
- If automatic wake remains supported, capture a mode generation when scheduling and cancel before submission when the user changes mode.
- Add race tests for Orchestrator to Plan and Orchestrator to Build while a child completes, errors, or stops.

## Mitigation

`backgroundJobs.orchestratorWake.enabled: false` prevents periodic wakes and the exact stopped-job recovery path, but it does not cover the separate completed/error notification path.

mhenke:
Fixes #1079

Background task completion notifications inject `promptAsync({ agent: 'orchestrator' })` regardless of the parent session's current mode. When a user switches to Plan or Build, the notification still fires and its `agent: 'orchestrator'` flows through `chat.message` → `sessionMetadata.setAgent()` → flips the session metadata back to orchestrator, re-arming all orchestrator machinery (wake scheduler, tool gates, etc.).

## What does this change?

Adds `shouldManageSession` gate to `createRevivedRunTracker`. Guard in `notifyParent` checks whether parent session is still in orchestrator mode before injecting the synthetic prompt. 3 lines in source + 1 test case.

## Why this approach?

`shouldManageSession` is already the canonical predicate for orchestrator-mode gating, used 6+ times across the codebase (wake scheduler, task-session-manager, input-wait-tracker, cancel-task, task-revive, wait-for-user). Direct `sessionMetadata` access inside the tracker would couple it to a module it does not import. The optional callback preserves backward compatibility — callers that do not pass it get the old behavior.

## Related work

- Related: #1079 (root issue). No duplicate open/closed PR found for this exact bug.

## AI assistance

- Model / harness: OpenCode (deepseek-v4-flash) orchestrator; 4 @skeptic subagents for code review ran on opencode/x-preview-f-free (primary, with fallbacks opencode-go/qwen3.7-plus and opencode/qwen3.7-plus per `oh-my-opencode-slim.jsonc` skeptic preset).
- [x] Human reviewed the full diff before creating the PR.

## Checklist

- [x] `bun run check:ci`, `bun run typecheck`, and `bun test` pass
- [x] Docs updated if behavior changed — no doc change needed (internal callback only)

### DIFF
--- a/src/hooks/task-session-manager/revived-run-tracker.test.ts
+++ b/src/hooks/task-session-manager/revived-run-tracker.test.ts
@@ -6,6 +6,7 @@ function createHarness(
   messages: () => unknown,
   prompt = mock(async () => ({})),
   assertBound = false,
+  shouldManageSession?: (sessionID: string) => boolean,
 ) {
   const board = new BackgroundJobBoard();
   board.registerLaunch({
@@ -59,6 +60,7 @@ function createHarness(
     notificationRetryDelayMs: 0,
     onSettled: settled,
     pruneContext: pruned,
+    shouldManageSession,
   });
   return {
     board,
@@ -320,6 +322,31 @@ describe('revived run tracker', () => {
     harness.tracker.onTerminal(terminal);
     await new Promise((resolve) => setTimeout(resolve, 0));
     expect(harness.prompt).toHaveBeenCalledTimes(1);
+  });
+
+  test('skips parent notification when shouldManageSession returns false', async () => {
+    const harness = createHarness(
+      () => ({ data: [] }),
+      undefined,
+      false,
+      () => false,
+    );
+    harness.tracker.register({
+      taskID: harness.run.taskID,
+      generation: harness.run.generation,
+      parentSessionID: 'parent',
+      description: 'inspect the change',
+    });
+    const terminal = harness.board.updateStatus({
+      taskID: harness.run.taskID,
+      expectedGeneration: harness.run.generation,
+      state: 'completed',
+      resultSummary: 'done',
+    });
+    if (!terminal) throw new Error('missing terminal record');
+    harness.tracker.onTerminal(terminal);
+    await new Promise((resolve) => setTimeout(resolve, 0));
+    expect(harness.prompt).toHaveBeenCalledTimes(0);
   });
 
   test('discards a retry when the task generation is relaunched', async () => {
--- a/src/hooks/task-session-manager/revived-run-tracker.ts
+++ b/src/hooks/task-session-manager/revived-run-tracker.ts
@@ -68,6 +68,7 @@ export function createRevivedRunTracker(options: {
   onSettled?: (taskID: string) => void;
   contextFilesForPrompt?: (taskID: string) => ContextFile[];
   pruneContext?: () => void;
+  shouldManageSession?: (sessionID: string) => boolean;
 }): RevivedRunTracker {
   const runs = new Map<string, RevivedRun>();
   const maxNotificationRetries =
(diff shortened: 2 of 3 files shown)


dhaern:
Heads-up before either of these merges: #1195 is also open against #1079 and covers this same notification path, so the two will conflict in `notifyParent` (`revived-run-tracker.ts` + `index.ts`).

The difference is the semantics for the Plan/Build case:

- This PR gates `notifyParent` behind `shouldManageSession`: a parent that left orchestrator mode gets no injected turn and the result stays in the board.
- #1195 resolves the parent's current selection at send time and continues in it: a Plan parent receives the terminal inline as Plan (`agent` + model from the persisted selection), and the same treatment applies to the wake paths (periodic/children/stopped-recovery), which also hardcode `orchestrator` today. It additionally covers the stale-model axis (a manual model change falling back to the agent's default on the next wake), a v2 `switchModel` stale-pin revert, and the metadata poisoning that flips the tracked agent back to orchestrator after the first synthetic turn — which is what makes `shouldManageSession` itself report orchestrator afterwards.

Under #1195 this gate would be redundant: the tracker no longer hardcodes `orchestrator`, so there is nothing to suppress. The issue discussion mentions both directions (canceling the injection / passive delivery was one suggestion), so which behavior ships is the maintainer's call — just flagging the overlap so the merge order is a decision rather than an accident. Happy to rebase either way once there's a direction.

PR #1087, opened 2026-08-25
fix(foreground-fallback): match bailian quota-exhausted error and prevent backward chain fallback

### LINKED ISSUE #1083: [Bug]: bailian token-plan quota error ("quota has been exhausted" / insufficient_quota) not recognized as failover — fallback chain never consulted
### Summary

When a bailian (Alibaba Cloud Model Studio / token-plan) subscription expires, the provider returns an HTTP 429 whose message is:

```
Your token-plan 1-week quota has been exhausted. The quota will reset at 08-27 15:33:00 UTC.
```

with API-level `type`/`code` = `insufficient_quota`.

This error **matches none of the `RETRYABLE_ERROR_PATTERNS` / `PROVIDER_OUTAGE_PATTERNS`** in `src/hooks/foreground-fallback/index.ts`, so `isFailoverError()` returns `false`, the fallback chain is **never consulted**, and every subagent session on that model just dies (`AI_APICallError`) instead of switching to the next model in its configured array.

### Environment

- oh-my-opencode-slim: **2.2.15**
- opencode: **1.18.22**
- OS: Linux
- Config: per-agent model arrays (preset form), e.g. explorer:
  ```json
  ["bailian-cli/deepseek-v4-flash-0731", "ollama-cloud/deepseek-v4-flash:0731", "opencode-go/deepseek-v4-flash", "opencode/north-mini-code-free", "openrouter/deepseek/deepseek-v4-flash-0731"]
  ```

### Why it doesn't match

Checked against the current pattern lists:

| Pattern | Message | Match |
|---|---|---|
| `/quota.?exceeded/i` | "quota **has been exhausted**" | ❌ |
| `/resource.?exhausted/i` | no "resource" token | ❌ |
| `/insufficient.?(quota\|balance)/i` | word "insufficient" only appears in the API `code` field, not in `error.message` | ❌ |
| status-code path (`statusCode === 429`) | the raw API response is 429, but the serialized error reaching the fallback manager apparently carries only the message text — consistent with the `extractStatusCode` gap already described in #955 | ❌ |

Raw provider response (API key redacted):

```
HTTP/429
{"error":{"message":"Your token-plan 1-week quota has been exhausted. The quota will reset at 08-27 15:33:00 UTC.","type":"insufficient_quota","code":"insufficient_quota"}}
```

### Evidence (single day, one machine)

`opencode.log` — six subagent sessions failed hard on bailian, zero `[foreground-fallback]` log lines for any of them:

```
19:00:40 stream error providerID=bailian-cli modelID=deepseek-v4-flash-0731 agent=explorer mode=subagent  AI_APICallError: Your token-plan 1-week quota has been exhausted...
19:24:32 stream error ... agent=explorer mode=subagent   (same)
19:39:06 stream error ... agent=explorer mode=subagent   (same)
19:47:57 stream error ... agent=explorer mode=subagent   (same)
20:04:30 stream error ... agent=fixer   mode=subagent   (same)
20:04:38 stream error ... agent=fixer   mode=subagent   (same)
```

Contrast: an error whose message *does* match a pattern triggers fallback normally:

```
19:21:04 stream error providerID=opencode-go "Monthly usage limit reached..."   ← matches /monthly usage limit/i (added by #620)
19:21:04 [foreground-fallback] switched to fallback model {agentName: orchestrator, from: opencode-go/deepseek-v4-flash, to: bailian-cli/qwen3.8-max}
```

This is the third instance of the same bug class: #620 ("Monthly usage limit reached") and #957 (`auth_unavailable`) were both fixed by adding patterns.

### Secondary observation: backwards fallback onto the dead head

When a session starts on a model that is **not** `chain[0]` (e.g. after a manual `/model` switch or a prior generation's switch), the first failure sends it *back up* to `chain[0]`, because `sessionTried` only records models that failed through tracked fallback events:

```
15:44:58 [foreground-fallback] switched to fallback model {agentName: orchestrator, from: ollama-cloud/glm-5.2,        to: bailian-cli/qwen3.8-max}   ← dead provider
19:21:04 [foreground-fallback] switched to fallback model {agentName: orchestrator, from: opencode-go/deepseek-v4-flash, to: bailian-cli/qwen3.8-max}   ← dead provider
```

`execFallback` picks `chain.find(m => !tried.has(m))` which yields `chain[0]` even when `currentModel` sits at a later index. Seeding `tried` with every chain entry at or before the current model's index would fix this.

### Suggested fix

1. Add failover patterns covering this class, e.g.:
   - `/\bquota\b[^.]*\bexhausted/i`
   - `/insufficient.?quota/i` (catches the API `code` when `responseBody`/`data` is retained)
2. Per #955's "chain gap" item: instrument `isFailoverError` to confirm whether `statusCode` survives provider → SDK → event → fallback manager; if it does not, a 429-only classification silently depends on message text alone.
3. Seed `sessionTried` from the current model's chain position to avoid backwards fallback onto a known-dead head.

### Related

- #955 (consolidated foreground-fallback bugs — dedup window swallowing second failure also reproduced here at 19:21:07)
- #966 (opposite misclassification of quota errors)
- #620, #957 (same class, fixed by adding patterns)

mhenke:
## Description

Fixes #1083 — bailian's token-plan quota exhaustion error (`"quota has been exhausted"`) was not matched by the existing `/quota.?exceeded/i` pattern, preventing the foreground fallback chain from being consulted.

### Changes

1. **Pattern fix** — added `/\bquota\b.*\bexhausted/i` to `RETRYABLE_ERROR_PATTERNS` to match bailian's exact error message: "Your token-plan 1-week quota has been exhausted. The quota will reset at 08-27 15:33:00 UTC."

2. **Backward fallback guard** — in `execFallback`, after adding `currentModel` to the `tried` set, also seed all chain entries at indices `0..idx-1`. This prevents a session that was already moved past `chain[0]` from falling backward onto the dead head model when a downstream model fails.

### Verification

- New test case covers the exact bailian error message
- All 2214 existing tests pass
- `bun run check:ci` and `bun run typecheck` clean

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -121,6 +121,15 @@ describe('isFailoverError', () => {
     );
   });
 
+  test('returns true for bailian "quota has been exhausted" (issue #1083)', () => {
+    expect(
+      isRetryableError({
+        message:
+          'Your token-plan 1-week quota has been exhausted. The quota will reset at 08-27 15:33:00 UTC.',
+      }),
+    ).toBe(true);
+  });
+
   test('returns true for codex quota-threshold errors', () => {
     expect(
       isFailoverError({
--- a/src/hooks/foreground-fallback/index.ts
+++ b/src/hooks/foreground-fallback/index.ts
@@ -37,6 +37,7 @@ const RETRYABLE_ERROR_PATTERNS = [
   /rate.?limit/i,
   /too many requests/i,
   /quota.?exceeded/i,
+  /\bquota\b.*\bexhausted/i,
   /quota.?threshold/i,
   /usage.?exceeded/i,
   /ExceededBudget/i,
@@ -678,6 +679,12 @@ export class ForegroundFallbackManager {
       // biome-ignore lint/style/noNonNullAssertion: We just set this above
       let tried = this.sessionTried.get(sessionID)!;
       if (currentModel) tried.add(currentModel);
+      // ponytail: seed chain entries at or before the current model's index
+      // to prevent backward fallback onto models the session already left.
+      if (currentModel) {
+        const idx = chain.indexOf(currentModel);
+        for (let i = 0; i < idx; i++) tried.add(chain[i]);
+      }
 
       let nextModel = chain.find((m) => !tried.has(m));
       if (!nextModel) {


mhenke:
Closing — created without following the omos-create-pr workflow. Reopening properly.

PR #1089, opened 2026-08-25
docs: document OpenCode v2 migration

alvinunreal:
## Summary\n- pin the local OpenCode source reference to the v2 beta commit and add its codemap\n- document the current v1-to-v2 adapter, its limitations, and the native v2 migration plan\n- correct v2 configuration and capability guidance; add README and clonedeps references\n\n## Verification\n- Checked 329 files in 216ms. No fixes applied.\n- \n- documentation accuracy review against the pinned OpenCode beta source


alvinunreal:
Moving this documentation work directly to the beta branch.

PR #1092, opened 2026-08-25
feat(config): support explicit agent model inheritance

### LINKED ISSUE #1025: Feature Request: Independent model assignment for librarian and fixer (break the inheritance chain)
### Problem

Currently, when `fixer` has no explicit `model` set, **oh-my-opencode-slim inherits the `librarian`'s model**. This makes it impossible to have:

- `librarian` → cheap local model (e.g. `ollama/qwen3.8:27B`) for research/doc lookups
- `fixer` → frontier/session model (e.g. `zai-coding-plan/glm-5.2`) for implementation work

Setting both explicitly works, but there's no way to say _"use the session/orchestrator model"_ for one agent while pinning another to a local cheap model. The only workaround is an external plugin wrapper that strips the inherited model after the fact.

**I want/need this librarian/fixer decorrelation because my cheap model is ollama local model, such as ollama/qwen3.8:27B, and i noticed too many issues when qwen is the fixer, I prefer to avoid it, but it's a really good model for other purpose (librarian, git-ops tasks, tracer, etc.). Maybe someday qwen 4.x or 5.x could fixes these issues in still a small distilled version, finger crossed :).**

### Current behavior (the inheritance chain)

```
librarian.model = "ollama/qwen3.8:27B"
fixer.model    = (not set)
  → fixer silently inherits librarian.model  ← unwanted coupling
```

### Desired behavior

**Option A — explicit sentinel (preferred):**

```jsonc
{
  "agents": {
    "librarian": {
      "model": "ollama/qwen3.8:27B"
    },
    "fixer": {
      "model": "$session"   // or "$primary", "$orchestrator"
    }
  }
}
```

`$session` / `$primary` / `$orchestrator` would mean _"follow whatever model the current session/orchestrator is using"_, breaking the librarian→fixer inheritance.

**Option B — config flag (simpler):**

```jsonc
{
  "agents": {
    "librarian": {
      "model": "ollama/qwen3.8:27B"
    },
    "fixer": {
      "model": "ollama/qwen3.8:27B",
      "inheritModelFrom": "session"   // or "orchestrator", "primary"
    }
  }
}
```

**Option C — convention (zero-config):**

If `fixer` has no `model` and `librarian` has an explicit `model`, treat `fixer` as _"use the session model"_ rather than inheriting from librarian. This is the most intuitive behavior: the implementation agent should default to the session's main model, not the research agent's cheap local model.

### Why this matters

- **Cost control**: Librarian does heavy web/doc research on a cheap local model; fixer does code implementation on a frontier model. Mixing them means either expensive research or weak implementation.
- **Separation of concerns**: Research quality and coding quality have different model requirements. Forcing them to share a model couples two unrelated cost/quality trade-offs.
- **Current workaround is fragile**: Requires a custom plugin (`oh-my-opencode-slim-plugin.ts`) that intercepts the config hook and strips `fixer.model` post-build. This breaks on any internal refactor of the agent config shape.

### Workaround today (for reference)

```ts
// oh-my-opencode-slim-plugin.ts
const wrapper = async (input: any) => {
  const hooks = await slim(input)
  if (Array.isArray(hooks.agent)) {
    hooks.agent.forEach(a => {
      if (a.name === "fixer" && a.config) a.config.model = undefined
    })
  }
  // also strip after config hook runs
  const originalConfig = hooks.config
  if (typeof originalConfig === "function") {
    hooks.config = async (cfg: any, ...rest: any[]) => {
      const result = await originalConfig(cfg, ...rest)
      if (cfg?.agent?.fixer) cfg.agent.fixer.model = undefined
      return result
    }
  }
  return hooks
}
```

### Related

- The inheritance behavior is documented implicitly in the codebase but not as a user-configurable feature.
- Similar use case: any user who wants a cheap research lane + a strong coding lane in the same session.

HeZ2z:
Fixes #1025

When `fixer` has no explicit model, oh-my-opencode-slim inherits the `librarian` model. This makes it difficult to use a cheap model for research while keeping a stronger session or orchestrator model for implementation.

The same coupling can affect the reverse setup: users may want Librarian to follow the orchestrator model while Fixer uses an independently configured model.

## What does this change?

- Add `inheritModelFrom` to agent configuration.
- Support `inheritModelFrom: "session"` to use the current parent session model.
- Support `inheritModelFrom: "orchestrator"` to use the orchestrator model resolved during plugin configuration.
- Preserve explicit `model` precedence.
- Preserve the existing fixer -> librarian fallback when neither `model` nor `inheritModelFrom` is configured.
- Clear stale lower-layer or host models when inheritance is explicitly requested.
- Support root agent overrides, preset overrides, and custom agents.
- Update the generated JSON Schema and configuration documentation.
- Add unit and `hooks.config()` integration coverage.

Runtime fallback is intentionally out of scope. `"orchestrator"` refers to the model resolved during configuration and does not dynamically follow later foreground fallback changes.

## Why this approach?

An omitted agent model already tells OpenCode to use the current parent session model, so `"session"` can use the host runtime behavior directly without introducing a synthetic model identifier.

The inheritance directive is applied after the plugin and host configuration layers are merged. This is necessary to remove stale host or preset models when the user explicitly requests session-based inheritance.

The default fixer -> librarian behavior remains unchanged for backward compatibility.

## Related work

- [x] Checked open and closed issues and PRs for duplicates.
- Related: #1025

## Verification

- [x] Related tests pass - 164 passed
- [x] `bun run typecheck` - clean
- [x] Changed-file Biome checks - clean
- [x] `git diff --check` - clean
- [x] Generated schema is synchronized
- [ ] `bun run check:ci` - blocked by two pre-existing formatting issues in unrelated task-session files
- [ ] `bun test` - 2238 passed; 4 pre-existing interview dashboard integration tests failed

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -77,6 +77,7 @@ A glossary of the terms used in this project's domain. Definitions describe what
 - **Plugin config** — The user-facing configuration loaded from `oh-my-opencode-slim.jsonc`.
 - **Preset** — A named set of per-agent overrides. The same word also names council councillor lineups (see Flagged).
 - **Model entry** — A normalized model reference with an optional variant, used in fallback chains.
+- **Model inheritance** — An explicit agent policy selecting the current `session` model or the configured `orchestrator` model when no agent model is set.
 - **Variant** — An optional model qualifier (e.g., a preview build) used in fallback resolution.
 - **Fallback / failover** — The mechanism that switches models when a call is rate-limited or returns empty.
 - **Disabled agents** — Agents turned off via config; `observer` is disabled by default.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -361,6 +361,47 @@ Notes:
 - Display names must be unique
 - Display names cannot conflict with internal agent names like `oracle` or `explorer`
 
+### Independent agent model inheritance
+
+By default, the `fixer` agent inherits the `librarian` model when no fixer
+model is configured. To decouple agents, set `inheritModelFrom` on the agent
+that should follow the current session or the configured orchestrator model:
+
+```jsonc
+{
+  "agents": {
+    "librarian": {
+      "model": "ollama/qwen3.8:27B"
+    },
+    "fixer": {
+      "inheritModelFrom": "session"
+    }
+  }
+}
+```
+
+Supported values are:
+
+- `session`: omit the agent model so OpenCode uses the current session model
+- `orchestrator`: use the orchestrator model resolved during configuration; if
+  none is configured, fall back to the current session model
+
+`orchestrator` means the model resolved during plugin configuration. It does
+not dynamically follow a later foreground fallback to another model.
+Runtime fallback behavior is independent of `inheritModelFrom`.
+
+Model selection follows these rules:
+
+- If the same effective agent override contains both `model` and
+  `inheritModelFrom`, the explicit `model` wins.
+- If `model` is omitted, `inheritModelFrom` is an explicit higher-layer
+  directive: it clears a lower-layer `model` value, including a model supplied
+  by the host agent configuration, and resolves the requested source.
+- If neither field is present, the existing model precedence and the historical
+  fixer-to-librarian fallback remain unchanged.
+
+The setting works in both root `agents` overrides and preset agent overrides.
+
 ### Per-preset agent configuration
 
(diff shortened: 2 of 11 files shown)


alvinunreal:
thanks @HeZ2z

PR #1108, opened 2026-08-28
fix(orchestrator): strengthen delegation discipline for strong models

### LINKED ISSUE #1012: [Bug]: Orchestrator forcing its usage

opencode latest 

With a strong model such as the new DeepSeek V4 Pro (New) opencode GO, the Orchestrator starts taking over most of the actual work instead of distributing it to the other agents.

Normally, when watching its reasoning, you can clearly see the Orchestrator coordinating the other agents and deciding who should handle what. With a strong model, however, you can literally watch it take over the task itself.

It starts reasoning through the problem directly, solving things on its own, and effectively ignoring the other agents that are supposed to be doing the work.

It feels like the old-school “good manager” problem: the manager knows how to do everything, so instead of managing the team, he just sits down and does the work himself. 

The stronger the model, the more noticeable this becomes. It seems to forget its role as Orchestrator and gradually turns into the main implementation agent.

What is especially interesting is that it seems to need constant reminders about its role. Without those reminders, it starts taking over more and more of the actual work and leaves the other agents out of the process.

In short: the better the underlying model, the more the Orchestrator wants to become the worker instead of the manager.

### Steps to reproduce

_No response_

### oh-my-opencode.json

```jsonc
{
  "preset": "opencode-go",
  "presets": {
    "opencode-go": {
      "orchestrator": {
        "skills": [
          "*"
        ],
        "mcps": [
          "*",
          "!context7"
        ]
      },
      "oracle": {
        "model": "opencode-go/gpt-5.6-luna",
        "skills": [
          "simplify",
          "vision"
        ],
        "mcps": []
      },
      "librarian": {
        "model": "opencode-go/mimo-v2.5",
        "skills": [],
        "mcps": [
          "websearch",
          "context7",
          "gh_grep"
        ]
      },
      "explorer": {
        "model": "opencode-go/mimo-v2.5",
        "skills": [],
        "mcps": []
      },
      "designer": {
        "model": "opencode-go/kimi-k2.7-code",
        "variant": "medium",
        "skills": [
          "vision"
        ],
        "mcps": []
      },
      "fixer": {
        "variant": "medium",
        "skills": [],
        "mcps": []
      },
      "observer": {
        "model": "opencode-go/minimax-m3",
        "variant": "medium",
        "skills": [
          "vision"
        ],
        "mcps": []
      }
    }
  },
  "disabled_agents": [],
  "companion": {
    "enabled": false,
    "position": "bottom-right",
    "size": "medium",
    "gifPack": "default",
    "loopStyle": "classic",
    "speed": 1,
    "debug": false
  }
}
```

### OpenCode version

1.18.18

_No response_

### Operating system

all of them - using wiith windows/mac and linux

```shell

```

mhenke:
Fixes #1012

**What changed, and why was it needed?**

The orchestrator prompt let strong models keep coding work instead of delegating. The old prompt had "Handle work directly when delegation overhead exceeds doing it yourself" — a strong model never sees a task as high-risk or delegation as worth the overhead. The "Don't delegate when" sections (fixer's "single small change", oracle's "first bug fix attempt", council's "routine implementation") were written for weaker models. Strong models read them as instructions.

We removed all escape hatches while keeping the capability descriptions that help the model route work: "Delegate when" sections, stats for cost/speed decisions, and the role identity. Added the task() mechanism in the routing section so the model knows how to dispatch work.

**Changes**

- Removed all "Don't delegate when" sections from every agent description
- Removed "handle directly", "answer directly", "delegation overhead exceeds" permissions
- Removed "Do not delegate merely because an agent exists"
- Kept "Delegate when" sections (capability descriptions for routing)
- Kept stats (cost/speed information for routing decisions)
- Kept role identity ("workflow manager... not the default implementation worker")
- Added explicit task() tool mechanism in routing section
- Kept UI/design work guardrail

**Verification**

- 11 tests verify no escape hatches, task() mechanism present, Delegate when present
- bun run check:ci — clean
- bun run typecheck — clean
- bun test — 2250 pass, 0 fail
- Snapshots updated


alvinunreal:
@mhenke - I'd be careful with changing system prompt

I know its not perfect but the reported issue is for deepseek and we don't know how it will change other models.
Users can always customize their orchestrator prompt but I'd say we keep prompt as is only make small iterative changes manually - no AI coding for prompts.

mhenke:
@alvinunreal Thanks. I thought I had this closed. I'm working on some evals and was going to run them through as orchestrator using deepseek v4 pro

mhenke:
@alvinunreal you can take a look here omos/evals-clean https://github.com/alvinunreal/oh-my-opencode-slim/tree/omos/evals-clean  so far a problem is trying speed up the evals

PR #1115, opened 2026-08-29
Reject textless completed task results

Alfiegerner:
## Reject textless completed task results

“completed” is OMO Slim’s terminal subagent-task state. This change prevents a task with no public result text from being recorded as a successful completion.

- Reject textless completed status reports unless text was already retained.
- Re-read a revived task briefly for delayed persistence; then record an explicit error rather than false success.
- Cover direct status ingestion, board updates, and foreground task hooks.

A textless completion can occur when the host marks a turn terminal before its final text part is persisted or returned.

---
Raised with OMO Slim for review.

## Validation

- bun run check:ci
- bun run typecheck
- focused task-status, board, revived-run, and retrieval tests


Alfiegerner:
@greptileai please re-review this PR after fd6e87d: the revived-run stabilization probe now schedules its terminal error transition, with an automatic scheduling regression test.

alvinunreal:
thanks @Alfiegerner

PR #1129, opened 2026-09-02
fix(tui-sidebar): long model names wrapping in the TUI sidebar

### LINKED ISSUE #1128: [Bug]: Long model names break alignment in the TUI sidebar
## What happened

Long model names wrap onto a second line in the compact TUI sidebar. The continuation is not aligned with the model column, so some rows become taller than others and the list looks uneven.

This is easiest to see with entries such as:

- `observer` — `muse-spark-1.2-contributor-free`
- `plan` — `nemotron-3-ultra-free`

## What I expected

Each agent should stay on one line. The model name should remain aligned with the other model names, or be truncated cleanly when the sidebar is too narrow.

## Steps to reproduce

1. Configure an agent with a long model ID.
2. Open OpenCode with the compact sidebar enabled.
3. Look at the agent/model list in the sidebar.

## Environment

- oh-my-opencode-slim: 2.2.18
- Operating system: macOS

## Additional context

The problem appears to be limited to the sidebar layout. Shorter model names line up correctly; longer names wrap and break the row spacing.

vmvarela:
## Summary

- Keep long model names on one line in the compact TUI sidebar.
- Truncate model text when the sidebar is too narrow instead of wrapping it.
- Keep the agent/activity column fixed and add a focused regression test.

## Testing

- `bun test src/tui.test.ts`
- `bun run typecheck`
- `bun run build:plugin`
- `bun run build:tui`

Fixes #1128

image

### DIFF
--- a/src/tui.test.ts
+++ b/src/tui.test.ts
@@ -82,6 +82,89 @@ describe('tui sidebar agents', () => {
     expect(getSidebarActivityIndicator(true, 100)).toBe('⠙');
     expect(getSidebarActivityIndicator(true, 1_000)).toBe('⠋');
   });
+
+  test('keeps compact agent rows single-line with truncated right-aligned model IDs', async () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'omos-compact-row-'));
+    const projectDir = path.join(root, 'project');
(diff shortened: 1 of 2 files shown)


alvinunreal:
thanks @vmvarela

PR #1134, opened 2026-09-04
fix: always name the session agent on plugin prompts (prevents silent rewrite to `build`)

tsankotsanev:
## The bug

When a `session.prompt` / `session.promptAsync` body omits `agent`, opencode
resolves the agent with `agents.defaultInfo()` — the first visible `primary`
agent, i.e. built-in **`build`** whenever `default_agent` is unset,
user-overridden, or not effectively applied. It then compares that against the
session's current agent and calls `sessions.setAgentModel()` when they differ
(`session/prompt.ts:672-689`), which **permanently rewrites the session's
agent**.

So a single agent-less plugin prompt doesn't just mislabel one turn — it
re-homes the whole session, and every later turn is served by `build` with
`build`'s prompt and toolset.

I reproduced this empirically against opencode 1.18.26 in an isolated temp
project with a purpose-built `probe-sub` subagent:

```
child agent BEFORE prompt:      probe-sub
promptAsync(child, no agent) -> HTTP 204  error=null
child agent IMMEDIATELY after:  build
child agent AFTER turn:         build
>> agent field CHANGED probe-sub -> build
```

The rewrite lands **on accept** (204, before generation starts) and persists.
Per-message attribution confirms the session was genuinely re-homed rather than
mislabelled — `user/probe-sub`, `assistant/probe-sub`, then `user/build`,
`assistant/build`, inside a session that is still a `probe-sub` subagent child.
The control run, identical but with `agent` supplied, left `session.agent`
untouched and still delivered the injected instruction — so naming the agent is
a complete fix, not a workaround.

Full write-up and raw evidence:
`docs/agents/build-agent-empty-input-diagnosis.md`. This is the same root class
as the `/preset` fix in #818; that doc previously called it "resolved for
interview injections", but two call sites were still exposed.

## The two agent-less sites on master

**1. `src/interview/runtime.ts` — `notify()`.** Easy to miss because it passes
`noReply: true`. `noReply` suppresses the *model turn*, not the agent
resolution, so the rewrite happened anyway on a session the user is actively
using. It now resolves the session's own agent and falls back to `orchestrator`
only for a probe-confirmed parentless session.

**2. `src/tools/smartfetch/secondary-model.ts`.** A throwaway helper session
created and deleted within the call. It has no history, so there is nothing to
preserve — but the field must still be *named*, so it is set inline to `build`
via `FALLBACK_HELPER_SESSION_AGENT`.

`build` **deliberately, never `orchestrator`**: a message tagged `orchestrator`
makes the task-session-manager adopt the throwaway session as a managed
orchestrator session (`registerSessionAsOrchestrator`), which then attracts
phase reminders, post-tool nudges, background-job-board injection, companion
busy/idle flips, and the orchestrator system prompt — billed on every fetch.
`build` is always registered (`disableDefaultAgents` explicitly preserves
`build`/`plan`) and no hook keys off it.

All other v1 prompt sites already named their agent
(`orchestrator-wake`, `task-revive`, `task-message`, `foreground-fallback`,
`interview` `continue`).

## The resolver

`src/utils/prompt-agent.ts` is now the single place that decides the field:
`resolveSessionAgent()` + `withAgent()`. Rules, each one load-bearing:

- **Order:** recorded hint → `session.get().agent` → newest **user** message →
  fallback.
- **User messages only.** Assistant messages report whichever agent *served* a
  turn, which includes core's `compaction` / `summary` / `title` primaries
  writing into the user's own session — and includes `build` in any session
  this bug already re-homed. Trusting them would make the resolver launder the
  corruption it exists to prevent.
- **`compaction` / `summary` / `title` are denylisted** at every derived step.
  They are real registered primaries, so they are otherwise indistinguishable
  from a legitimate agent, and prompting as one would re-home the session just
  as badly.
- **Fallback only for a session confirmed parentless** (or declared plugin-owned
  via `assumeTopLevel`). An unresolvable *child* session gets **no `agent`
  field** rather than a guess — guessing would cause the exact rewrite this
  module prevents. `orchestrator` is safe as the top-level fallback because it
  is in `PROTECTED_AGENTS` and can never be disabled by user config.
- Probe reads accept an optional `directory`, forwarded as `query.directory` to
  match every other `session.get` / `session.messages` call site.

## Regression scan

`src/utils/prompt-agent.test.ts` walks all of `src/` and fails the build on any
`prompt` / `promptAsync` call site that neither passes `agent:` inline nor
routes its body through `withAgent()`. It also follows
`session.prompt.bind(session)` aliases (used by `task-message.ts`) and
`promptWithTimeout` wrapper calls, which a naive regex would miss.

Notably it uses **no receiver-name filter**: an earlier `/session/i` heuristic
hid `src/v2/client-shim.ts`'s own `s.prompt(` site. Four narrow, individually
documented allowlist entries:

| File | Why |
|---|---|
| `utils/session.ts` | `promptWithTimeout` is a pass-through wrapper; its call sites are scanned instead |
| `v2/session-submit.ts` | v2 flat host `PromptInput` has no `agent` field |
| `v2/client-shim.ts` | same; plus a recorded follow-up — the shim currently *drops* `body.agent` when adapting v1 → v2 and should call `session.switchAgent`, documented in `docs/opencode-v2-compatibility.md` |
| `utils/session-calls.contract.ts` | type-only contract whose `@ts-expect-error` blocks intentionally hold rejected bodies |

## Tests

`bun run typecheck`, `bun test`, `bun run build` all clean.

- Full suite: **2523 pass / 0 fail** (144 files)
- New: `prompt-agent.test.ts` **43 tests** (unit + the repo-wide scan)
- New: `session-error.test.ts` **14 tests**
- `interview/runtime.test.ts`: +3 (agent preserved; derived from newest user
  message, not a compaction turn; child session gets no agent)
- `smartfetch/secondary-model.test.ts`: +1 (helper names `build`, never
  `orchestrator`, and resolves without probing)

`oh-my-opencode-slim.schema.json` is intentionally untouched. Heads-up: `bun
run build` regenerates it and currently produces a one-line diff on master
(`backgroundJobs.defaultConcurrency` description) — the committed schema is
slightly stale versus `src/config/schema.ts`. Pre-existing and unrelated, so I
reverted the regeneration rather than smuggle it in here.

### One note on scope

`src/utils/session-error.ts` (+ its tests) is a small standalone classifier —
`isTransportError` / `isAbortedSessionError` / `sessionErrorMessage` — that
landed alongside this work while probing the same area. It has no callers in
this PR yet; it carries a drift guard pinning its copied transport patterns
against `src/hooks/foreground-fallback/index.ts` (copied, not imported, because
`src/utils` must not depend on `src/hooks`). Happy to drop it from this PR and
send it separately with its consumer if you'd prefer to keep this strictly to
the agent fix.

### DIFF
--- a/docs/agents/build-agent-empty-input-diagnosis.md
+++ b/docs/agents/build-agent-empty-input-diagnosis.md
@@ -1,7 +1,7 @@
 # Diagnosis: "build agent empty input" after orchestrator output
 
-**Status:** Resolved for interview injections.
-**Date:** 2026-08-10
+**Status:** Resolved repo-wide and guarded by a regression scan.
+**Date:** 2026-08-10 (interview fix) · 2026-09-02 (repo-wide fix + scan)
 **Related PR:** #818 (`fix/preset-tui-slash-command`) — same root class as the original `/preset` fix.
 **Suspected sibling bug reported by user:** During `superpowers` / `brainstorm` skill conversations, when the orchestrator asks for confirmation or work is interrupted (subagent completes, background task finishes), a `build` agent turn sometimes appears with an empty user input.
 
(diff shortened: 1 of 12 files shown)


alvinunreal:
@orchestrat0r fix comments

alvinunreal:
@orchestrat0r fix conflicts mare pr ready to bemerged fix greptile too

alvinunreal:
@greptileai

alvinunreal:
@orchestrat0r fix conflicts mare pr ready to bemerged fix greptile too

PR #1135, opened 2026-09-04
fix: trigger model fallback on content-policy moderation rejections

dhaern:
# fix: trigger model fallback on content-policy moderation rejections

## Problem

When OpenAI flags a request under its cyber-security content policy, the request fails outright with no model fallback:

```
AI_APICallError: This content was flagged for possible cybersecurity risk.
If this seems wrong, try rephrasing your request. To get authorized for
security work, join the Trusted Access for Cyber program: https://chatgpt.com/cyber
```

The error surfaces as HTTP 400 `invalid_request` with the structured code `cyber_policy` in the body. `isFailoverError()` intentionally treats unrecognized 400s as hard errors (a genuine `invalid_request` would be reproduced by the next model in the chain), so the moderation rejection never reaches the fallback pipeline, neither for foreground sessions nor for delegated subagent sessions (the task-session-manager event router imports the same classifier).

## Why fallback is correct here

A content-policy rejection is deterministic per provider: retrying the same model will fail again (observed back-to-back in production logs), but other providers in the fallback chain do not share the policy. This is the same family as the already-handled `403/forbidden` case: a provider-side refusal that the next model can serve.

## Change

Two pattern groups in `RETRYABLE_ERROR_PATTERNS` (`src/hooks/foreground-fallback/index.ts`):

- `/\bcyber_policy\b/` and `/\bcontent_policy_violation\b/`: the structured provider error codes (may appear in `responseBody` JSON)
- `/flagged for possible cybersecurity risk/i` and `/rejected as a result of our safety system/i`: the exact provider wording

Generic "flagged"/"policy"/"cybersecurity" wording is deliberately not matched, so ordinary error text stays a hard error.

Because both the foreground manager and the delegated-session event router use this classifier, the fix covers interactive sessions and subagents, and the rejection inherits the existing retry budget (`fallback.maxRetries`, default 3), dedup window, and chain-exhaustion guard with no additional code.

## Validation

- New unit tests: string form, `{ data: { statusCode: 400, message } }` form, `responseBody` JSON with the structured code, plus three negative cases proving generic flagged/policy/cybersecurity wording does not match.
- `bun test`: 2463 pass / 0 fail.
- `tsc --noEmit`: clean. `biome check`: clean.

## Compatibility

Additive-only change to error classification; no config, schema, or API changes. Existing behavior for all other error shapes is unchanged.

### DIFF
--- a/src/hooks/foreground-fallback/index.test.ts
+++ b/src/hooks/foreground-fallback/index.test.ts
@@ -468,6 +468,47 @@ describe('ForegroundFallbackManager session.error', () => {
     expect(call[0].body.model.modelID).toBe('gpt-4o');
   });
 
+  test('triggers fallback on content-policy moderation session.error', async () => {
+    // End-to-end regression: a cyber_policy rejection (HTTP 400
+    // invalid_request in production) must advance the fallback chain to the
+    // next model instead of failing the session outright.
+    await mgr.handleEvent({
+      type: 'message.updated',
+      properties: {
+        info: {
+          sessionID: 'sess-1',
+          providerID: 'anthropic',
+          modelID: 'claude-opus-4-5',
+          role: 'assistant',
+        },
+      },
+    });
+
+    await mgr.handleEvent({
+      type: 'session.error',
+      properties: {
+        sessionID: 'sess-1',
+        error: {
+          message:
+            'This content was flagged for possible cybersecurity risk. If this seems wrong, try rephrasing your request. To get authorized for security work, join the Trusted Access for Cyber program: https://chatgpt.com/cyber',
+        },
+      },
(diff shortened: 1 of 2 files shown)


dhaern:
@greptile-apps[bot] Added the missing session-level regression in 42cb1858: a `cyber_policy` moderation rejection delivered as `session.error` now has an end-to-end test proving the fallback chain advances to the next model (same `ForegroundFallbackManager` harness and assertions as the existing rate-limit case). Classifier-level tests are unchanged. `bun test`: 102 pass / 0 fail on the module; full suite green.

dhaern:
@greptile-apps review the changes please

Issue #1068, opened 2026-08-23
Harden wait_for_user against background tasks (call counting)


mhenke:
> *This was generated by AI during triage.*

## Background

#1000 reported the orchestrator looping on wait_for_user after spawning background tasks. That is now fixed on master: the orchestrator prompt has an "End Turn After Background Tasks" section and an explicit rule that background tasks are not external manual work and wait_for_user must not be used to await them.

The fix is prompt wording only. As noted in #1000's thread, a model can still invent wait_for_user as a wait mechanism. A belt-and-suspenders guard at the tool layer would catch that.

## Proposal

Count wait_for_user calls (or detect a wait_for_user call) while background tasks are outstanding for the session. When one occurs, either block it with a clear error or redirect the model to end the turn and let the wake scheduler resume.

## Key interfaces

- The wait_for_user tool definition and its permission layer
- Session state that tracks which background tasks are pending
- Optional config knob to enable or disable the guard

## Acceptance criteria

- [ ] A wait_for_user call while background tasks are pending is intercepted and does not block on manual input
- [ ] The model receives guidance to end the turn instead
- [ ] A unit test covers the guard firing and the pass-through when no tasks are pending
- [ ] bun run check:ci, bun run typecheck, bun test pass

## Out of scope

- Changing the orchestrator prompt wording (handled in #1000)
- General wait_for_user behavior for genuine external manual work

Issue #1000, opened 2026-08-11
[Bug]: Endless `wait_for_user` loop after spawning background tasks


stephanschielke:
### What happened, and what did you expect?

The orchestrator gets stuck in an endless loop of calling `wait_for_user` after spawning background tasks. Each call re-sends the full accumulated context, burning through token budgets rapidly. 

I have mainly observed this behavior with the latest Anthropic Claude models (Fable-5, Opus-5, Sonnet-5).

<img width="1269" height="810" alt="Image" src="https://github.com/user-attachments/assets/4e4c5978-10da-4860-9ef9-eff625e49f08" />

**Expected:** After spawning background tasks with `task(..., background: true)`, the orchestrator should end its turn. Hook-driven completion or the orchestrator wake scheduler resumes it automatically when the background task finishes.

**What actually happens:** The orchestrator calls `wait_for_user` as its "wait" mechanism for background task completion, then receives the tool result (which says "End this turn now"), but instead of ending the turn, calls `wait_for_user` again. This repeats until the model finally stops or the session is killed.

**Root cause:** The orchestrator prompt tells the agent to "wait for completion" after spawning background tasks, but does not specify HOW. The model invents `wait_for_user` as its wait mechanism, even though `wait_for_user` is designed exclusively for external human manual work (HITL), not background task completion. Additionally, the `wait_for_user` tool description does not explicitly exclude background task use cases.

**Evidence:** Session export shows the agent correctly ended its turn the first time after spawning tasks (`finish=stop`), then the orchestrator-wake scheduler re-engaged it, and the loop began.

Related to: https://github.com/alvinunreal/oh-my-opencode-slim/issues/924 (https://github.com/alvinunreal/oh-my-opencode-slim/pull/925)

### Steps to reproduce

1. Start an orchestrator session with background orchestration enabled
2. Spawn 2+ background tasks via `task(..., background: true)`
3. End the turn (agent does this correctly)
4. Wait for orchestrator-wake scheduler to re-engage (default 5 min idle)
5. Observe: agent calls `wait_for_user` to "wait for completion", enters loop

### oh-my-opencode.json

```jsonc
{
    "agents":
    {
        "orchestrator":
        {
            "model":
            [
                "omniroute-anthropic/claude/claude-fable-5"
            ]
        }
    },
    "provider":
    {
        "omniroute-anthropic":
        {
            "name": "Omni",
            "npm": "@ai-sdk/anthropic",
            "options":
            {
[...]

PR #1140, opened 2026-09-06
fix: block repeated wait tool calls within a turn

### LINKED ISSUE #1139: [Bug]: wait_for_user can loop indefinitely when orchestratorWake is disabled (exempt from tool-loop-guard)
> Following up on the invitation in #1000: *"counting wait_for_user calls … as a guard beyond prompt wording. Open that as a separate enhancement rather than reopening this one."*

**Version**: 2.2.17 installed; verified the relevant code is unchanged in the 2.2.18 npm tarball.
**Setup**: opencode with a non-Anthropic model (GLM via zai), `backgroundJobs.orchestratorWake.enabled: false`.

## What happened

A long-lived orchestrator session got stuck in a ~14-minute `wait_for_user` loop inside a single never-ending turn:

- **85 `wait_for_user` calls** in one turn, at 1–6 per minute
- Each call returned **instantly (~8 ms)** and triggered a full-context re-inference: **~285K input tokens per step** (~283K cache read)
- The loop only stopped when the user pressed Esc (cancel)
- The session held idle background PTY jobs (two `kubectl port-forward` + one finished CLI job), but — as shown below — the loop does not depend on them

## Root cause chain

**1.** With `orchestratorWake.enabled: false`, the #1068/#1077 guard can never fire, because its predicate ANDs the config flag:

```js
// src/tools/wait-for-user.ts (wiring, dist/server/index.js, v2.2.18)
hasOutstandingBackgroundTasks: (sessionID) =>
  runtime.backgroundJobs.orchestratorWake.enabled && backgroundJobCoordinator.hasRunning(sessionID)
```

So the tool always takes the normal path and returns instantly with:

```
state: waiting_for_user
protocol: oh-my-opencode-slim.wait_for_user.v1
End this turn now. Do not call more tools until the user responds.
```

**2.** The model does not end the turn. Each instant tool result re-triggers reasoning, which issues another `wait_for_user` call (in our logs the model literally reasoned "Idle wake again" between calls). A tool whose contract is "end your turn" still returns a tool result — so any model that continues after a tool result can loop here indefinitely.

**3.** Nothing at the harness layer stops it, because tool-loop-guard (#1074) exempts this exact tool:

```js
// src/hooks/tool-loop-guard/hook.ts (dist, v2.2.18)
var LOOP_GUARD_WARN_AT = 3;
var LOOP_GUARD_BLOCK_AT = 5;
var LOOP_GUARD_EXEMPT = {
  task: true, task_cancel: true, task_message: true, task_revive: true,
  wait_for_user: true, wait_for_background_tasks: true
};
var LOOP_GUARD_BLOCK_TOOLS = { read: true, grep: true, glob: true };
```

## Why the existing fixes don't cover it

- **#1000** was closed as fixed via orchestrator prompt wording ("wait_for_user must not be used to await background tasks"). In this incident the model was **not** waiting on background tasks — it called `wait_for_user` as its final action per the tool description, received the instant "End this turn" result, and kept looping anyway. Prompt wording cannot stop a model that re-loops on every tool result.
- The **#1068** guard only intercepts `wait_for_user` when `orchestratorWake.enabled && hasRunning()`. Users with wake disabled get no interception at all — the guard hole is config-gated, not model-gated.

## Suggested fix

- Count consecutive `wait_for_user` (and `wait_for_background_tasks`) calls per turn **regardless of orchestratorWake state or background-task presence**; hard-error after 2–3 with a "end your turn with text only, no tool calls" message.
- Alternatively/additionally: remove these tools from `LOOP_GUARD_EXEMPT`, or make the exemption apply only to the first call per turn.
- Extending `LOOP_GUARD_BLOCK_TOOLS` beyond `read`/`grep`/`glob` would harden the guard generally.

## Workaround we run locally

A local plugin hooking `tool.execute.before`: it counts `wait_for_user`/`wait_for_background_tasks` calls per session (counter reset on real user input via `chat.message`, or `session.idle`/`session.deleted`), tolerates 5, and throws a hard error from the 6th call instructing the model to end the turn with plain text. This reliably breaks the loop.

gitslim:
Fixes #1139

## Problem

`wait_for_user` / `wait_for_background_tasks` return instantly and were fully exempt from `tool-loop-guard`, so a model that keeps calling tools after a tool result could loop them indefinitely. In the incident behind #1139: **85 `wait_for_user` calls in one turn** at 1–6/min, each triggering a **~285K-token** full-context re-inference, ended only by the user pressing Esc. Neither the #1000 prompt wording nor the #1068 guard catches this (the latter is gated on `orchestratorWake.enabled && hasRunning()`).

## Change

`src/hooks/tool-loop-guard/hook.ts`:

- Removed `wait_for_user` / `wait_for_background_tasks` from `LOOP_GUARD_EXEMPT`.
- Added a dedicated wait stream: a per-session counter of **completed** wait-tool calls, keyed by tool name only (arguments and output may vary between loop iterations, so the fingerprint stream cannot catch it). Both tools share one stream, so alternating them is still recognized.
- Thresholds: `WAIT_GUARD_WARN_AT = 2` (corrective text appended to the 2nd call's output), `WAIT_GUARD_BLOCK_AT = 2` (the 3rd call is refused in `tool.execute.before` with an "end your turn, no more tools" error). A legitimate flow is exactly one wait per turn, so two completed waits + a warning is already generous.
- Counter resets on `observeNewUserMessage`, `resetTurn`, `resetSession`; bounded by the same `MAX_TRACKED_SESSIONS` eviction; wait-tool `after` hooks still honor the stale-after (no matching `before`) protection via `callKeys`.
- Wait-tool calls no longer touch the fingerprint stream, so read/grep/glob detection is unaffected.

## Tests

`src/hooks/tool-loop-guard/index.test.ts`: updated the exemption test (task tools remain exempt; wait tools no longer) and added 7 cases — first call passes, 2nd call with different args warns, 3rd refused, cross-tool shared stream, `observeNewUserMessage` reset, `resetTurn` reset, and no poisoning of the fingerprint stream.

## Verification

- `bun run check:ci` — clean
- `bun run typecheck` — clean
- `bun test src/hooks/tool-loop-guard/` — 30/30 pass
- Full `bun test` — 2471/2472; the single failure (`interview manager - dashboard election failure fallback`) reproduces on a clean `master` checkout and is unrelated.

## Docs

No README/docs changes: the loop guard has no user-facing documentation surface, and this changes no user-visible configuration (the hook has no config knob today).

### DIFF
--- a/src/hooks/tool-loop-guard/hook.ts
+++ b/src/hooks/tool-loop-guard/hook.ts
@@ -36,8 +36,14 @@ import { log } from '../../utils/logger';
  *   warn at N calls but stay warn-only (never hard-block) to avoid deadlocking
  *   terminal result retrieval for long-running background tasks.
  * - Task management and lifecycle tools (task, task_cancel, task_message,
- *   task_revive, wait_for_*) remain exempt; task-session-manager owns its
- *   own duplicate-spawn guards (#1056/#1070).
+ *   task_revive) remain exempt; task-session-manager owns its own
+ *   duplicate-spawn guards (#1056/#1070).
+ * - Wait tools (wait_for_user, wait_for_background_tasks) use a dedicated
+ *   per-turn counter keyed by tool name only: their contract is "end the
+ *   turn", so a repeat call within one turn is always degenerate regardless
+ *   of arguments or output. They warn at 2 completed calls and the 3rd call
+ *   is refused (#1139). The counter resets on a new user message, a
+ *   completed turn, or session deletion.
  *
  * Precedent: json-error-recovery (output warning) and task-session-manager
  * (before-hook refusal).
@@ -55,8 +61,6 @@ const LOOP_GUARD_EXEMPT: Record<string, true> = {
   task_cancel: true,
   task_message: true,
   task_revive: true,
-  wait_for_user: true,
-  wait_for_background_tasks: true,
 };
 
 /**
@@ -70,6 +74,28 @@ const LOOP_GUARD_BLOCK_TOOLS: Record<string, true> = {
   grep: true,
   glob: true,
 };
+
+/**
+ * Wait tools whose contract is "end this turn now". Repeating them within a
+ * turn is always degenerate: the first result already told the model to stop
+ * calling tools, and because the tool returns instantly, each repeat
+ * re-triggers a full-context inference — a non-compliant model can loop
+ * indefinitely (#1139). Counted by tool name only (arguments and output may
+ * vary); both tools share one per-session stream.
+ */
+const WAIT_TOOLS = new Set(['wait_for_user', 'wait_for_background_tasks']);
+const WAIT_GUARD_WARN_AT = 2;
+const WAIT_GUARD_BLOCK_AT = 2;
+
+export const WAIT_GUARD_MARKER = '[REPEATED WAIT TOOL - END TURN]';
+
+export const WAIT_GUARD_WARNING = `
+${WAIT_GUARD_MARKER}
+
+You have already called a wait tool ${WAIT_GUARD_WARN_AT} times in this turn. Its result instructed you to end the turn — do not call it again.
+
+STOP calling tools. Respond to the user in plain text and end your turn. The next distinct external user message resumes normal continuation.
+`;
 
 const LOOP_GUARD_MARKER = '[REPEATED TOOL CALLS - STOP]';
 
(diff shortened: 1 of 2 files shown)


gitslim:
Addressed the review finding in 4a10d8f: the wait-tool callKeys entry is now recorded only after the refusal check passes, so refused calls no longer leave entries that can never be consumed (there is no error hook to clean them up).

PR #1149, opened 2026-09-09
fix(perf): avoid repeated tui-state writes, image re-decoding and history hashing

dhaern:
## Problem

Three hot paths do redundant synchronous work inside the opencode process on every session:

1. **tui-state**: every `message.updated` with the same agent/model re-acquires the inter-process lock, re-reads and re-parses the snapshot, and re-serializes three sections just to discover nothing changed. 100 identical events cost ~7.3 ms of synchronous fs work plus 100 lock acquisitions. With two opencode instances on the same project, a no-op can also block up to 1 s in `Atomics.wait` while the other process holds the lock.
2. **image-hook**: every `messages.transform` re-runs the base64 regex, decodes and sha1-hashes every historical data-URL image, even when the content-addressed file already exists. Ten historical 2 MiB attachments cost ~62 ms on every later turn, including text-only ones.
3. **board-injection**: `injectLatestBoard` serializes and sha256-hashes the entire real history to build `promptShapeKey` before checking whether anything consumes it. With no terminal deliveries pending (the common case while subagents run), that hash is discarded: ~3.8 ms per orchestrator request on a 2 MiB history.

## Change

- `tui-state.ts`: memo of the last confirmed snapshot per project, keyed by file identity (ino, mtimeMs, size). Identical updates return before the lock with a single `statSync`. Only confirmed persistences seed the memo: a failed write or a transient read error (EACCES/EIO) leaves it untouched, so the mutation is retried on the next event instead of swallowed. ENOENT still opens the first-write path, and an external atomic replacement is caught by the inode change even when mtime and size are restored.
- `image-hook.ts`: memo of resolved data-URL attachments keyed by target dir + effective filename + sha256 of the URL. The raw payload is never used as a key (an earlier attempt at this retained megabytes of base64 in heap). Hits revalidate with `lstat` and refuse symlinks, matching `writeUniqueFile`'s contract. Suffixed collision names (`-N`) are not memoized, so the canonical path returns once the obstacle is gone.
- `board-injection.ts`: `promptShapeKey` is computed only when a consumed terminal job needs reconciliation or a new terminal delivery must be registered. First delivery still registers, repeated requests keep the delivery, new content reconciles before rendering. `injectCheckpointBoard` is untouched.

## Benchmarks

Same host and Bun 1.4.2, real exports, base `3c4d2547` vs this branch:

| Path | Before | After | |
|---|---:|---:|---:|
| 100 identical `recordTuiAgentModel` (median of 20 batches) | 7.289 ms | 0.956 ms | 7.6x |
| 1 historical 2 MiB image transform | 6.526 ms | 2.862 ms | 2.3x |
| 10 historical 2 MiB images transform | 62.408 ms | 23.175 ms | 2.7x |
| board inject, 2 MiB history, no jobs | 3.835 ms | 0.080 ms | 48x |

Memory (instrumented, 32 distinct 2 MiB attachments, after GC): heap retention from memo keys dropped from ~88 MiB to ~3 MiB.

In practice: sessions without model/agent changes stop touching the lock per event; long conversations with pasted images stop paying per-turn reprocessing and stop retaining payload copies; orchestrator sessions stop hashing the whole history when there is nothing to reconcile. These are milliseconds per call, so the gain shows up as less accumulated jank on long sessions rather than a visible speedup on a short one.

## Tests

12 new regressions covering the sequences caught during review: no-op without lock/fs access, real change after a no-op, retry after a failed rename, memo invalidation on external atomic replacement with identical mtime+size, transient read failure not seeding an empty snapshot, memo reuse without `Buffer.from`, re-save after deletion, symlink refusal on hits, filename-keyed separation, suffixed resolutions not memoized.

## Verification

- `bun test`: 2509 pass, 1 fail pre-existing on master (`runtime.test.ts:134`, introduced by #1130, unrelated to this change)
- `bun run typecheck`: clean
- `biome check` on the five touched files: clean
- Read-only review of all integration points: tui-state consumers, `cleanupAllSessions`/`writeUniqueFile` interaction, checkpoint board path

### DIFF
--- a/src/hooks/image-hook.test.ts
+++ b/src/hooks/image-hook.test.ts
@@ -1,4 +1,4 @@
-import { afterAll, describe, expect, it } from 'bun:test';
+import { afterAll, describe, expect, it, spyOn } from 'bun:test';
 import { spawnSync } from 'node:child_process';
 import { createHash } from 'node:crypto';
 import {
@@ -71,6 +71,20 @@ function makeUserMsg(parts: MessageWithParts['parts']): MessageWithParts {
 
 function imagePartCount(message: MessageWithParts): number {
   return message.parts.filter((part) => part.type === 'image').length;
+}
+
+const AUTO = {
+  imageRouting: 'auto' as const,
+  disabledAgents: new Set<string>(),
+  log: () => {},
+};
+
+function processAuto(messages: MessageWithParts[], workDir: string): boolean {
+  return processImageAttachments({ messages, workDir, ...AUTO });
+}
+
+function nudgeText(message: MessageWithParts): string {
+  return message.parts.find((part) => part.type === 'text')?.text ?? '';
 }
 
 afterAll(() => {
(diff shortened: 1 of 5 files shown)


dhaern:
@greptile-apps good catch — you were right about the gap.

A metadata-preserving in-place write (same inode, same size, mtime restored via `utimes`) could indeed leave the TUI memo trusting stale content: our identity check only compared ino/mtime/size, and `utimes` can restore two of the three.

Fixed in 547f047e: `ctimeMs` is now part of the file identity. On POSIX an in-place write always bumps ctime, and userspace cannot restore it (calling `utimes` to fix mtime bumps ctime again), so the memo now invalidates on both in-place rewrites and atomic replacements (different inode) even when every other field matches.

Added a regression that reproduces the exact scenario you described: same inode, same length, mtime pinned and restored exactly, only ctime differs — verified red without the fix and green with it. The no-op fast path still costs a single `statSync`; nothing else changed.

dhaern:
@greptile-apps Update the score please

PR #1152, opened 2026-09-10
feat: preset inheritance + develop merge

ZenStudioLab:
## What changed

This branch adds single-parent preset inheritance behind an optional `$extends` field on each preset. A preset without `$extends` works as before. A preset can set `$extends` to one preset name, or to `null` to detach a parent that came from a lower-priority config layer.

Config without the marker keeps current behavior. The marker is metadata only. It never shows up as an agent, a summary row, a count, an alias, or a runtime override.

## How inheritance resolves

Resolution runs after the user and project layers merge, so the merged graph is what counts. For the selected preset the order is parent, then child, then root `agents`. Nested objects merge. Arrays and plain values replace.

A higher layer that omits `$extends` keeps the lower-layer parent. A higher-layer string replaces it. A higher-layer `null` removes it. There is no other field deletion in v1.

Inline `prompt` and `orchestratorPrompt` fields inherit through the merge. Parent prompt directories do not. Prompt file lookup stays scoped to the active preset and the existing project and user roots.

Parent-only presets stay selectable. Selecting one applies its own config plus its ancestors.

## Validation and errors

Every preset in the merged set is checked before selection, including presets that are not active. A missing parent or a cycle stops plugin startup and names the preset plus the bad reference or the cycle path.

Empty or whitespace-only `$extends` stays a schema error and follows the existing warning and fallback path. Malformed JSON keeps its existing handling. Graph errors are separate and fatal by design.

## Surfaces touched

Startup and reload selection use resolved values. Disk switching (`switchPresetOnDisk` in `src/tools/preset-switch.ts`) reads the resolved entry for agent updates and summaries and keeps the raw marker on write. TUI edit, save, overwrite, and delete preserve `$extends`, validate the merged candidate graph before any write, and refuse writes that would add a missing parent or a cycle. Deleting a preset that a child still names is refused. Doctor (`src/cli/doctor.ts`) checks files per file as before, then checks the merged graph and returns a failing result on graph errors. The generated JSON schema now describes `$extends` and is regenerated from the Zod schema instead of hand edited.

Out of scope: multiple parents, mixins, general field deletion, parent prompt directories, and dormant live runtime preset switching.

## Example

A small config shows the shape:

```json
{
  "presets": {
    "base": { "agents": { "reviewer": { "model": "openai/gpt-5-mini" } } },
    "child": { "$extends": "base", "agents": { "reviewer": { "temperature": 0.2 } } }
  }
}
```

`child` gets the base reviewer config with temperature replaced. If the user layer sets `child.$extends` to `base` and the project layer sets it to `null`, the merged `child` has no parent.

## Verification

I rebased this onto upstream/master as one bundle: 2 commits, 25 files, about 3327 added and 132 removed. `bun run typecheck` passes. `bun run build` passes. Feature tests pass (49). Full `bun test` shows 2498 passing with unrelated environment and baseline failures. `bun run check:ci` flags only `src/hooks/foreground-fallback/index.ts`, which sits outside this diff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,6 @@ companion/target/
 .slim/deepwork/
 .codegraph/
 .worktrees/
+
+# openspec
+openspec/
--- a/.gitignore
+++ b/.gitignore
@@ -105,4 +105,5 @@ companion/target/
 .worktrees/
 
 # openspec
-openspec/
+openspec/.worktrees/
+.worktrees/
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ The main idea is simple: instead of forcing one model to do everything, the plug
   showing which agents are active, including parallel background specialists.
 - **[Multiplexer integration](docs/multiplexer-integration.md)** - watch agents
   work live in Tmux, Zellij, Herdr, cmux, or kitty panes.
-- **[Preset switching](docs/preset-switching.md)** - swap the whole team's
-  models at runtime with `/preset`.
+- **[Preset switching](docs/preset-switching.md)** - select and persist a preset
+  with `/preset`; presets support single-parent inheritance via `$extends`.
 - **[Code intelligence tools](docs/tools.md)** - LSP tools, AST-aware search
   across 25 languages, and built-in MCPs for docs and GitHub code
   search.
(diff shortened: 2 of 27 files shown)


alvinunreal:
its godd idea

alvinunreal:
@orchestrat0r what you think?

alvinunreal:
feature is added

PR #1166, opened 2026-09-12
fix(task-session-manager): drain unresolvable no-ID takes via guarded first-match fallback

GoldJohnKing:
## Summary

Follow-up to #1162. The no-tool-call-ID identity resolution added there has a liveness gap in its own target scenario: when a **parallel no-ID burst produces no early claims** (no titles → only placeholder registrations), both identity sources miss and the after-hook hits `if (!pending) return;` *before* the try/finally — the call's output is dropped, the pending is stranded, its concurrency ticket never releases, and because `solePendingIdForParent` counts *all* pendings for the parent, **every subsequent no-ID take from that parent refuses forever** (parent poisoned until session deletion / FIFO eviction). Master's pre-#1162 behavior drained the map (at the cost of label guessing); #1162 made the failure non-draining.

## Fix

When both `take()` and `takeByTaskID()` miss but the call's **own output parsed a task ID** (authoritative identity), consume the oldest *eligible unmarked* pending via a new guarded fallback `takeUnresolvedFirstMatch(parent, { identityTaskID, agentType, ownerBoard })` with a deterministic warning log, then flow through the unchanged try/finally (ticket release + output processing as usual).

Selection skips: marked pendings (`earlyRegisteredTaskID` / `earlyRegistrationRejected` — each has its own owner-resolution path), resumed pendings pinned to a different task ID, agent mismatches (agent hint derived from the board record only when it belongs to this parent), and other board generations. An **unknown call ID** still early-returns — a first-match guess must never steal a live call's pending. Unknown-ID + no-parsed-ID cases keep the current early return (placeholder already covers the board).

Result: the failure mode for genuinely unresolvable identity regresses exactly to bounded label mis-attribution (contained by the authoritative-output registration), never a liveness defect.

## Verification

- [x] `bun run check:ci` / `bun run typecheck` — clean
- [x] `bun test` — 2699 pass / 0 fail
- [x] New tests: the B1 burst end-to-end (drain asserted via `peekByParent`, ticket release via `BackgroundTaskConcurrency.snapshot()` 2→1→0, authoritative task ID + placeholder label correction, subsequent no-ID take succeeds); exactly-one-warning subprocess test; plus contract tests for the fallback selection rules (oldest-unmarked, agent-constrained, marked-never-consumed, pin-skip both directions, empty parent)
- [x] All four #1162-validated behaviors preserved (unique-claim peek, no early-registration drop, cross-mark fix, placeholder preserveRun) — walked in review

Reviewed with REQUEST-CHANGES → fix → APPROVE by an independent review pass over the interleaving matrix.


GoldJohnKing:
Closing in favor of #1167, which contains this commit plus follow-up cleanups (maintainer preference: single combined PR). No branch deletion needed — chore/tsm-micro-cleanup is stacked on this branch head.

PR #1168, opened 2026-09-12
fix(v2): restore /preset, stop stray orchestrator wakes, and drop beta-era paths

mxalbert1996:

OpenCode v2 graduated from preview builds (`0.0.0-beta-xxxxx`) to stable releases. This PR aligns the plugin with the stable v2 host surface and fixes three user-visible problems found while running against it.

- **`/preset` missing on v2** — the TUI entry registered the preset layer from plugin setup, but the host's `keymap.layer` is provider-scoped; the call threw and was silently swallowed. It now registers from a slot render (under the provider) with the host's command schema and toast API.

- **Stray wakes after conversations ended, resetting reasoning effort** — stable hosts expose no `session.list`, so the wake watchdog's event-tracked children carried no terminal outcome: finished children kept counting as "active" and woke idle parents. Every wake also reset the session model's variant because the model pin omitted it. Fallback children are now refreshed from `session.get` each evaluation, and the variant travels with the pin.

- **Beta-era leftovers** — removed the pre-stable `session.status` → `session.idle` synthesis path (dead on stable hosts) and scrubbed beta/patch-version references from the docs.

## Verification

Live-verified in sandboxes against a stable v2 host; `bun test` (2703 pass), typecheck, lint, and build all green. v1 behavior unchanged.

Also verified manually against OpenCode 2.0.2.


alvinunreal:
thanks @mxalbert1996

PR #1176, opened 2026-09-13
fix(orchestrator-wake): include stopped-job facts in recovery wake

### LINKED ISSUE #1051: Stopped-job recovery wake references a board snapshot that internal-initiator messages cannot create
When a background job is marked `stopped`, the recovery path triggers an internal orchestrator wake whose prompt directs the model to consult the Background Job Board.

Internal-initiator messages are excluded from producing a current board snapshot. When the recovery wake is delivered, only previously retained snapshots are present, and none reflect the stopped job that triggered the wake.

Under `strategy: "checkpoint-compatible"` this presents as a stale board at the point of recovery.

Requested: emit a current snapshot as part of the recovery wake, or defer the wake until a snapshot reflecting the stopped job is present.

Possibly related: #1031, #1044.

Version: 2.2.15

dhaern:
Fixes #1051

## What changed and why

The stopped-job recovery wake is an internal-initiator message. Under the `checkpoint-compatible` board strategy those messages cannot create a snapshot, so the first board the parent sees after the wake still reflects the job as running. A reminder that only says "consult the board" therefore points at a stale picture.

This appends a self-contained delta to the wake: alias, task id, generation, state, and why it stopped. Stops that arrive while a recovery wake is in flight stay queued; only the keys that were actually sent are retired after delivery. A failed delivery leaves the batch in place. Duplicates of the same `(taskID, generation)` collapse.

The periodic TODO wake is unchanged. Empty recovery entries are dropped so a delivered stop does not re-fire on the next idle.

## Verification

- `bun test src/hooks/orchestrator-wake/index.test.ts`: 77 pass, 0 fail (covers inline delta, in-flight second stop, dedup, restore-on-fail, no re-fire)
- full `bun test`: 2710 pass, 0 fail
- `tsc --noEmit`: clean
- `bunx biome check` on the four changed files: clean


dhaern:
@greptile-apps the unbounded recovery-batch concern was valid. A parent that stayed busy or waiting for input could keep every distinct stop in one join, and a failed oversized delivery retried the same batch.

`e23e8ac6` caps the per-parent queue at 32 (FIFO, oldest dropped) and sends at most 8 deltas per wake. Remaining entries stay queued and drain on later idle events, so a failed join cannot wedge the batch. Please update the review.

PR #1177, opened 2026-09-13
fix(council): except host compaction templates from report format

### LINKED ISSUE #1154: council prompt's mandatory output format breaks OpenCode V2 compaction template validation
# council prompt's mandatory output format breaks OpenCode V2 compaction template validation

## Summary

When the `council` agent is the active session agent, OpenCode V2 compaction fails with:

```
Compaction summary did not match the required template
```

OpenCode V2 validates the compaction summary by requiring at least one heading line that matches its checkpoint template exactly (`## Objective`, `## Requirements`, `## Decisions`, `## Work State`, `## Next Move`, `## Relevant Files`, `## Important Context`). It retries once, then fails. The council agent prompt mandates a different output format, so both attempts fail.

## Environment

- oh-my-opencode-slim `2.2.18`
- OpenCode V2 `0.0.0-beta-19425`
- Agent `council`, model `openai/gpt-6-astra` (variant xhigh)

## Reproduction

1. Have a session with a long history (near the compaction threshold), or just use manual compaction on any history.
2. Make sure the session agent is `council` (e.g. switch the top-level agent to it).
3. Trigger compaction:
   - automatic: send any message, or
   - manual: `POST /api/session/{sessionID}/compact`.
4. Compaction runs, retries once, then fails with `compaction.failed` and the message above.

## Observed behavior

Same session, three compaction attempts:

| Attempt | Reason | Result |
|---|---|---|
| 1 | auto | failed, `Compaction summary did not match the required template` |
| 2 | manual | failed, same error |
| 3 | auto, after switching session agent back to `orchestrator` | completed, summary starts with `## Objective` |

Additional evidence: assistant messages produced while `council` was the session agent contain `## Council Response` / `## Council Summary`, confirming the mandated format.

## Root cause

- `COUNCIL_AGENT_PROMPT` in `src/agents/council.ts` says *"Always include these sections in your final response"* and requires `## Council Response`, `## Per-Councillor Details`, and `## Council Summary`.
- `COUNCIL_SYNTHESIS_REINFORCEMENT` is appended after any user prompt override and reiterates the requirement: *"You MUST ... produce the required output with a synthesized Council Response..."*.
- OpenCode V2 compaction reuses the session's selected agent (prompt, model, variant) and appends a user message requesting its own checkpoint template. The summary is accepted only if it contains an exact heading from the template; one corrective retry is made, then the request fails.
- The council format wins, so both the first response and the retry fail validation.

## Suggested fixes

1. Add a checkpoint/compaction exception to the council prompt (and/or its reinforcement) so a compaction request follows OpenCode's requested template.
2. Alternatively, detect hidden checkpoint/compaction requests in the V2 bridge (`ctx.session.hook("context")` → `experimental.chat.system.transform`) and suppress the council format mandate for those requests.
3. At minimum, document that `council` should not be the active session agent while compaction can trigger.

## Related

- #1059 — `experimental.chat.system.transform injects the orchestrator prompt into hidden title/compaction calls`. Same class of conflict: plugin prompt injection vs. OpenCode's internal requests.

## Local workaround (verified)

Adding a checkpoint/compaction exception via `~/.config/opencode/oh-my-opencode-slim/council_append.md` (documented prompt override mechanism) makes compaction complete with the expected `## Objective` summary under `council`, while normal council synthesis still produces `## Council Response` / `## Council Summary`. A bundled fix would still be preferable.

---

**AI disclosure:** generated with OpenCode + oh-my-opencode-slim.

dhaern:
Fixes #1154

## What changed and why

When council is the session agent, OpenCode V2 compaction fails because the council prompt requires `## Council Response` / `## Council Summary` and the host only accepts its own checkpoint headings.

The council report format stays for normal synthesis. If the host asks for a checkpoint or compaction summary in a specific template, follow that template instead. The exception lives in the default prompt and in the post-override reinforcement, and `createAgents` re-applies it after resolving inline/file/append prompts so a custom `agents.council.prompt` cannot drop it. The helper is idempotent and does not impose the synthesis reinforcement on overrides. It does not hardcode a particular host heading set.

## Verification

- `bun test src/agents/council.test.ts src/agents/index.test.ts`: 127 pass, 0 fail
- `bun test src/agents/`: 198 pass, 0 fail
- `tsc --noEmit`: clean
- `bunx biome check` on the four changed files: clean


dhaern:
@greptile-apps the host-prompt override was real. `agent.council.prompt` wins the shallow merge in `config()` and dropped the exception applied in `createAgents`.

`7c1f1c5f` re-applies it after that merge on both the canonical `council` key and the visible display-name alias. Please update the review.

PR #1180, opened 2026-09-13
feat(config): per-agent skills_add/skills_remove directives

Aveer:
## Summary

Adds optional per-agent `skills_add` / `skills_remove` configuration directives so project-local config can add or remove skills on top of an inherited list without duplicating the global `skills` array. General deep-merge array replacement semantics are unchanged (arrays elsewhere — models, disabled agents, MCPs, etc. — remain replacement-based).

## Semantics

The directives are configuration inputs, not SDK agent properties. They are folded into the effective `skills` token list during config resolution — in `loadPluginConfig()` after user + project + env-preset + preset-merge (root precedence) resolution, and at the tail of `RuntimeConfig.agents()` (covering raw seeds, file presets, and runtime `/preset` switching) — so downstream consumers (agent definitions, skill permission generation, the available-skills filter hook) only ever see a plain effective `skills` list. The existing resolver (`getSkillPermissionsForAgent`) continues to interpret the tokens; `*` is never expanded during merging and no second wildcard layer is introduced.

1. Resolve the inherited/configured `skills` list as before (user config, root `agents`, presets, project-local overrides, startup-selected and runtime `/preset` switching).
2. Apply `skills_add` (union, first-occurrence dedupe).
3. Apply `skills_remove`, operating on final skill tokens: a plain name removes that name token, and a `"!name"` entry removes the exclusion token itself (lifting an existing exclusion). Removal wins over addition for the same skill.
4. When the effective list contains `"*"`, a plain-name removal is granted implicitly by the wildcard, so it is expressed by appending the existing `"!name"` exclusion token — unless that exclusion is already present (no duplicate tokens).

Worked example:

```jsonc
// ~/.config/opencode/oh-my-opencode-slim.jsonc (global)
{ "agents": { "oracle": { "skills": ["codemap", "deepwork", "verification"] } } }

// <project>/.opencode/oh-my-opencode-slim.jsonc (project-local)
{ "agents": { "oracle": { "skills_add": ["nexus-backend"], "skills_remove": ["verification"] } } }

// effective
["codemap", "deepwork", "nexus-backend"]
```

Token edge case (no `*` expansion):

```jsonc
{ "skills": ["*", "!legacy"], "skills_add": ["project-skill"], "skills_remove": ["!legacy"] }
// effective: ["*", "project-skill"]  — the exclusion token is removed, wildcard kept as-is
```

Existing `skills` behavior is preserved exactly (explicit lists, `["*"]`, `["*", "!item"]`).

**Layering note:** the directives are ordinary config fields and follow the existing field-level layering of preset/root agent config (the same contract `temperature`, `mcps`, etc. already follow): a lower layer's `skills_add`/`skills_remove` survives a higher layer that does not override that directive field. To suppress a lower layer's additions, declare `skills_add: []` in the higher layer (array replacement).

## Changes

- `src/config/schema.ts` — optional `skills_add` / `skills_remove` (`string[]`) on `AgentOverrideConfigSchema` (presets inherit the same schema); `oh-my-opencode-slim.schema.json` regenerated
- `src/cli/skills.ts` — `resolveEffectiveSkills()` + `getDefaultGrantedSkillNames()` (no behavior change to `getSkillPermissionsForAgent`)
- `src/config/utils.ts` — `normalizeAgentSkillDirectives()` (folds directives into `skills`, strips them)
- `src/config/loader.ts`, `src/config/runtime.ts` — normalization at config resolution and at runtime agent resolution
- `src/config/skills-add-remove.test.ts` — 33 new tests
- Docs: `docs/configuration.md` (option reference), `docs/project-local-customization.md` (new section with the primary use-case example), `docs/skills.md` (new "Adding or removing skills on top of an inherited list" section)
- `chore(tsm): biome-format same-provider test files` — separate commit formatting `src/hooks/task-session-manager/same-provider-*.test.ts` as merged into master (unformatted in #1179); required for the repo's `check:ci` gate to pass on any PR

## Test coverage

Resolver unit cases (remove-wins-over-add, dedupe, wildcard token preservation, exclusion-token removal/lifting, duplicate-removal dedupe, no-existing-`skills` add-only / remove-only), loader end-to-end across user + project layers (add, remove, add+remove with duplicates, custom agents, preset-defined agents, root-agent precedence, `["*"]` + removal, backward compatibility), invalid values rejected by `PluginConfigSchema.safeParse`, `RuntimeConfig` seed + `setRuntimePreset` switching, and `createAgents` permission output.

## Verification

- `bun run typecheck` — pass
- Biome check on all touched files — clean
- Targeted config/cli/agent suites — 197 pass / 0 fail
- Full suite on the pre-review base: 2704 pass / 156 fail with this change vs 2676 pass / 156 fail on the same base commit; failure set name-identical (pre-existing/environmental on Windows)


Aveer:
Addressed the review findings:

- **Wildcard exclusion duplication / token semantics** (fixed): `skills_remove` now operates on final skill tokens - a plain name removes that token, and a `!name` entry removes the exclusion token itself (lifting the exclusion). `*` is never expanded; the existing resolver keeps interpreting tokens. `[*,!legacy]` + add `[project-skill]` - remove `[!legacy]` now normalizes to `[*,project-skill]`, and plain-name removal under `*` appends `!name` only when not already present (no duplicate tokens). Regression tests added.
- **Runtime preset retaining lower-layer directives**: intentional field-level layering - consistent with how every other agent field (temperature, mcps, ...) layers across presets, and with the requirement that directives apply after the effective `skills` value is fully resolved across all layers. A higher layer can explicitly reset with `skills_add: []` (array replacement). Documented in the PR description.

Also rebased onto current master and added a separate `chore(tsm)` commit that biome-formats `src/hooks/task-session-manager/same-provider-*.test.ts` as merged in #1179 - the required `check:ci` gate fails on master itself without it, which is what failed the first CI run of this PR.

Issue #1147, opened 2026-09-09
[Bug]: TUI subagent activity spinners bleed across multiple opencode windows sharing one project dir


zhangjinglei137:
### What happened, and what did you expect?

When two (or more) **independent opencode TUI windows** are opened on the **same project directory** at the same time and each window runs its own subagents, **every window's sidebar shows the running-activity spinners of all subagents from all windows**, not just the subagents belonging to that window.

Concrete scenario:

- Window A (`opencode1`) runs an `oracle` subagent via the `task` tool.
- Window B (`opencode2`) runs a `fixer` subagent via the `task` tool.
- Result: **both** Window A and Window B show **both** `oracle` and `fixer` animating in the sidebar.

Expected: each opencode window's TUI sidebar should only animate the subagents that belong to that window's own session tree. Activity from other windows/processes should never appear, so the user can tell at a glance which subagent belongs to the window they are looking at.

This is distinct from #1132 (parent-row cross-attribution + indicators never stopping within a *single* window) and from #1133 (remote-attach `tui-state.json` local-file-only). This report is specifically about **cross-window / cross-process activity bleed on the same project**.

### Steps to reproduce

1. `cd` into a project dir and open opencode TUI as Window A.
2. In a second terminal, `cd` into the same project dir and open another opencode TUI as Window B.
3. In Window A, dispatch a background `oracle` subagent (e.g. via the `task` tool, background mode).
4. In Window B, dispatch a background `fixer` subagent.
5. Watch both sidebars: each window displays running spinners for **both** `oracle` and `fixer`, regardless of which window actually spawned them.

### Root-cause analysis (from `src/tui-state.ts`)

The activity state is **not per-window/per-instance**:

- `activeSessions` is a flat global map `{ sessionID: agentName }` (see `recordTuiAgentActivity`).
- The state file lives at `~/.local/share/opencode/storage/oh-my-opencode-slim/{projectScopeHash}/tui-state.json`, where the scope hash is derived **only from the project directory** (`projectScope(projectDir)`).
- Two opencode instances on the same project therefore share one `tui-state.json`, and the TUI sidebar renders the union of all recorded active sessions — there is no notion of "which opencode instance / window owns this session".

[...]

PR #1183, opened 2026-09-13
fix(tui): scope sidebar spinners to the visible conversation

dhaern:
## Problem

Two independent opencode windows on the same project directory animate each other's subagents in the sidebar (#1147). The activity map in `tui-state.json` is global per project, so every window renders the union of all recorded activities.

Process identity cannot scope this: v1 windows are separate processes, but v2 windows share one detached daemon that records activities for all of them from a single process.

## Change

Scope by session tree instead:

- **Persistent `sessionParents` index** in the snapshot. `session.created` records each child→parent link via `recordTuiSessionParent`, surviving restarts and revives.
- **Host hydration**: when a session becomes active with no known link (preexisting sessions, missed events), ask the host via `session.get` and walk up to a confirmed root. SDK receiver preserved (`.call`), cycle-guarded, errored lookups released for retry, malformed responses never cached as "confirmed root".
- **Sidebar filtering**: `getActiveSidebarAgentNames` resolves the visible route session and every active session against the same index. Navigating into a child keeps its spinner; the home route keeps the union.
- **Startup sweep** keeps live activities recorded by other windows and drops crash residue (dead recorder pid) and legacy entries.

## In practice

Window A on conversation 1 sees only its own subagents spinning; window B on conversation 2 sees only its own. Activity between windows no longer bleeds.

## Tests

8 regressions: conversation scoping, child-route navigation, multi-level ancestry, late-discovered ancestry, startup sweep with live foreign activities, hydration chain with receiver verification, errored-lookup retry, plus the updated activity suites.

## Verification

- `bun test`: 2840 pass / 0 fail
- `tsc --noEmit`: clean
- `biome check`: clean

Refs #1147


dhaern:
@greptile-apps Both P2s are addressed in f64d6182:

1. **Hidden activity drives rerenders** — both animation timers (v2 `setup` and the v1 contract) now pass the visible route session to `getActiveSidebarAgentNames`, matching the render scoping, so foreign-conversation activity no longer rerenders this window's sidebar each 100ms frame.

2. **Malformed parents block retries** — a non-string `parentID` now releases the `hydratedTuiParents` slot instead of caching a false confirmed root; `null`/absent still count as a legitimate root. Added a regression test (`does not cache a malformed parentID as a confirmed root`).

Suite 2841 pass / 0 fail. Please update the review.

Issue #1133, opened 2026-09-04
[Bug]: Sidebar shows "pending" for all agents when TUI attaches to a remote opencode serve (tui-state.json is local-file-only)


haotianfei:
### What happened, and what did you expect?

When the TUI runs on machine A and attaches to `opencode serve` on machine B, the OMO-Slim sidebar permanently shows the fallback agent list with `pending` instead of resolved models (screenshot attached below). Expected: sidebar shows each agent's configured model, as it does when the TUI and server share a filesystem.

Root cause (from source analysis):
- The server-side plugin writes agent models to the *server's* local file: `$XDG_DATA_HOME/opencode/storage/oh-my-opencode-slim/<sha256(resolve(projectDir))[:12]>/tui-state.json` (`src/tui-state.ts`, called from the config hook via `recordTuiAgentModels`).
- The TUI plugin polls that path on the *client's* local filesystem every 1s and renders `snapshot.agentModels[agentName] ?? 'pending'` (`src/tui.ts`).
- In remote-attach mode the two processes don't share a filesystem, so the client always reads an empty snapshot → fallback agent list (`FALLBACK_SIDEBAR_AGENTS`, note the orchestrator row disappears) with every model column showing `pending`.
- Notably the project-scope hash matches on both ends (in attach mode `api.state.path.directory` is the server-reported path string), so the only gap is file visibility, not keying.

### Steps to reproduce

1. On machine B: `opencode serve --hostname 0.0.0.0 --port 4096` inside a project directory
2. On machine A: `opencode attach http://B:4096` and open any session in that project
3. Observe the OMO-Slim sidebar: explorer/librarian/oracle/designer/fixer all show `pending`
4. On machine B itself, the same project's sidebar correctly shows model names (state file exists in B's `$XDG_DATA_HOME`)

### oh-my-opencode.json

```jsonc
// active preset: alibaba-token-plan-cn — orchestrator/explorer/librarian/designer/fixer -> alibaba-token-plan-cn/qwen3.8-flash; oracle -> alibaba-token-plan-cn/deepseek-v4-pro; variants: orchestrator=xhigh, explorer=low, librarian=high, oracle=high, fixer=high
```

### OpenCode version

1.18.18（server 与 TUI 客户端同版本）

### oh-my-opencode-slim version

2.2.18

### Operating system

Fedora 44 Linux (both server and client)

### Logs, screenshots, or extra context

```shell
Proposed fix options:
[...]

PR #1184, opened 2026-09-13
fix(tui): resolve sidebar models over the host SDK on remote attach

dhaern:
## Problem

When the TUI attaches to a remote `opencode serve`, the OMO-Slim sidebar shows every agent as `pending`. The server plugin writes `tui-state.json` on the server filesystem; the TUI polls that path on the client. The two machines do not share a filesystem, so the snapshot is always empty.

## Change

If the local snapshot has no `agentModels`, ask the host for the agent list over the TUI client and fill the gaps. Local snapshot entries still win.

- v1 (`api.client`, `@opencode-ai/sdk/v2`): `app.agents({ directory })`, `{ name, model: { providerID, modelID } }`
- v2: `agent.list({ location: { directory } })` or `v2.agent.list(...)`, `{ id, model: { providerID, id } }`
- Cache: hits kept; empty misses retry after 5s
- First tick hydrates immediately instead of waiting for the 1s poll

## Tests

v1 argument `{ directory }`, v2 `{ location: { directory } }`, nested `v2.agent.list`, local-wins merge, no-op without a client.

## Verification

- `bun test src/tui.test.ts`: 32 pass / 0 fail
- `bun test`: 2858 pass / 0 fail
- `tsc --noEmit`: clean
- `biome check`: clean

Refs #1133


dhaern:
@greptile-apps please update the review.

P1-1 (partial snapshots skip hydration): the empty-snapshot guard is intentional. `getSidebarAgentNames` uses `Object.keys(agentModels)` as the visible agent list, so a non-empty snapshot is already the configured set (including disabled agents omitted). Filling gaps from the host list would reintroduce those agents. Local snapshot entries still win via `applyRemoteAgentModels`; the remote path only runs when the client cannot see `tui-state.json` at all.

P1-2 (overlapping refreshes): fixed in 778a2872. `createSerializedRefresh` skips a tick while a fetch is in flight, and `isRefreshCurrent` drops a result whose directory changed while waiting. Covered by the serialized-refresh test.

Issue #1143, opened 2026-09-07
Agents sometimes hardcode wrong absolute paths -> NotFound: FileSystem.access on Bash/Read tools


jwcrystal:
## Summary

When an oh-my-opencode-slim agent (e.g. @orchestrator) dispatches a Bash/Shell or Read tool, it occasionally hardcodes an **incorrect absolute path** (a path that doesn't exist). OpenCode then calls `fs.access` on that path before running the tool and Bun throws:

```
NotFound: FileSystem.access (/Users/<user>/Documents/Workspace/<ProjectName>)
```

where the real cwd is e.g. `/Users/<user>/Documents/Workspace/<ParentDir>/<ChildDir>/<ProjectName>` (a `<ParentDir>/<ChildDir>/` segment was dropped by the model).

## Example

- Tool: `npx vitest run --config vitest.contract.config.ts apps/api/src/pilot/pilot.contract.test.ts`
- Error: `NotFound: FileSystem.access (/Users/<user>/Documents/Workspace/<ProjectName>)` — wrong path, missing a directory segment
- A parallel command using **relative paths** (`pwd && ls apps/api/src/pilot`) succeeded fine

So this is the model guessing a wrong absolute path rather than a plugin/permission problem.

## Related but different

- #276 was about `apply_patch` absolute-path blocking and is fixed. This issue is broader: it hits **any** file tool (Bash/Shell, Read, Glob) when the model writes a wrong absolute path, and it fails hard (`NotFound`) instead of being rescued.

## Mitigation already made locally

Added a rule to the @orchestrator prompt (in `buildOrchestratorPrompt`'s "Dispatch efficiency" section):

> Use relative paths (e.g. `src/app.ts`), never hardcode absolute paths; when unsure of the working directory, run `pwd` first

This is a prompt-level mitigation (probabilistic). It does not hard-guarantee the model will comply.

## Question / suggestion

Would it make sense to add a **hard defense** at the tool/hook layer — e.g. detect tool args containing an absolute path that does not exist and is under the workspace, then normalize to workspace-relative (or rewrite the error to hint `pwd`)? That would be a deterministic fix for this class of `NotFound: FileSystem.access` failures.

Happy to open a PR if the maintainers think the hook-layer defense is worth adding.

dhaern:
PR up: #1186.

[...]

PR #1186, opened 2026-09-14
fix(tools): rescue guessed absolute paths with dropped directory segments

dhaern:
## Problem

Agents occasionally guess an absolute path with dropped parent segments (e.g. `/home/u/Work/Project` instead of `/home/u/Work/Parent/Child/Project`). OpenCode calls `fs.access` on the path before running the tool and Bun throws `NotFound: FileSystem.access (<path>)`, failing the call before the tool even runs (#1143). A parallel command with relative paths succeeds fine, so this is purely a guessed-path problem.

## Change

New `absolute-path-rescue` hook on `tool.execute.before` for `read`/`list`/`glob`/`grep`, with a deliberately narrow contract:

- Only absolute `filePath`/`path` arguments whose absence is a plain ENOENT (permission/I/O errors pass through untouched).
- Only guesses NOT already anchored under the workspace — a missing path inside the workspace is a legitimate absence and must surface, never be silently redirected to a different file.
- The anchor is the longest contiguous suffix of the workspace path appearing in the guess; it must appear exactly once (ambiguous → untouched). Only its full relative tail is considered — a missing tail leaves the call untouched. No fallback to shorter anchors; the tail is never trimmed, so the rescue can never swap the requested destination for a different file.
- Candidates require a successful `stat`: ENOTDIR/EACCES/EIO never trigger a rewrite.
- `..`/`.` segments anywhere in the guess disable the rescue.
- Runs before `search-path-guard` so grep/glob guesses are rescued instead of being blocked by the guard's ENOENT check.

## Tests

14 hook tests including regressions for: ambiguous anchor (repeated occurrences), no-fallback from longest anchor, meaning-change (`…/nonexistent-subproject/README.md` stays untouched), ENOTDIR candidates, `..` literals, and the canonical dropped-parents rescue.

Suite: 2868 pass / 0 fail; Biome and typecheck clean.

Refs #1143


dhaern:
@greptile-apps Addressed in a1aacb21 → c74c142b: the rescue now segments both the guess and the workspace root with the injected path flavor's own separator and normalizes forward slashes before anchor matching, so a `C:/Users/…` dropped-parents guess re-anchors correctly on win32. Added a regression test driving the win32 flavor (injected path operations + exists) so it runs on every platform. Suite 2869 pass / 0 fail. Please update the review.

PR #1190, opened 2026-09-14
feat(config): per-agent skills_add/skills_remove directives

Aveer:
## Summary

Adds optional per-agent `skills_add` / `skills_remove` configuration directives so project-local config can add or remove skills on top of an inherited list without duplicating the global `skills` array. General deep-merge array replacement semantics are unchanged.

## Semantics

The directives are configuration inputs, not SDK agent properties. They are folded into the effective `skills` token list at `RuntimeConfig.agents()`, after user + project + preset/root precedence merging and runtime `/preset` switching, so downstream consumers only see a plain effective `skills` list.

- Without a configured `skills` list, directives resolve against the agent's default grants (orchestrator: `*`; others: bundled/default grants).
- `skills_add` appends/dedupes additions without overriding active `!name` exclusions.
- `skills_remove` removes final skill tokens and wins over addition for the same plain skill name.
- `skills_remove: ["!name"]` explicitly lifts an inherited exclusion before additions are evaluated, so the same skill can be re-added in that layer.
- With `*`, a plain-name removal is represented as `!name`; `*` is never expanded during merging.
- Legacy agent aliases are canonicalized before default grants are derived.

### Project-local add/remove on top of a global list

```jsonc
// ~/.config/opencode/oh-my-opencode-slim.jsonc
{
  "agents": {
    "oracle": {
      "skills": ["codemap", "deepwork", "verification"]
    }
  }
}
```

```jsonc
// <project>/.opencode/oh-my-opencode-slim.jsonc
{
  "agents": {
    "oracle": {
      "skills_add": ["nexus-backend"],
      "skills_remove": ["verification"]
    }
  }
}
```

Effective skills:

```json
["codemap", "deepwork", "nexus-backend"]
```

### Remove one skill from a wildcard grant

```jsonc
{
  "agents": {
    "orchestrator": {
      "skills": ["*"],
      "skills_remove": ["legacy-skill"]
    }
  }
}
```

Effective token list:

```json
["*", "!legacy-skill"]
```

The wildcard is kept as-is; the removal becomes an explicit exclusion.

### Lift an inherited exclusion and re-add the skill

```jsonc
{
  "skills": ["a", "!foo"],
  "skills_add": ["foo"],
  "skills_remove": ["!foo"]
}
```

Effective skills:

```json
["a", "foo"]
```

Removing `!foo` lifts the inherited exclusion before additions are evaluated.

### Add-only configuration keeps default grants

```jsonc
{
  "agents": {
    "oracle": {
      "skills_add": ["project-skill"]
    }
  }
}
```

This does not replace Oracle's default grants. `project-skill` is added on top of them. Likewise, `skills_remove` without an explicit `skills` list prunes from the defaults rather than replacing them.

## Layering

The directives are ordinary agent config fields and follow the existing field-level layering of user config, project config, presets, root agent config, startup-selected presets, and runtime `/preset` switching. Resolution happens at read time in `RuntimeConfig.agents()`, after those layers are combined.

This means a higher layer can replace/suppress a lower layer's directives, including with an empty array, while general array merge behavior elsewhere remains unchanged.

## Review follow-up

Addresses the maintainer review on this PR:

- fixes the documented re-allow case above;
- adds a focused regression for exclusion lifting and remove-wins-over-add;
- canonicalizes legacy agent aliases before deriving default grants;
- adds an integration test that drives `skills_add` / `skills_remove` through `RuntimeConfig.agents()` and the actual `filter-available-skills` prompt transform.

The unrelated CI/test-repair and formatting commits that were previously on this branch have been removed from the PR history. The test repair Alvin called out has since landed upstream independently, so this PR is again scoped only to the skill-directive feature and its tests/docs.

## Main changes

- `src/config/schema.ts` - optional `skills_add` / `skills_remove` on agent overrides/presets
- `src/cli/skills.ts` - effective skill resolver, alias-aware default grants
- `src/config/utils.ts` / `src/config/runtime.ts` - normalize directives at runtime agent resolution
- `src/config/skills-add-remove.test.ts` - resolver/config/runtime/agent-permission coverage
- `src/config/skills-add-remove-review.test.ts` - maintainer-review regressions
- `src/hooks/filter-available-skills/skills-directives.integration.test.ts` - end-to-end available-skills filtering coverage
- docs/schema updates for the new directives

## Verification

Current head `49cca1a8` is green in GitHub Actions:

- Package Smoke - Ubuntu: pass
- Package Smoke - macOS: pass
- CI Canary / Bun latest: pass
- CI Required / Bun from package.json: pass
- formatting/lint, typecheck, full tests, and build all pass


alvinunreal:
The additive overlay design is good: keeping `skills` replacement-based and adding plural `skills_add` / `skills_remove` directives is clearer than overloading token syntax or introducing a nested permission object. Resolving once in `RuntimeConfig.agents()` and stripping the directives before consumers see them is also the right boundary.

One correctness issue to fix before merging: the documented re-allow path currently does not work. With:

```json
{
  "skills": ["a", "!foo"],
  "skills_add": ["foo"],
  "skills_remove": ["!foo"]
}
```

`skills_remove: ["!foo"]` should lift the inherited exclusion and allow `foo`, but the resolver derives exclusions before it removes `!foo`, so it drops the addition and returns `["a"]`. Apply exclusion-token removals before filtering additions (or otherwise account for lifted exclusions), and add this as a regression test.

Two follow-ups worth addressing:
- Canonicalize legacy agent aliases before deriving default grants for directive-only configurations.
- Add an integration test for the available-skills filter, not just agent permission output.

No new secret/key exposure is introduced by these fields. They should remain skill identifiers only; credentials should be read by the invoked skill from its runtime environment, never interpolated into `skills_*` config values.

The unrelated test repair and formatting commits would be cleaner as separate PRs.

Aveer:
Addressed the review feedback on the current head:

- fixed the re-allow path so `skills: ["a", "!foo"]` + `skills_add: ["foo"]` + `skills_remove: ["!foo"]` resolves to `["a", "foo"]`;
- added regression coverage for that exact case and for remove-wins-over-add;
- canonicalized legacy agent aliases before deriving directive-only default grants;
- added an integration test exercising the directives through `RuntimeConfig.agents()` and the actual `filter-available-skills` transform;
- removed the unrelated `src/index.test.ts` / formatting commits from this PR's branch history.

The TUI hydration test repair is already present upstream as `ad9e8111` (`fix(tests): fire session.status busy before chat.message in #1183 hydration tests`), so I did not open a duplicate follow-up PR for it.

Current head `49cca1a8`: CI (pinned Bun + Bun latest) is green, and Package Smoke is green on Ubuntu/macOS.

alvinunreal:
thanks @Aveer

Issue #1056, opened 2026-08-21
Orchestrator reconcile loop: acknowledging a terminal job by spawning new tasks creates self-reinforcing cycle


snow930:
## Summary

When a background task completes, the Background Job Board lists it under "Active / Unreconciled" and the sentinel instructs: "Reconcile terminal jobs before final response." The orchestrator prompt does not specify HOW to reconcile. In practice the model interpreted "reconcile" as "spawn another subagent to acknowledge/verify the completed task". Each spawned acknowledgment task itself becomes a new terminal-unreconciled job, producing a self-reinforcing loop (30+ redundant dispatches in one observed session).

## Reproduction

1. Dispatch any background task; let it complete.
2. Board shows it as unreconciled. Orchestrator misreads this as pending work and dispatches a new specialist task to "acknowledge" it instead of consuming via task_result.
3. The new task completes -> appears as another unreconciled entry -> repeat indefinitely while the session is idle.

## Expected

Reconciliation should be unambiguous in the orchestrator prompt: consume the terminal result with task_result(task_id) — never spawn a new task to acknowledge one.

## Suggestions

1. Prompt fix: in the orchestrator prompt's Background Task Discipline section, add explicit rules, e.g.
   - "Reconcile a terminal job only by reading its final message via task_result(task_id) and integrating the conclusion. Never dispatch a new subagent to acknowledge/close/verify an already-completed task."
   - "Reconcile each task_id at most once; ignore unreconciled entries whose results were already integrated."
2. Board hardening: detect repeated same-parent dispatches targeting the same objective/task_id within a short window and suppress or warn in the board snapshot (anti-loop guard). Optionally distinguish orchestrator-spawned meta/acknowledgment tasks from user-driven work so they do not feed back into Active / Unreconciled.

## Environment

- oh-my-opencode-slim master (2026-08), OpenCode with native background subagents enabled

mhenke:
> *This was generated by AI during triage.*

**Prompt clarification:** 'reconcile' in the Background Job Board sentinel means 'integrate delivered results into the final response' — never 'spawn acknowledgment subagents'. Acknowledgment dispatches self-reinforce (30+ redundant dispatches). Fix: adopt clarified semantics in #1051's fix; separately surface lastLiveBusyAt/elapsed in board tail + task_status (see #1032 force gate).

mhenke:
> *This was generated by AI during triage.*

## Triage Notes

**What we've established so far:**

[...]

PR #1193, opened 2026-09-14
fix(post-file-tool-nudge): steer drifting orchestrators back to delegation

dhaern:
## Problem

Stronger models drift out of the orchestrator role. Instead of dispatching bounded work to specialists, the coordinator inspects, edits, and runs things itself. The existing drift nudge only fired after `Read`/`Write` and reused the phase-reminder payload, so an orchestrator working through `edit`, `bash`, `apply_patch`, or `shell` never saw a reminder at all, and when both fired they were deduplicated into one message.

## Design

Two delivery paths, both derived only from stable history. No session state, no pending flags, nothing volatile on the stable prefix.

Stable part: every orchestrator user message whose preceding assistant block belongs to the same session's orchestrator and used an implementation tool carries a tagged nudge part with its own text and metadata. The condition is evaluable on the first render and never changes, so re-renders reproduce byte-identical history.

Reconciliation is selective, never strip-and-reappend: a part that still corresponds to history keeps its exact position even when other hooks (phase-reminder) append their own parts after it, a part that no longer corresponds is removed from that message alone, and a missing one is appended. Specialist blocks (different agent or session) never attribute drift to the coordinator.

Volatile trailing: when the payload ends in an orchestrator assistant block mid tool-loop, the nudge rides as its own tagged trailing message, published late through a dedicated `deliverTrailingNudge` step that runs after phase-reminder and the job board. Neither hook can anchor on it or append to it, so board checkpoints only ever anchor on real history. The tail scan skips only board-tagged infrastructure; internal-initiator user turns are persisted history and count as the loop having ended.

## What came out of review

Six review rounds with a second-opinion reviewer (four rejections) shaped the current shape, each blocker reproduced before the fix:

- In-flight pending delivery mutated the last user message between renders and broke the cached prefix; delivery is now history-derived.
- A trailing message published before the job board let `checkpoint-compatible` anchor a checkpoint to it, so the next turn discarded retained checkpoints as if compaction had happened; publication moved after the board.
- Strip-then-reappend reordered stable parts behind parts other hooks had appended; reconciliation became selective.
- A generic synthetic-message skip treated internal wake turns as infrastructure and nudged over them; the scan now skips only board-tagged messages.
- Board checkpoint snapshots (user-shaped) were accepted as nudge recipients and cut the backward scan; they are now transparent to both.

## Tests

35 tests covering: tool-set gating (implementation tools vs `grep`), byte-stable position across renders with phase-reminder composition, gate flips leaving no residue, same-session attribution (a fixer's edits never nudge the orchestrator), internal turns ending the loop condition, and snapshot transparency with three consecutive renders byte-identical. The cache-safety harness runs the real pipeline order including the late trailing step. Full suite: 2892 pass; the single failure is the pre-existing `src/index.test.ts` failure on master (2.2.20), identical on a clean checkout.


dhaern:
@greptile-apps addressed the P2 in aea1159f.

You were right that the reconciliation made a stable part mutable: when `shouldInject` flipped to false after a part was rendered, the next render removed it and rewrote the cached prefix.

The fix separates the two concerns: history eligibility (immutable for an existing message, since its preceding assistant block cannot change) now alone decides whether an already-rendered part is KEPT, and the mutable session gate only decides whether a MISSING part may be appended now. A gate flip can therefore never remove rendered bytes; it can only stop future appends. The gate also became lazy on the stable path, so it is not consulted for messages history does not qualify (the updated `passes the derived session ID to shouldInject` test now exercises it through the trailing path, which always consults it).

Regression added: part rendered under gate=true, re-render under gate=false → part survives, bytes identical. Affected suites 36/36; full suite 2893 pass with the single pre-existing master failure (`src/index.test.ts`, 2.2.20).

Please update the review.

dhaern:
Closing this PR, and apologies for the noise.

It targets #1012, and I framed it as a fix for a confirmed defect. It is not: the triage note on #1012 (comment 5383874197) reclassifies it as an enhancement, groups it into the orchestrator-prompt-discipline cluster with #1056, and explicitly leaves it in evaluation with no concrete spec yet ("a grilling session can turn it into an actionable brief; otherwise it stays in evaluation"). Opening an implementation PR against an issue parked in evaluation, presented as a bug fix, was the wrong call. My mistake.

The branch stays available if a brief ever comes out of that evaluation; the cache-safety mechanics here (history-derived stable parts, late volatile trailing) are the parts worth salvaging in that scenario.

@alvinunreal sorry for the review burden.

PR #1195, opened 2026-09-14
fix(orchestrator-wake): continue lifecycle prompts in the session's current selection

### LINKED ISSUE #1079: Background task lifecycle forces Orchestrator after switching parent to Plan or Build

dhaern:
Fixes #1079

## Problem

Two symptoms, one root cause.

1. A parent session manually switched from Orchestrator to Plan/Build still takes lifecycle turns as `orchestrator`. You are working in Plan, a background child finishes, and the injected turn runs orchestrator logic with orchestrator instructions. Switching mode stops working as an escape hatch, and it self-perpetuates: the synthetic turn re-marks the session as orchestrator in the plugin's own tracking, so the next lifecycle event does the same.

2. A parent whose model was manually changed wakes on the old model. The wake body resolved its model from `latest.model ?? snapshot.model ?? getObservedWakeModel(...)`. After a manual model change none of those sources matches, the body is sent without a model field, and the host applies the agent's configured default. This is the usual quota case: you switch models to avoid a limit, the wake goes straight back to the limited one.

Root cause: every lifecycle continuation (periodic TODO wake, children wake, stopped-recovery wake, completed/error notification) is injected with `agent: 'orchestrator'` hardcoded and a model resolved from stale history. Three defects cooperate:

- `orchestrator-wake/index.ts` and `revived-run-tracker.ts` hardcode the agent and a stale model chain in their `promptAsync` bodies.
- Those same synthetic prompts flow back through `chat.message` / `message.updated` and overwrite the tracked selection (`setAgent` / `setModel`). v1 never registered internal admissions by message id, so `message.updated` could not even classify them; v2 had a registry but only fed it from the shim.
- On v2 hosts the shim re-asserted the stale pin via `switchModel`, which reverts a persisted user selection that changed after the pin was captured.

The existing `backgroundJobs.orchestratorWake.enabled: false` switch only disables the periodic wake and stopped recovery. The completed/error notification path has no switch, so there was no configuration-level way out.

## Resolution hierarchy

New `src/utils/session-selection.ts` resolves what the session uses *now*:

1. Host-persisted selection via `session.get()` (receiver-preserving `.call`, `{data}`/`{info}`/bare envelopes, bounded to 2s so a hung host read cannot stall a wake).
2. Slim metadata of the last externally observed selection (internal admissions excluded).
3. Unknown. Callers fall back to `orchestrator`, which matches the historical behavior for sessions that always ran orchestrator. The timeout bounds only the host read: if the host hangs and metadata knows Plan/model-B, the wake still goes out as Plan/model-B.

## Send-time resolution, with revalidation

The wake and the terminal notification resolve the selection right before sending, then re-check everything the await could have invalidated: parent generation, continuous idle, schedulability, and the resolved agent against the Plan/Build eligibility gate. Stop deltas are pruned a second time after the await, so a child that leaves stopped/unreconciled during the resolve window is dropped from the wake instead of being delivered stale. Model and variant travel as one selection, so a session on A/high that resolves to B never sends B/high.

Plan/Build parents remain eligible for lifecycle wakes while they have pending delegated work (running children or unreconciled terminals). A Plan session that never delegated gets no wakes. This keeps the issue's expected behavior (the child's result reaches the parent in the mode the parent chose) without turning every Plan/Build session into an orchestrator.

## Tracking

- `chat.message` inspects both `input.parts` and `output.parts` (an empty input array no longer hides output parts) and records internal admissions by message id. `message.updated` then skips selection writes for that id and for assistant replies carrying its `parentID`. Executed-model accounting (`migrateTask`) still runs for both, so concurrency and fallback behavior are unchanged.
- `registerSessionAsOrchestrator` now marks task-management membership instead of rewriting the tracked agent. `shouldManageSession` accepts orchestrator selection or membership.
- Task-management membership no longer makes a session permanently unevictable: when the store is over capacity and only protected entries remain, eviction falls back tiered (unprotected first, then oldest task-managed, then in-flight orchestrators), so the configured `maxEntries` cap always holds even when deletion events are missed.
- The v2 agent-discovery forward (agent/model without parts) is classified through the same admission registry.

## v2 shim

Lifecycle continuations (wake, terminal notify) pass an explicit `modelSelection: 'inherit'` argument: the shim routes them through `session.synthetic({resume: true})` and lets the host's persisted selection stand, on the synthetic path and on the degraded `session.prompt` path. Their body pin — with or without a variant — is a snapshot by delivery time, and re-asserting it reverts selections the user changed in between. The flag is deliberately opt-in: purely internal callers with a real pin (interview continuations track their own model) keep switching. Mixed bodies (foreground-fallback replay) still switch, and `modelSwitch: 'required'` still switches.

The terminal notification also moves to `delivery: 'queue'`, the same TOCTOU class #1192 closed for task-revive: on v2 the default steer would hijack an in-flight parent run if the session went busy between the status check and the send.

## Known limitation

A selection changed in the TUI but not yet admitted by any message lives only in the TUI's local store. Neither v1 nor v2 exposes it to plugins, so the wake cannot see it. Coverage starts from the first message sent with the new selection. Fixing the remaining window needs a host-side capability (publishing the pending composer selection), which is out of scope here.

## Tests

Global suite: 2946 pass / 0 fail; typecheck and Biome clean.

Line breakdown — **production: +540 / −76 across 6 files**; **tests: +978 / −41 across 6 files** (net +1,401: the two new files account for 378 of the production side, and the test side carries the race/retry regressions the review rounds required).

New coverage, by area:

- Reader: receiver preservation (v1 `this._client`), `{info}` envelope, host timeout with metadata (still Plan/B) and without (unknown).
- v1 tracking: internal `chat.message` registers the admission; `message.updated` with the same id and an assistant reply with `parentID` do not overwrite the tracked model. No manual registry seeding.
- Wake: Plan parent with pending delegated work gets the recovery wake as `agent: 'plan'`; Plan without delegated work gets none; external message during the resolve aborts the send; a stop fact that goes stale during the resolve is dropped; A/high resolved to B sends B with no variant, on both the v1 and the v2 (`hostFlavor: 'v2'`) paths.
- Tracker: notification carries the resolved agent/model, re-resolves on retry, no send after dispose mid-resolve (with entry signal), queue delivery.
- Shim: lifecycle continuations with `modelSelection: 'inherit'` never call `switchModel` (synthetic and degraded paths, with or without variant); a purely internal body without the flag (interview-style, real pin) still switches; mixed replay with `modelSwitch: 'required'` still switches.
- Metadata: task-managed sessions survive overflow while unprotected entries exist, and the oldest task-managed entry is evicted when nothing else is evictable.


dhaern:
@greptile-apps both findings addressed in 18da056b:

**P1 (interview model pin discarded):** agreed, and fixed by inverting the gate. Inheriting the persisted host selection is no longer inferred from the body being purely internal — it now requires an explicit `modelSelection: 'inherit'` argument on `promptAsync`, which only the lifecycle callers (orchestrator-wake, terminal notification) pass. Purely internal bodies without the flag keep the existing pin semantics, so interview continuations switch to their tracked model again (`client-shim.ts`, regression test `interview-style internal continuation keeps its explicit pin`). On v1 hosts the flag is dropped by the SDK's RequestInit path (same as `delivery` in #1192), and v1 has no `switchModel` race to begin with — the send-time-resolved body model is applied directly.

**P2 (metadata cap bypass):** fixed with tiered eviction. Task-managed sessions are still preferred (they survive overflow while unprotected entries exist), but when nothing unprotected remains, the oldest task-managed entry is evicted next, and in-flight orchestrators last — so `maxEntries` is always enforceable even when deletion events are missed (`session-metadata.ts#track`, regression test `evicts the oldest task-managed session when nothing else is evictable`).

Also updated the PR description: the v2 shim section documents the opt-in flag, and the tests section now carries the line breakdown (production +540/−76, tests +978/−41). Full suite 2946/2946, typecheck and Biome clean. Please re-run the review and update the score.

PR #1197, opened 2026-09-14
feat(tui): clickable subagent sessions in the sidebar

dhaern:
## What this does

Agent rows in the sidebar are now mouse targets.

- One active subagent: clicking the row opens its transcript in the current window.
- Several instances of the same agent (two oracles, for example): the row expands into an inline list showing alias, model, and a colored status (`active` in green, `retrying` in amber). Clicking an entry opens that session.
- Rows highlight on hover, one row at a time, including each entry of the expanded list.

## Why it helps

Reaching a subagent transcript today means scrolling the session list or hunting for its pane. With the pane system disabled (multiplexer off), subagents run in the background and the sidebar is the only place they surface. With panes enabled, the right pane is often offscreen or buried among others. A click on the row that already shows the agent and its spinner is the shortest path to the transcript. Returning to the orchestrator is one keystroke with the existing `session.parent` keybind.

## How it works

- Session identity (alias, model, status) is projected from the background job board into `tui-state.json` through a launch-identity listener on the coordinator. The model is published when observed in `chat.message` (covers v2, whose synthesized `message.updated` carries no provider/model) and again on activation when it was learned earlier, so both orders converge without resurrecting idle sessions.
- Targets are scoped to the visible conversation root. Windows viewing different conversations never mix sessions, and activity from another root does not appear in the list.
- Navigation is feature-detected: v1 hosts use `route.navigate('session', ...)`, v2 hosts use `router.navigate({type:'session', sessionID})`. Hosts without the capability keep the sidebar informative.
- Hover mutates the row renderables in place. The render reads no hover signal, so the 100 ms spinner tick and the hover never rebuild the clickable tree, and a click cannot be dropped between press and release. Clicks fire only on the primary button and are suppressed while a text selection is active.
- Interaction state (expanded groups) lives per window. Two TUIs attached to the same daemon share activity data through `tui-state.json` (lock-serialized writes) but never share expansion or navigation state.

Two fixes came out of the review rounds:

- The startup sweep no longer drops live activity just because it belongs to the current PID. A second plugin initialization in the same process (server plus another instance) used to wipe the first instance's in-flight sessions from the sidebar.
- The 1 s snapshot poll stat-caches the file (dev:ino:mtimeMs:size). An unchanged file costs one `stat` instead of a read plus parse. Writers publish through tmp+rename, so a changed write always invalidates the cache.

## Notes

- Mouse only in this iteration; keyboard navigation is not included.
- `onClick` is inert in @opentui 0.5.8, so rows use `onMouseUp`. This is pinned by a contract test that exercises both handlers on the real renderer.

## Size

- **Production: +932 / −62 (net +870)** across `tui.ts`, `tui-state.ts`, `index.ts`, `background-job-coordinator.ts`
- **Tests: +976 / −4 (net +972)** across the matching test suites

## Testing

- 129/129 in the affected suites (`tui`, `tui-state`, `index`, `background-job-coordinator`); typecheck and Biome clean.
- New coverage includes mounted-sidebar tests with `testRender`: expansion with several sessions, column layout with long model names, hover limited to the hovered row while its sibling stays unchanged, navigation clicks, and the no-navigation fallback. Integration tests cover both busy/launch orders of alias persistence and both model projection paths. Sweep regressions cover live same-PID activity and dead residue.


dhaern:
@greptile-apps Addressed the P2 (unbounded snapshot cache) in da440532: the async cache is now a bounded LRU (8 entries). Hits re-insert to refresh recency, misses insert at the front of eviction order, and reads that fail drop their entry. Covered by a regression test that polls 8 other projects and asserts the first entry is evicted (fresh object identity on re-read).

Also rebased onto master (2fc0ea82) to resolve the conflict in src/index.ts: master's #1079 internal-admission guard for sessionMetadata.setModel is preserved, while the sidebar's per-session model detail still publishes for every executing model regardless of admission origin. Full suite: 2993 pass / 1 fail (dependency-contract.test.ts, pre-existing on clean master, verified by running it directly on origin/master).

Please update the review.

alvinunreal:
@orchestrat0r fix conflics and merge

PR #1198, opened 2026-09-15
feat(tui): position the sidebar slot by plugin-list index

dhaern:
## What this does

Slim's sidebar section used to register with `order: 900`, so it always rendered at the bottom. This PR derives the order from the plugin's position in the host's plugin list: index 0 in `tui.json` (or `opencode.json`, whichever the host merged) maps to order 110, right after the host's context section. Each later index adds 100.

Moving a line in an existing config file now moves the section. No new setting.

## Scope

This moves slim's own slot only. Other plugins keep whatever order they register with, so the list order relative to them is not guaranteed. A plugin registering exactly `110 + index × 100` would tie, and registration order would decide. Ordering every plugin by list position needs either a shared convention or a host change; neither is in reach from a plugin.

v1 hosts only. The v2 slot claim API takes `{append, render}` and has no order parameter, so v2 keeps its current position.

## How it works

- The order is resolved once, at slot registration, from `api.tuiConfig.plugin`: the effective, already-merged list the host exposes. No config files are read or re-parsed.
- Spec matching handles plain strings and `[spec, options]` tuples (the format the installer writes for versioned installs). `file://` checkouts and plain absolute local paths match by path or basename; npm specs match with or without a version. A scoped package with the same basename never matches: different package.
- Anything unexpected (missing list, absent spec, malformed entries) falls back to the historic 900. Existing setups render exactly as before unless the plugin is listed.

## Size

- **Production: +51 / −1** (`src/tui.ts`)
- **Tests: +133 / −0** (`src/tui.test.ts`)

## Testing

- 47/47 in `src/tui.test.ts`: band mapping for indexes 0 and 1, tuple entries, versioned npm specs, scoped-package rejection, `file://` with trailing slash, plain absolute local paths, malformed entries keeping their index slot, and fallback to 900 for missing or non-array lists. Two registration tests verify the wiring: `tuiConfig.plugin` with slim at index 1 registers order 210, and an absent `tuiConfig` registers 900.
- Full suite 2973/2974; the one failure is `dependency-contract.test.ts`, pre-existing on clean master. Typecheck and Biome clean apart from the known `mirror-conformance.ts` errors on master.


dhaern:
@greptile-apps P1 is fixed in 33e50513.

Plain absolute local paths (`/workspace/oh-my-opencode-slim`, the form the installer writes for source installs) now match by basename, same as `file://` checkouts. Added a unit test for that path. Please re-review.

Issue #1200, opened 2026-09-15
[Bug]: Background task remains running and retries with <none> after provider stream timeout


ricatix:
### What happened, and what did you expect?

A background Oracle task using a single configured model received a terminal upstream stream error, but the task lifecycle did not become terminal. The TUI retried it as `Retrying (attempt 2) · <none>`, while plugin logs continued to report the child as `running` for more than two minutes after the gateway had already ended the request with `stream stall timeout`.

Expected behavior:

- the provider error should be preserved instead of rendered as `<none>`;
- the background task should transition once to a terminal `error` state;
- the parent should receive that terminal result and be able to reconcile it.

This resembles the lifecycle family tracked in #595 and #998, but this reproduction does **not** use a model-array fallback: Oracle is configured with one model string. The trigger is a terminal streaming error from the provider/gateway.

The plugin was also running in degraded compatibility mode: startup logs report unavailable `ctx.storage`, `ctx.generate.text`, `ctx.tool.transform`, `ctx.session.hook`, `ctx.tool.hook registration`, and `ctx.event.subscribe`, followed by `children-driven degraded mode`. This may be relevant to the missing terminal reconciliation.

### Steps to reproduce

1. Use OpenCode 1.18.31 with oh-my-opencode-slim 2.2.20.
2. Configure Oracle with one streaming model (not a fallback array).
3. Start Oracle through `task(..., background: true)` with a large context.
4. Make the upstream Responses stream end with an error after a long silent gap (observed: `stream stall timeout`).
5. Observe the task UI and plugin lifecycle logs.

Observed sequence (UTC, identifiers redacted):

```text
03:16:46 child state=running; upstream streaming request starts
03:29:25 gateway ends request: stream stall timeout
03:31:39 child is still reported state=running
TUI: Retrying (attempt 2) · <none>
```

A separate lifecycle observation in the same environment produced:

```text
state=error
timedOut=false
hasResult=true
terminalUnreconciled=true
```

### oh-my-opencode.json

```jsonc
{
  "agents": {
    "oracle": {
      "model": "9router/cx/gpt-5.6-sol",
      "variant": "high"
    }
  }
}
```

### OpenCode version

1.18.31

### oh-my-opencode-slim version

2.2.20

### Operating system

macOS (client); Linux/Docker (gateway)

### Logs, screenshots, or extra context

```shell
ctx.storage unavailable
ctx.generate.text unavailable
ctx.tool.transform unavailable
[...]

PR #1202, opened 2026-09-15
fix(task-session-manager): preserve serialized NamedError detail in board error summaries

dhaern:
## What this does

When the host publishes a `session.error`, the payload on the wire is a serialized `NamedError`: `{ name, data }`, with the human-readable message inside `data.message` (`APIError`, `ProviderAuthError`, and the rest all follow this shape; see `NamedError.toObject()` in the core). The event router's error branches read only the top-level `message` field, which is `undefined` for every serialized NamedError, so the board recorded the generic `Session error` even when the provider detail existed two levels down.

This extracts `data.message` first, falls back to a top-level `message` for non-NamedError payloads, and keeps the generic summary when neither exists. It runs in both `session.error` branches (managed session and child). Terminalization, fallback, generation, and reconciliation decisions are unchanged; only the recorded summary improves.

Motivated by the diagnostics in #1200: with this fix, a terminal provider failure records the actual message ("stream stall timeout") instead of a generic string. To be clear about scope, that issue involves several things this PR does not address: the gateway stream stall itself, the host's retry policy (a session stays legitimately busy while the core retries, so `running` during `Retrying (attempt N)` is not a plugin bug), and the `<none>` retry message, which comes from the provider-to-core error chain rather than the board.

## Size

- **Production: +30 / −4** (`src/hooks/task-session-manager/event-router.ts`)
- **Tests: +169 / −0** (`src/hooks/task-session-manager/index.test.ts`)

## Testing

- 167/167 in the task-session-manager suite.
- New regressions: serialized NamedError detail preserved on both the child and managed-session branches; `data.message` preferred over a top-level message (NamedError instances carry their class name at the top level); empty `data.message` falls back to the top-level message; no message at all falls back to the generic summary. Existing top-level-message tests keep passing.
- Typecheck and Biome clean (only the pre-existing `mirror-conformance.ts` errors on master).


dhaern:
@greptile-apps Good catch. Fixed in f439e466: both extraction levels now validate with `trim()`, so a whitespace-only nested message falls through to the top-level message or the generic summary instead of leaving an empty board summary. Regression added (whitespace-only `data.message` with a valid top-level message). Suite at 168/168. Please re-review.

Issue #998, opened 2026-08-10
Explorer task shows 'cancelled' during model fallback


mhenke:
## Problem

When the orchestrator delegates to a child task (like Explorer) and the first model fails, the foreground fallback kicks in to retry with a different model. But the task shows "cancelled" in the TUI even though the task is still running in the background. The orchestrator sees "cancelled" and reacts by retrying or canceling duplicates.

## Root cause

Hook registration order in `src/index.ts`. The task-session-manager's event handler runs **before** the foreground-fallback's event handler:

```
// Current (broken) order:
await handleTaskSessionEvent(...);  // runs first
await foregroundFallback.handleEvent(input.event);  // runs second
```

When a child session errors:

1. Task-session-manager checks `isFallbackInProgress(sessionID)` → `false` (fallback hasn't started yet)
2. Job marked as `"error"` immediately
3. Foreground-fallback runs after → sets `inProgress` → but too late

## Fix

Swap the order so `foregroundFallback.handleEvent` runs first. The synchronous `inProgress.add(sessionID)` at the top of `tryFallback()`/`tryFallbackWithAbort()` means the task-session-manager sees `isFallbackInProgress() === true` and skips premature terminalization.

```typescript
// Fixed order:
await foregroundFallback.handleEvent(input.event);  // runs first
await handleTaskSessionEvent(...);  // runs second
```

## Verification

- All 176 task-session-manager tests pass
- Typecheck clean
- Tested with real orchestrator delegation: Explorer task continues running in background despite first model failure, no premature "cancelled" status

BaconDroid:
Confirmed independently on OMO Slim 2.2.11 with OpenCode 1.18.16. This is exactly the event-order race described here, and it affects background subagents after a fallback-eligible quota/rate-limit failure.

From `oh-my-opencode-slim.20260813T144227.log` (UTC):

```text
14:51:43.840  child idle; multiplexer sees backgroundJobState:"error"
14:51:43.954  [foreground-fallback] switched explorer deepseek-v4-flash -> gpt-5.6-luna
14:51:43.990  child busy, but task manager reports previousState:"error" -> updatedState:"error"
14:52:01.856  parsed task output completed hasResult:false; board updated completed, terminalUnreconciled:true
14:52:02+    child continues emitting busy events while board remains completed/reconciled
```

[...]

PR #1203, opened 2026-09-15
fix(task-session-manager): confirm stop from transcript evidence, own fallback delivery, treat missing status as idle

dhaern:
**P1 — recommended to merge.** A finished background child can be published as `stopped` with no result, so the parent revives and either duplicates the work or gets nothing. That happened in production: a 17k-character report was already in the transcript while the board said the job had stopped empty. OpenCode v1.18.31 drops idle sessions from `session.status`, so the idle timer treats "absent" as "no result." Background fallback re-prompts have the same hole: the native notifier stays bound to the original job, and nobody delivers the substituted run. This is not a crash or data-loss P0, but it is a real lifecycle break on revive/fallback and it is worth landing.

## What this does

Follow-up to the lifecycle work in #1191, #1192, and #1195. Those PRs recovered stopped sessions, delivered revive prompts with queue semantics, and kept lifecycle prompts on the session's current model. Three older holes still let a finished background child look stopped with no result.

### 1. Stop confirmation reads the transcript

[ef623dc6](https://github.com/alvinunreal/oh-my-opencode-slim/commit/ef623dc6e61c4de4d6d15f8503d610092b8a8345) in #1161 by @GoldJohnKing settled quiescent jobs from host outcome / an absent status-map entry so they would not stick in `running` forever. On v1 that map deletes idle sessions, so "absent" is the normal idle state, not proof of no result.

The stop gate now reads the child transcript and uses the same classifier as the revive tracker (`classifyAssistantTurnEvidence`). A valid empty post-baseline segment still becomes `stopped` (jobs must not stick in `running`). Hung, malformed, or timed-out reads stay `running` with `statusUncertain`; retries are per quiescence episode and never fall through to `stopped`. A later valid read settles the job.

Substituted snapshots cannot terminalize: the open SDK read is keyed by generation + revision + baseline + quiescence episode, and the outcome publisher (`readSessionOutcome`) checks the same identity plus any pending fallback handoff before `updateStatus` / `markReconciled`. That keeps #998 closed (premature terminalize of a replaced run).

### 2. Fallback re-prompt transfers delivery ownership

[231531ae](https://github.com/alvinunreal/oh-my-opencode-slim/commit/231531ae6ca4ba2b30f9a3dda979b7c5de55abae) in #199 by @jmvbambico added foreground fallback via `session.promptAsync`. The native host notifier stays bound to the original job; `RevivedRunTracker` only covered explicit revives. After a successful fallback re-prompt the child could finish with nobody delivering to the parent.

Before `promptAsync`, the fallback now prepares an observation handoff so every terminal publisher (stop gate, tracker probe, outcome reader) defers while admission is in flight. Admit converts that preparation into a tracked run and probes immediately. Reject covers an explicit host refusal (error envelope / typed capability rejection). A transport failure with no response promotes the preparation into the owning run instead of dropping it (`settleUnresolved`): the host may still have accepted the replay. After promotion the gate stays fenced for one more expiry window, then the fence lifts without discarding the owner, so a still-empty transcript can confirm `stopped` and a late result still has a delivery owner.

A v2 `switched: false` result means the replay was delivered on the current model. That is admitted work; only the model-switch claim is suppressed. A confirmed background child whose generation changed during the transcript read aborts the replay instead of prompting the wrong execution.

### 3. `task_cancel` treats a missing map entry as idle

[940730ee](https://github.com/alvinunreal/oh-my-opencode-slim/commit/940730ee0c884a47c9f77716812792d094bbee3f) in #481 by @alvinunreal required `status === 'idle'` on the live session map. The same host deletes idle entries, so a finished child looked unknown and cancel refused. A valid map with no entry is now quiescence. Busy, retry, transport errors, and malformed payloads still refuse. The stability window and lease/generation fences are unchanged.

[b9fdc7b0](https://github.com/alvinunreal/oh-my-opencode-slim/commit/b9fdc7b0a7e2ba2d4fe100be36751ba88f1c46eb) in #1007 by @alvinunreal is the board path that notified `stopped` without a result; with (1) that notification only fires after a valid empty transcript.

## What this does not change

- Valid empty post-baseline transcripts still stop (the settle-quiescent guarantee from #1161).
- #998: a snapshot or outcome from a replaced generation/episode/handoff cannot publish terminal.
- #1115: textless assistant turns stay non-terminal.
- No new synthetic messages, no new prompts, no change to model selection.

## Size

- **Production: +1,371 / −70** (`stop-confirmation.ts`, `revived-run-tracker.ts`, `idle-reconciliation.ts`, `runtime-status-reconciliation.ts`, `foreground-fallback/index.ts`, `fallback-observation-transfer.ts`, `cancel-task.ts`, `child-transcript.ts`, `index.ts`)
- **Tests: +1,532 / −4** (gate, tracker handoff, idle outcome fence, fallback admission, cancel map-miss, classifier, bounded unresolved-admission)

The extra production is the three independent terminal publishers (gate, tracker probe, host outcome) plus the fallback handoff that has to fence all three. Collapsing any of those reopens the incident above.

## Testing

- Affected suites run individually (combined `revived-run-tracker` + `cancel-task` in one process still contaminates the same way it does on clean `origin/master`).
- Typecheck clean; Biome clean on the touched files.
- New coverage: valid absence still stops; unknown evidence never stops (`maxEvidenceRetries=0` included); hung SDK read is rejoined not piled; substituted generation/revision/baseline/episode cannot use the prior snapshot; outcome await during an armed handoff does not publish; prepare/admit/reject/unresolved; `switched:false` admits without claiming the switch; stale background generation aborts replay; missing status-map entry cancels as idle; `timeoutMs<=0` disables the consumer deadline; an unresolved promoted handoff lifts the gate fence after one expiry window without dropping the owner.


dhaern:
@greptile-apps Good catch on the permanently pending handoff. That path is a P1: after a transport failure on the retried `promptAsync`, `settleUnresolved` promoted the observation and left `isObservationPending` true with no later admit/reject, so even a valid empty transcript could not confirm stopped and the board stayed `running`.

Fixed in e51e7ac3. After promotion the owner stays installed (a late result still has a delivery owner) but the gate fence now lifts after one more expiry window if admit/reject never arrive. Regression added: unresolved admission lifts the fence without dropping the owner. Tracker suite 25/25, typecheck clean. Please re-review.

PR #1209, opened 2026-09-17
fix(task-session-manager): gate all terminal publications on runtime evidence

dhaern:
**This should be merged.** #1203 closed the false-`stopped` vector only; every other terminal state kept the same class of bug, so without this PR the job board still publishes states the session never reached. The remaining fix is not additive hardening, it is the missing half of #1203, and it lands with less code than it removes.

## The bug this finishes

We hit it on a live install: a background job was reported `completed` while its session was still working, and `task_result` returned an empty placeholder because the final assistant row was a re-prompt. The session kept running for another 45 seconds after its "terminal" publication. A false terminal invites a full redispatch of the subagent's work, which for long tasks means paying for the same job twice.

The root cause was structural. Eight different writers could publish `completed`, `error` or `cancelled`, each with its own idea of when that was safe:

| Writer | Path |
|---|---|
| Native tool output | `tool-execute-hooks.ts` |
| Synthetic injection | `board-injection.ts` |
| `session.error` events | `event-router.ts` |
| Revive / fallback | `revived-run-tracker.ts` |
| Idle and deferred errors | `idle-reconciliation.ts` |
| Rehydration | `task-session-manager/index.ts` |
| Cancellation | `cancel-task.ts` |
| Wall-clock watchdog | `background-job-supervisor.ts` |

`git blame` places the defective paths in August commits, before #1203. That PR fenced the stop path; the other seven writers still accepted synthetic injections, idle events and deferred errors as terminal authority without contrasting the live session map.

## What changed

All eight writers now go through one component, `src/utils/background-job-terminal-gate.ts`:

- Observation identity (generation, activity revision, attempt, episode) is captured before any await and rechecked after every read. Evidence gathered for an older identity can never authorize a publication for a newer one.
- A commit requires a single-use opaque authorization. The board validates it against its own gate instance through a private registry, so no consumer can forge or replay one. The previous `validate(record)` callback seam is gone.
- A busy host invalidates open evidence and withdraws the publication, restoring the concurrency slot. One timer per observation, no per-writer timers.
- Events whose timestamp precedes or equals the publication request a fresh runtime contrast instead of reopening directly or being discarded; `completedAt` is local publication time, not proof of host termination. The contrast obligation survives collisions on both the SDK reader and the gate's internal open-reads registry through a single continuation mechanism, coalescing repeated requests with identity and disposal fences.
- Every publication bumps `terminalRevision`. Acknowledgements, notifications and parent handoffs are versioned, so a stale ack from an earlier publication cannot reconcile a newer run (the old A→busy→B failure).
- The shared gate is disposed by its owner on all three teardown paths (init failure, `server.instance.disposed`, plugin dispose), and pending reads can no longer publish after disposal.

`stop-confirmation.ts` is removed outright; the gate subsumes its discipline, which is where most of the production savings come from. One deliberate contract change: an unknown `task_id` in `tool.execute.before` is now rejected before the call reaches the host, instead of being silently adopted.

## How it was verified

The suite grew from 3,116 to 3,135 tests while the repo got smaller. Numbers at HEAD (0d38cfbf):

- `bun test`: 3,135 pass, 0 fail, 8,055 assertions across 167 files.
- `bun run typecheck`, `bun run lint` (389 files) and a full `bun run build`: clean.
- Production: +1,892 / −2,118 (net −226). Tests: +2,457 / −2,657 (net −200). 45 files changed overall.

Beyond the suite, the implementation went through five rounds of adversarial review against the real plugin assembly (gate, board, coordinator, scheduler and hooks, with SDK responses controlled to force the races). Seven P1 findings were found and closed that way, each with an executed reproduction before the fix. Those reproductions are permanent regression tests now:

1. Idle event with a busy host must not publish.
2. Native after-hook must not release capacity twice after a reopen.
3. The shared gate must not accept work after disposal.
4. Terminal context must attach to ordinary tasks, not just revived ones.
5. Busy delivered after commit must trigger a contrast and withdraw the publication.
6. Busy at the exact publication timestamp must trigger a contrast, not a discard.
7. A historical echo with a quiescent host must contrast without reopening.
8. A contrast colliding with an open SDK read must survive the collision.
9. A new signal during the gate's fresh read must survive the internal registry collision.

Each collision test also asserts single-flight behavior: exactly one SDK read active at any time (`maxActiveReads === 1`), recovery without external signals, same generation, revision incremented, result withdrawn.

Finally, a live smoke test on a running daemon: background completion with real result extraction, cancellation through the plugin, and a manual abort that left the job honestly `running` until the session actually stopped.

## Compatibility

No public API changes beyond the rejected unknown `task_id` noted above. Hosts without `session.status` (v1) keep working: quiescence is deduced from host outcomes, unknown evidence stays `running` with a diagnostic instead of guessing, and one host signal after a retry budget is exhausted resumes resolution.


dhaern:
@greptile-apps Fixed in 0d38cfbf.

Unknown or absent runtime evidence now consumes the evidence retry budget once per failed inspection instead of being a dead end: the timer re-arms up to maxEvidenceRetries (3), so the one-shot callers you flagged (processed final idle, single deadline pass) are reconsidered automatically. Genuine signals (busy, idle, attributed events) still replenish the budget, and collided readers keep their availability-driven continuation without spending it. Once the budget is exhausted the record stays running with a status-uncertain diagnostic until the next host signal resumes resolution, so there is no perpetual polling.

Covered by integration regressions for both variants you described (absent outcome and failed lookup on a host without status polling: recovery publishes completed with no external signal, two gets, zero status calls, slot released) and an extended v2 shim test asserting the bounded cutoff (4 reads total, then silence beyond graceMs x4) with signal-driven recovery afterwards. 3135 tests, 0 failures; typecheck, lint and build clean.

Could you re-run the review?

PR #1210, opened 2026-09-17
fix(tui): clear inactive sidebar agents on subagent completion and idle events

kumar-shivang:
### What changed, and why was it needed?

Fixes an issue where the sidebar agent spinner continues spinning indefinitely even after a subagent/agent finishes and closes.

#### Problem
1. When background subagents terminated, `backgroundJobCoordinator` notified multiplexer and revival managers, but never invoked `markTuiAgentInactive(record.taskID)`. As a result, completed subagent sessions remained indefinitely in `activeSessions` within `tui-state.json`.
2. Foreground task executions (`pending.background === false`) returning in `tool.execute.after` did not mark still-`running` records on `backgroundJobBoard` as `completed`, which also prevented ticket release and terminal listener invocation.
3. OpenCode event payloads can send `event.properties.status` as a raw string (`'idle'`) rather than an object (`{ type: 'idle' }`), which evaluated to `undefined` when reading `status?.type` and skipped inactivity clearing. Additionally, `session.error` did not clean up active sessions.

#### Solution
- Wired `markTuiAgentInactive(record.taskID)` to `backgroundJobCoordinator.addTerminalOutcomeListener` and `addTerminalStateListener` in `src/index.ts`.
- Ensured `markTuiAgentInactive(sessionID)` deletes any buffered `pendingTuiBusySessions` entries.
- Transitioned foreground task records to `completed` in `src/hooks/task-session-manager/tool-execute-hooks.ts`.
- Normalized `status` extraction from `event.properties.status` to handle both `string` and `{ type: string }`, and cleared active sessions on `idle`, `completed`, `stopped`, `error`, `failed`, `session.deleted`, and `session.error`.
- Added unit tests in `src/index.test.ts` covering subagent terminal output, string status `'idle'`, and `session.error`.
- Fixed stdin TTY mocking in `src/cli/install.test.ts` so tests do not hang in interactive terminal environments.

#### Verification
- `bun run check:ci`: Passed (0 errors)
- `bun run typecheck`: Passed (0 errors)
- `bun test`: Passed (all 3,119 tests across 166 files)

PR #1215, opened 2026-09-17
fix(task-session-manager): route foreground completion through the terminal gate

dhaern:
## What happened

#1210 (clear inactive sidebar agents) added a block in `handleToolExecuteAfter` that marked a finished foreground task completed by calling `board.updateStatus({ state: 'completed' })`. #1209 (gate all terminal publications on runtime evidence), merged right after it, narrowed `updateStatus` to accept only `state: 'running'` and moved every terminal publication behind a gate-authorized `commitTerminal`. Current master fails because of that interaction:

- `tsc --noEmit`: `'"completed"' is not assignable to '"running"'` in `tool-execute-hooks.ts`
- `bun test src/index.test.ts`: `terminal subagent output clears active session in TUI snapshot` fails

At runtime the #1210 call is a silent no-op under the new contract: the foreground job stays `running` forever, and the terminal listener that clears inactive sidebar agents (the reason #1210 exists) never fires.

## The fix

Foreground completions now publish through the same terminal gate as everything else. The status branch of `handleToolExecuteAfter` already routed native outputs to `terminalGate.reconcile`; that signal now carries a `callIDConfirmed` flag, and the gate treats a synchronous run's native terminal return as terminal evidence on its own. When the host hands back `state: completed` for the exact callID it launched, the child cannot still be working, so the gate fast-paths the publication and the #1210 listener runs again. Background runs, and any output whose identity was parsed from the text instead of matched by callID, keep the full runtime-evidence discipline from #1209.

Three supporting changes came out of adversarial review:

1. Atomic task-output parsing (`task.ts`). Identity and state come from one host-authored opening: an anchored scan finds the opening tag, a quote-aware tokenizer walks complete `name="value"` pairs, and each tag kind keeps its own identity attribute (`task` reads `id`, `subagent` reads `sessionID`). Duplicated attributes invalidate the header, and a malformed opening (unterminated quote, inner `<`, exhausted input) rejects outright instead of falling back to textual markers hunted from the body. This closes every attribution vector the review found: quoted foreign task headers in both nesting directions, `id=`/`sessionID=`/`state=` quoted inside another attribute's value (the host interpolates `description` without escaping), a quoted `>` truncating the tag, duplicated identity attributes, and textual headers supplanted by quoted XML in the result body (which could read a failure of A as a success of quoted B).
2. Ambiguous busy handling in the gate. An untimestamped `busy`/`retry` status event now drops held terminal candidates, same as a timestamped busy. Without this, a candidate captured before unproven activity could still fast-path a publication that resumption had superseded.
3. Dead code removal. The legacy #1210 block is gone, and `releaseConcurrencyTask` is deleted from both the deps type and its only supply site, since nothing consumed it after the gate took over settlement.

## Numbers

| Scope | Files | Lines |
|---|---|---|
| Production | `task.ts`, `background-job-terminal-gate.ts`, `tool-execute-hooks.ts`, `index.ts` | +211 / −100 |
| Tests | `task.test.ts`, `background-job-terminal-gate.test.ts` | +181 / −42 |
| Total | 6 files | +392 / −142 |

The parser regressions live in one table-driven `test.each` with 21 rows (8 format controls plus the 13 adversarial cases, inputs written literally so no constructor can normalize the condition under test). The gate regressions are a two-row table: one row isolates the attribution barrier, the other holds a candidate behind a handoff and proves an ambiguous busy kills it.

## Validation

- `bun test`: 3162 pass, 0 fail (includes the master-failing #1210 regression test, now green)
- `tsc --noEmit`: clean, the master type error is gone
- `biome check src/`: clean
- Differential corpus: the pre-change parser and this one return identical results across 34 adversarial inputs (quote orientations, attribute permutations, unquoted values, empty values, case variants, unknown and truncated tags) on all five exports
- Red-green check: removing the `callIDConfirmed` requirement from the fast path makes the unattributed regression test fail, so the attribution barrier is actually under test

**Master currently fails typecheck and one test, and #1210's sidebar cleanup is dead on master. This should be merged as soon as reviewed.** It is a direct follow-up to #1209 and repairs its interaction with #1210.


dhaern:
@greptile-apps both P1 findings are fixed in 5781625f, please update your review.

**Textless completion becomes success** — the foreground fast path no longer invents a summary. The confirmed return now goes through `guardCompletedStatusText`, which downgrades a textless `completed` to `error` with the `COMPLETED_WITHOUT_TEXT_DIAGNOSTIC` summary (the empty-result incident rule the knowledge base points at), and only falls back to a state-descriptive summary for textless non-completed states, which that rule does not cover. The guard is now generic so callers keep their state narrowing. Regression test: `confirmed foreground textless completion publishes error, not invented success` runs with a non-terminal transcript so the native return is the only evidence and the guard decides the publication.

**Quoted `>` corrupts results** — `parseTaskResultFromOutput` now scans the subagent opening tag quote-aware, the same boundary rule as the header parser: the tag closes at the first `>` outside quoted attribute values. For the `description="Check a > b"` fixture the result is `done` again, and an unterminated tag yields no result instead of text leaking from inside the tag. Regression tests: `quoted gt in a subagent attribute does not leak into the result` and `unterminated subagent tag yields no result`.

Full validation after the fixes: 3165 tests pass, 0 fail; `tsc --noEmit` clean; `biome check src/` clean.

dhaern:
@alvinunreal heads up on an overlap between this PR and #1213 by @GoldJohnKing, since both touch the same breakage on master and the merge order decides who rebates.

What happened on master: #1210 added a direct `updateStatus({state:'completed'})` call for finished foreground tasks, and #1209 (merged right after) narrowed `updateStatus` to `state:'running'` only, so master fails typecheck and one test, and the #1210 sidebar cleanup is dead at runtime. Both PRs fix that same interaction, with different shapes.

Where they overlap:

- #1213 contains `d597f2d3` ("remove dead foreground updateStatus path"), which deletes the same 14-line block this PR removes. That part is literally identical.
- #1213 also rewrites the failing sidebar test (`c0eae739`) to expect the active session to survive the foreground terminal output and only clear after an idle `session.status` event, which is the opposite semantics of this PR's fast path: here, a foreground completion whose output the host attributed by exact callID publishes immediately, because a synchronous call that returned cannot still be working. The two tests assert opposite outcomes, so both cannot land as-is.
- There is also a content conflict in `background-job-terminal-gate.ts` and its test file (verified with `git merge-tree`), so whichever merges second needs a rebase either way.

Suggested order if you want both:

1. Merge this PR first. It is the minimal repair (6 files, +487/−157), master goes back to green (typecheck, the #1210 test, full suite), and it keeps #1210's intended UX: the sidebar clears as soon as the foreground task returns.
2. Rebase #1213 dropping `d597f2d3` and `c0eae739` (its other v2.0.5 migration work is unaffected by this PR), and adapt its two gate hardening commits (`ed8331cc`, `c72b27e0` on unhandled-rejection containment) on top. For the rewritten sidebar test, the idle-event assertion can stay as an additional case, but the immediate-clear expectation of the fast path needs to win.

Do I recommend keeping both? Yes. The overlap is three commits out of #1213's sixteen; the rest of its migration is complementary, and its rejection-containment fixes are real hardening this PR does not do. The only thing that genuinely needs a decision is the foreground publication semantics, and I'd argue for the immediate publication on confirmed synchronous returns, with the idle path remaining as the backstop for everything else.

GoldJohnKing:
Adopting this PR's semantics — the callID-confirmed fast path is the right synthesis of #1209's evidence gating and #1210's UX intent, and the parser hardening is welcome.

One correction for the record: this PR's failing `check:ci` gate is the **master-inherited lint debt** (21 `organizeImports` findings from #1209/#1210, repo-wide) — not a defect in this diff. #1213 already carries the fix (commit `1461eb24`) and will be rebased on top of this merge, restoring the lint gate for master.

Merging order: #1215 first (semantics owner), then #1213 rebases (dropping its duplicate dead-block removal `d597f2d3` and flipping its deferred-expectation TUI test to this PR's immediate-publication semantics; the idle-evidence case stays as an additional unconfirmed-attribution test). #1213 additionally brings the v2.0.5 plugin API migration and the fire-and-forget reconcile containment (`ed8331cc`/`c72b27e0`) that this PR doesn't have.

Issue #1216, opened 2026-09-17
Completed/reconciled child repeatedly fails resume with lifecycle ownership error on 2.2.21


stevekstevek:
Observed on OMOS 2.2.21 with OpenCode 1.18.31: a designer child completed with a terminal `finish: stop`, then repeatedly refused resume with “current generation is already owned by another lifecycle operation.” A status check between failures reported `reconciled (unconfirmed)`, `status_uncertain: true`, and no live status entry; messaging also failed because board state was reconciled, not running. Four resume attempts over about nine minutes returned the same ownership error. No restart or debugger was used, and the runtime lease owner was not captured. Included below is a sanitized, read-only DB/service/source evidence timeline. Related issues #1102/#1200 may be relevant, but a shared root cause is not established.

Possibly related to #1102 and #1200. Filing separately because #1102 is closed and this incident does not establish the provider-error path described in #1200.

<details>
<summary>Sanitized lifecycle evidence report</summary>

# OMOS lifecycle lock evidence — 2026-09-17

## Summary

A completed child session (`DESIGNER`) produced a terminal assistant response, but four later resume attempts from its parent (`PARENT`) were rejected by OMOS with the same lifecycle-ownership error. Between the first and second rejection, `task_status` reported the child as `reconciled (unconfirmed)` with `status_uncertain: true` and no live status entry; `task_message` then rejected the child because its board state was `reconciled`, not `running`.

This captures a state inconsistency visible to the caller. It does **not** identify the runtime lease owner or prove a root cause.

## Expected versus observed

**Expected:** after the child's terminal response is acknowledged/reconciled, either:

1. the child becomes safely reusable and a resume acquires a relaunch lease; or
2. the child remains non-reusable with a clear active/unreconciled status that identifies why.

**Observed:** the child had a terminal `finish: stop`, was later reported as reconciled but uncertain, could not accept a message because it was not running, and repeatedly refused resume because its current generation was allegedly owned by another lifecycle operation.

## Environment captured after the incident

| Item | Value | Evidence/qualification |
|---|---|---|
| OpenCode installed | `1.18.31` | `opencode --version` |
| OMOS installed/pinned | `2.2.21` | package manifest and plugin config |
| Bun | `1.3.14` | `bun --version`; OMOS manifest also declares Bun 1.3.14 |
| Node | `22.22.3` | `node --version` |
| OpenCode service | active/running | selected `systemctl --user show` properties only |
[...]

PR #1217, opened 2026-09-18
fix(task-session-manager): release the notification lease when its local wait ends

dhaern:
Fixes the leak mechanism reported in #1216. I did not mark it as closing the issue because the reporter ran 2.2.21 and a backport decision belongs to the maintainer.

## What breaks

When a child task goes terminal, `revived-run-tracker` notifies the parent session. The send acquires a `terminal-notification` lease so that retries and late settlements cannot double-deliver. The transport call races against a 10 second local timeout, but the lease was only released when the transport promise settled. A transport that never settles (the v1 SDK queues `prompt_async` and offers no portable cancellation) kept the lease forever. The board refuses to hand out a relaunch lease while any lease exists for the task, so every later `acquireRelaunchLease` failed and the child became permanently unresumable with "current generation is already owned by another lifecycle operation". That matches #1216 exactly: four resume attempts over nine minutes, all refused, record stuck at `reconciled (unconfirmed)` with `status_uncertain: true`.

A differential reproduction fails identically on v2.2.21 and on current master, so the gate rework in #1209 does not cover this path: #1209 decides who may publish a terminal state, this bug lives in delivery, after the record has already gone terminal correctly. The transport code predates both (introduced in `693aa4b7`, refined in `5ebfa44c`).

## The fix

`notifyParent` now owns the lease for the whole attempt and releases it in its `finally`, unconditionally. `awaitNotificationTransport` loses the board, the lease, and the `settled`/`timedOut` bookkeeping; it keeps one job: run the operation, classify the response, and report acceptance through a callback that fires whether the race was won or lost. An `{ error }` envelope never counts as acceptance.

Acceptance (`notification.sent`) becomes a monotonic latch per publication: any successful attempt of the same publication can set it, a late success from an attempt whose timeout already fired still counts, and the callback can modify nothing except the captured publication object and its retry timer. A shared `isCurrentNotification` check replaces the scattered guards, and a revalidation inside the deferred operation closes the acquire-to-send window, so a stale publication never reaches `promptAsync`.

The retry policy is untouched: same budget semantics (`maxNotificationRetries: 0` still means initial send only), same refund when a lease cannot be acquired due to contention.

The failure domain changes from "task permanently unresumable" to "a duplicate terminal message is possible in a narrow race". For a lifecycle plugin that is the right trade: a duplicated informational message costs a turn, a leaked lease costs the child. Delivery stays best effort and the docs say so plainly.

## Numbers

| Scope | Added | Removed |
|---|---:|---:|
| Production (`revived-run-tracker.ts`) | 77 | 113 |
| Tests (4 files) | 678 | 255 |
| Docs (`background-orchestration.md`) | 12 | 0 |

Production is net negative (-36) because the old timeout machinery is larger than the bounded-wait version that replaces it. Suite: 3222 tests, 0 failing. Typecheck, lint, and build clean.

## Test coverage

- Perpetually pending transport with budgets 0, 1, and 3: the task is resumable again right after the timeout, before any retry executes
- Release on all four outcomes (success, sync throw, async rejection, error envelope); only success marks acceptance
- Both microtask orderings of timer vs settlement; late success cancels a pending retry; late failure keeps it
- A times out, B sends, A settles late, B fails: no third send, and A's settlement never touches B's lease
- Run substitution, discard, dispose, and new generations: stale responses cannot affect the new owner
- Pre-send invalidation: `promptAsync` is never invoked once the publication is stale
- Board symmetry, parametrized across all four lease kinds: acquire excludes incompatible acquisitions, releasing the own token re-allows acquisition, double release is a no-op, a stale release cannot remove a newer lease, and terminal revision invalidation invalidates the send permission without blocking removal of the token's own lease
- `task_revive` integration: revive refused while a notification holds the lease, allowed after the timeout, old settlements cannot touch the new generation
- Gate `A busy B` scenario extended: B progresses after A's timeout without waiting for A's settlement

## Validated live

Ran on a real installation (OpenCode 1.18.31, plugin built from this branch). A child completed, its notification was delivered and acknowledged, the record reconciled, and a resume of the finished child was accepted on the first try, ran, completed, and reconciled again as a new generation. A second child was cancelled mid-flight and transitioned to a clean terminal record. No ownership errors appeared at any point.

## Related

- #1209 reworked terminal publication; this patch covers delivery. Different subsystems, and the live validation exercised both together.
- Three sibling asymmetries surfaced during the audit, left alone here because each acts on the child and needs its own admission analysis: `message` holds its lease after timeout until settlement (`task-message.ts:162-223`), `relaunch` awaits `promptAsync` without a timeout (`task-revive.ts:168-219`), and `cancellation` holds operation ownership while pending (`cancel-task.ts:437-468`).


dhaern:
Rebased onto `3ed5490a` (#1213 merged) to clear the conflict in `task-revive.test.ts`. Resolution: both isolation mechanisms coexist. #1213's `mock.module` at module scope stays, and this PR's per-test `spyOn(getClient)` in `beforeEach` layers on top of it; `mock.restore()` returns to the module mock after each test. The contamination pair (`task-message` + `task-revive` in one process) passes 33/0, the four touched files together 220/0, and the full suite is 3226 pass, 0 fail on the rebased branch.

Heads-up for whoever looks at CI: `bun run typecheck` currently fails on pristine `master` too (`src/v2/mirror-conformance.ts:123` and `:159`, inherited from #1213). This PR does not touch that file; the failure is not introduced here.

PR #1221, opened 2026-09-18
fix(task-revive): don't block revive on v2 hosts with no status map

kumar-shivang:

#### Problem
On v2 hosts, `task_revive` can never succeed. Before sending the revive prompt it verifies liveness via `getRuntimeSessionStatusSnapshot`, which calls `client.session.status(...)`. v2's `buildPluginInput` shim in `src/v2/client-shim.ts` intentionally omits `session.status` (v2 has no equivalent of the v1 live status map — it's event-driven only), so that call always throws `client.session.status is not a function`. `task_revive` treated any such error as an "unverifiable" live map and permanently refused with:

> Task X could not be verified against the live session map (...); the revive prompt was NOT sent. Retry task_revive.

From the user's side this shows up as a retained background task that `task_revive` refuses to touch, as if it were still active, even though it's idle.

`cancel-task.ts`'s own verifier already treats "no `session.status` function" as a capability signal and falls back to host-info-based confirmation (`session.get`'s outcome/idle timestamp) instead of failing. `task_revive`'s inline live-check never got the same treatment.

#### Solution
- Only run the live status-map check in `task_revive` when `session.status` is actually a function (same capability signal `cancel-task.ts` uses), skipping it on v2 hosts.
- On v2, liveness is already covered by the `freshLiveActivity`/`lastLiveBusyAt` check just above it, which stays current via `event-adapter.ts` synthesizing `session.status`/`session.idle` from `session.execution.*` into the board.
- Added a regression test simulating a v2 host (no `session.status` on the client) reviving a stopped session; it fails with the exact `client.session.status is not a function` error on `master` and passes with the fix.

#### Verification
- `bun run typecheck`: Passed (0 errors)
- `bun run lint`: Passed (no issues on changed files)
- `bun test`: 3242/3243 pass; the one pre-existing failure (`foreground-fallback/index.test.ts`, a dispose/backoff timing test) reproduces identically on unmodified `master` and is unrelated to this change.


kumar-shivang:
I tested it live locally, now I am able to resume the sessions.

PR #1222, opened 2026-09-18
fix(task-revive): bound relaunch deadlines and close admission ownership gaps

dhaern:
## Problem

An adversarial admission analysis of the task_revive relaunch path (cancel a running child, then relaunch it) found three failure windows, all reachable without any host misbehavior:

1. Baseline capture awaited network I/O with no deadline. A hung session.messages call left the tool stuck indefinitely, before any cancel or prompt had been sent. Pure availability loss.
2. The admission handshake accepted the new generation before the owner was registered. If the admission response was lost, the child kept running with no owner recorded and nothing to reconcile it.
3. The pre-send state read was not revalidated after the awaits in flight, so a record dropped mid-flow could still receive a fresh prompt.

The path shared the general weakness of the notification transport fixed in #1217: long-lived operations with no local bound.

## What changed

- Baseline capture is bounded (5s default, configurable through TaskReviveToolOptions). A hung read now fails the revive instead of parking it.
- The admission wait is bounded locally (10s) with ownership transferred to the tracker on admission. The local timer emits a dedicated sentinel, ReviveAdmissionDeadlineError, and classification reads only the timer, so a late-arriving admission can never be misreported as a deadline failure. The first cut classified from mutable post-await state and a race could return "deadline exceeded" while the new generation was already running; the audit caught it and the sentinel closed it.
- Supersession is tested as a contract: a stale candidate rejected during the observation phase must not corrupt the newer generation's record.
- The observation phase runs after the preparation lease is released, so observation no longer holds the relaunch lease.
- registerLaunch fences, delivery 'queue' semantics and the terminal-gate discipline from #1209/#1215 are untouched.

## Lines

Production +143/-85 in src/tools/task-revive.ts, tests +383/-82, docs/tools.md +14, schema +8. Net +548/-167.

## Verification

- Branch tests: 102 pass across the task-revive and revived-run-tracker suites, 0 failures
- Composed family branch (all four sibling branches merged together): 3353 pass, 0 failures excluding one pre-existing master failure (foreground-fallback dispose/backoff timing, reproduces on unmodified master, comes from #1218)
- Two-round adversarial audit: r1 rejected with the deadline-vs-admission race and supersession contract gaps; r2 passed after the sentinel classification and deterministic interleaving tests
- Validated live on a real host from the composed build: revive of a completed background session, clean cancel-in-flight, and terminal publication through the strict-attribution gate

## Family

This is the second of four sibling PRs from the same admission analysis:

1. #1217, terminal-notification lease release (merged)
2. This one, relaunch phases
3. Cancellation verification budget (follows)
4. Message quarantine contracts plus v2 shim semantics (follows)

Each PR is independent and mergeable in any order. This one and the upcoming shim PR both touch src/tools/task-revive.ts in different regions; merging this one first keeps the other rebase trivial.


dhaern:
@greptile-apps valid finding, fixed in 704d2922.

What the fix does, following "deletion wins":

- When a session.deleted (or clearParent) removes the record while admission is pending and the late admission response succeeds, the captured relaunch lease now transfers to a compensation owner that aborts exactly the captured taskID. No record resurrection, no tombstone clearing, no prompt re-send.
- The lease stays held while the compensating abort is in flight and is only released after the abort settles and a bounded live status read (session.status with a timeout budget) verifies quiescence. Abort failures, invalid reads and busy/retry keep the exclusion with an explicit unconfirmed-compensation diagnostic instead of claiming the child stopped.
- A caller still waiting gets an explicit "admission accepted but invalidated; compensation initiated" error instead of an admission timeout. Callers that already received admission_unknown keep that result. The ReviveAdmissionDeadlineError sentinel and its same-tick race are untouched, and registerLaunch fences plus background-job-board.ts are unchanged.

The old dropped test variant asserted exactly the abandonment you flagged; it now asserts the compensation contract. A new integration test drives the real session.deleted lifecycle callback for child and parent deletion with a sibling present, and the full matrix (revoked ownership, pending/failed/uncertain compensation, invalidation before the deadline) is covered with deferred promises, no sleeps. One honest limitation, documented in docs/tools.md: an unconfirmed compensation can retain the exclusion for that taskID for the process lifetime; there is no lease-inspection tool yet.

Could you update the review to reflect the fix? Thanks.

PR #1224, opened 2026-09-18
test(task-message): pin the deferred-write quarantine contracts

dhaern:
## Problem

The task_message quarantine (a deferred write to a paused child retains its lease so a late message cannot corrupt the context of a reused generation) is an intentional contract, but nothing in the suite named it as one. A future refactor could weaken it as dead weight, which is exactly how the transcript-corruption class of bugs returns.

## What changed

- Named quarantine-contract tests in src/tools/task-message.test.ts: deferred writes retain the lease, reused generations refuse cross-generation writes, and the refusal diagnostics stay stable.
- docs/tools.md documents the contract and its rationale (finding M1 context; the v2 noReply/delivery translation lives in the companion shim PR).
- Zero production lines: behavior is already correct and stays that way.

## Lines

docs/tools.md +22, tests +106/-36, production +0/-0.

## Verification

- Focused tests 21 pass, 0 failures; full family suite on the composed branch: 3353 pass excluding the pre-existing master failure (#1218, reproduces on unmodified master)
- Audited: verdict APTO on first round

## Family

Fourth of four sibling PRs from the same admission analysis (#1217 merged, #1222 merged, #1223 cancellation, this one). The v2 shim PR (noReply semantics + gate attribution) follows separately and depends on none of these.


dhaern:
@greptile-apps The P2 is addressed in 984708ed following your suggested option: the ignored duplicate transport settlements are removed, and after acquiring the replacement the test now explicitly calls `board.releaseLease(lease)` with the stale lease, asserting it returns `false`, retires nothing, and leaves the replacement valid with the board record unchanged. (One note: the release intentionally passes through the existing release spy, so `retirements()` counting is kept only for the production settlement assertion.) 21/21 focused tests pass, `check:ci` and typecheck clean. Branch also rebased onto current master (b943be86) after #1223 landed. Please update your review.

GoldJohnKing:
**Merged.** The quarantine contracts pinned here become guardrails for the follow-up wake work: the planned terminal-publication wake (one-shot, throttled, `delivery:'queue'`, `modelSelection:'inherit'`) and the reopen corrective notices deliver through the same deferred-write/lease semantics these tests pin — any quarantine regression now fails this suite before it can reach the wake paths.

PR #1225, opened 2026-09-18
fix(v2): honest noReply/revive semantics + strict outcome attribution

dhaern:
## Problem

Three v2-host defects, found by adversarial admission analysis of the task-control family:

1. **R4 — revive on v2**: v2 hosts omit session.status, so the revive refused as unverifiable. #1221 unblocked it by skipping the live-map check when the capability is absent, which revives with zero verification: promptAsync on a busy v2 session silently steers the in-flight run. This PR supersedes that approach with real verification.
2. **M1 — noReply lost in translation**: the v2 shim translated noReply into delivery:'steer', which can inject into an in-flight run and restart it — breaking task_message's no-restart guarantee on v2.
3. **Gate misattribution**: the terminal gate read session.get's outcome field without checking time.idle against the current attempt, so a historical failed/interrupted outcome of the previous generation published the new generation as failed (state: error, statusUncertain: false) in both v1 and v2.

## What changed

- task_revive capability precheck before any destructive abort: requires session.status (live-map verification) or experimental_v2.waitForSessionIdle (bounded idle wait, 5s default); a host with neither refuses clearly before any abort or prompt.
- Strict temporal attribution in the terminal gate: validates hostOutcomeClock and only accepts an outcome if lowerBound < time.idle <= readCompletedAt; non-attributable outcomes degrade to unknown with status_uncertain and never publish.
- The pre-admission frontier (admissionStartedAt) is captured before the prompt send and flows into registerLaunch.now and the tracker's attemptStartedAt (immutable across handoffs), so fast-completing G2 runs are not discarded by the attribution window.
- task_message: noReply translates to delivery:'queue' + resume:false; persisted per-session agent selection is inherited without global mutation; unrepresentable overrides reject instead of misbehaving.
- Revive output renders status_uncertain with the observation diagnostic.

## Lines

~+295/-90 production (task-revive.ts, background-job-terminal-gate.ts, client-shim.ts, types.ts, task-message.ts), ~+480 tests (incl. task-control-conformance suite), ~+58 docs. Fresh numbers in the diff against current master (includes #1222's merged flow).

## Verification

- Full suite: 3361 tests, 3360 pass; the single failure is the pre-existing master flake (foreground-fallback dispose/backoff from #1218, reproduces on unmodified master)
- Two-round adversarial audit (r1 found the stale-outcome attribution P1; r2 APTO) with independent reproductions on v1 and v2: historical outcomes keep G2 uncertain with zero publications, legitimate G2 outcomes publish exactly once
- Live validation of the revive/cancel/attribution paths from the composed family build

## Family

Last of four sibling PRs (#1217 merged, #1222 merged, #1223 cancellation, #1224 message contracts). This one supersedes #1221's skip-verification contract: its regression test is adapted to verify through the wait capability instead of skipping, and the capability-unavailable refusal is pinned in the conformance suite.


dhaern:
@greptile-apps The P1 is addressed in the head commit, following your first suggested direction — preserve the host event timestamp through the adapter — with one correction to the finding's premise: the v2 host **does** timestamp every event via `created` on the envelope (Clock.currentTimeMillis; the projector derives `time.idle` from the terminal event's `created`), but `mapV2EventToV1` dropped it when synthesizing the busy.

The synthesized `session.status busy` now carries `properties.activityAt = event.created` (finite, non-negative, present only when valid), which the router's existing `eventActivityAt` already forwards as `observedAt`. Consequences, all pinned by tests:

- **Short run with delayed queued start**: `start(10) < idle(50) < receipt(60)` — the boundary is the host time 10, not the receipt time, so the legitimate outcome attributes and publishes.
- **Independent resume**: a genuinely new run's `started(55) > previous idle(50)` still fences the historical outcome — zero publications, no weakening of the strict attribution window.
- Two contract clarifications are now explicitly pinned: an attributable `succeeded` without transcript evidence never publishes (attribution is separate from result availability), and an unusable/absent `created` synthesizes no receipt time.

`eventFromRead`'s lower bound and the gate's temporal checks are untouched. Gates clean, full suite 3384 passing plus the known pre-existing master flake (#1218). Please update your review.

GoldJohnKing:
**Merged** — after resolving a docs-only conflict with #1224 in `docs/tools.md` (both sides kept: the lease-quarantine contract, followed by this PR's v2 delivery semantics). Pre-merge verification in a clean worktree: `typecheck` and `check:ci` clean, full suite 3386/3387 pass — the single failure is the pre-existing `ForegroundFallbackManager dispose > dispose during retry backoff` timing flake from #1218 (passes on isolated rerun on this branch and on unmodified master).

**Follow-up work building on this PR** — direction and problem statements (one item was re-scoped today by independent review; mechanisms marked TBD will be pinned by instrumented reproduction before implementation):

1. **v2 terminal-evidence starvation (re-scoped).** Verified live twice on OpenCode 2.0.8: background children completed with `Session.Info.idle_outcome='succeeded'` already committed at idle time, yet the gate published nothing — board jobs stuck at `running, status uncertain` with the evidence-unavailable diagnostic after retry exhaustion. A third identical build terminalized normally in another location's plugin instance, so the failure is race/shape-sensitive. Independent review established that the initially suspected mechanism (a synthesized `session.status{idle}` preempting `pendingRuntime`) does not exist — the event router folds both synthesized idles to `origin:'session.idle'` (idleCandidate only). The real failure sits between the host-outcome read and attribution/publish (candidates: in-flight read-slot joining, outcome read visibility for child sessions, retry-exhaustion dead end with no re-arm). Sequence: observability first (build info; INFO logs on publication, wake evaluation, gate reads), then an adapter-driven reproduction through `mapV2EventToV1` + the real event hook against a shim-shaped client, then a fix aimed at the reproduced mechanism inside this PR's attribution window (`lowerBound < time.idle ≤ readCompletedAt`, including the `hostOutcomeClock` wiring).
2. **Bounded terminal-publication wake.** The native subagent notifier fires once per original job (first completion); self-continuing children (e.g. resumed by their own background-shell notification) leave idle parents unnotified for up to the 5-minute periodic interval. Plan: one-shot throttled wake (`delivery:'queue'`, `modelSelection:'inherit'`) on the existing wake gate. Known integration detail: `classifyChildrenSnapshot` returns `no-work` for the post-publication state unless a recovery/publication flag is threaded through, mirroring `triggerStoppedJobRecovery`.
3. **Reopen-after-reconcile + corrective notice.** Regression coverage for the self-continuation pattern (busy events driven with host envelope timestamps, per cdf9ec00), plus a cache-safe trailing corrective notice when a reconciled job reopens (concept adopted from #1066).
4. **Exhaustion re-arm + rehydrate sweep.** A late-attributable outcome arriving after evidence-unavailable must still terminalize; pre-existing stuck jobs must clear via the rehydrate probe on the parent's next activity — both pinned by tests.
5. **Read-tool permission-rule derivation for read-only child agents.** `explorer` children currently derive zero exact-match rules → permission prompts → input-wait, which also suppresses wakes.

#1221's superseded contract and this PR's strict attribution are the baseline and will not be relaxed. #1066 will be closed as superseded, with its corrective-notice and config-consolidation concepts adopted. A full implementation plan accompanies the first follow-up PR.

PR #1226, opened 2026-09-18
feat(task-control): provisional provenance for unattributed child placeholders

dhaern:
Unattributed child sessions (for example auxiliary sessions other plugins open through `client.session.create`) used to land in the job board as permanent `unattributed <agent> task` entries. They stayed visible forever, kept the periodic wake alive, and blocked `wait_for_user` for work nobody asked for.

The fix marks those registrations `provisional` at birth. A provisional entry still resolves (`task_status`, `task_result`) but stays out of the board listing, out of `hasRunning` and `hasTerminalUnreconciled`, and never fires wake listeners. It becomes a normal tracked task when real attribution arrives, through any of four paths: `registerLaunch`, cross-board adoption, `preserveRun` (including entries that already reached a terminal state; those get promoted without touching their recorded state), and ID-only task outputs through a new `promoteProvisional(taskID, parentSessionID)` board call. The parent fence comes from the pending itself, so it is always present, and no run state is invented at any promotion path.

That set of rules comes out of an audit round that found four gaps: terminal placeholders were never promoted, ID-only outputs never promoted, the two predicates still counted provisionals, and adoption skipped existing placeholders. All four are covered by tests that fail if you revert the matching fix.

Two notes:

`hasRunningJobs()` still counts provisionals on purpose. Its only consumer is the runtime status reconciliation gate, and excluding them would stop status polling for a board holding only placeholders, leaving those stuck `running` forever.

This branch also carries a small foreground-fallback commit (`85ccd9bc`). Bun aborts upstreams that never send response headers with `Provider response headers timed out after Nms`, and that phrase matched no failover pattern, so a held connection killed the task as an opaque error instead of walking the model chain. We ran into it live. If you'd rather have it in its own PR, say so and I'll split it.

Size: +149 lines of production code, +372 of tests, +14 of docs. Tests: 3404 pass, 1 fail (the known `#1218` flake; the file passes alone). `bun run check:ci` and `tsc --noEmit` clean.


dhaern:
Heads-up for reviewers: this branch also carries `85ccd9bc` (foreground-fallback).

It classifies Bun's client-side `Provider response headers timed out after Nms` aborts as failover errors (`TRANSPORT_MESSAGE_PATTERNS`). We hit this live: a proxy layer holding a request instead of answering left clients with an opaque ~300s transport error while a perfectly good fallback model sat unused in the agent's configured chain — the phrase matched no failover pattern, so the chain never walked.

The phrase is unanchored on purpose (the real message carries a `Provider` prefix); it is a narrow technical timeout phrase, and the suite pins both the match and adjacent generic wordings as negatives. Happy to split it into its own PR if you prefer it separate.

dhaern:
@greptile-apps both findings are fixed and pushed.

Recovery wake: `preserveRun`'s terminal branch and `promoteProvisional` now re-notify the terminal listeners after clearing `provisional`. The wake fires exactly once per record because the stop-time notification skips provisional records and promotion is a one-time transition, so an attributed child that stopped while hidden now produces the recovery wake that was previously swallowed.

Launch metadata: the ID-only path now passes the owning pending's metadata (agent, label, objective, background) into `promoteProvisional`, and after a real promotion the hook binds the concurrency ticket, arms `backgroundJobSupervisor.onLaunch` under the same `exactCallConfirmed` guard the parsed paths use, and tracks the child in `pendingManagedTaskIds`. Pendings with unresolved identity still promote under the parent fence but paint nothing and arm nothing, per the documented identity rule.

New tests: terminal-promotion wake re-fire at board level (terminal and running variants), metadata painting, and hook-level promotion with supervision armed. Suite is at 3404 pass with the known `#1218` flake being the only failure.

PR #1229, opened 2026-09-19
feat(tui): reusable history marker and row navigation for finished subagents

dhaern:
## Motivation

#1197 made sidebar rows open the subagent that is currently running. Once that subagent finished, the row went inert: reaching the last session of a category meant scrolling back through the chat to find it, with no shortcut on the sidebar.

## What it does

- A row whose category has finished history is clickable anywhere on the line. When the category is idle, clicking opens the most recently finished subagent, ordered by last use, so both when it finished and when it was last reused count.
- An emerald ✦ marker beside the agent name signals that history exists. It appears the moment the child session reaches a terminal state; it does not wait for the parent to acknowledge the result. While any task of the category is running the marker hides, and the running behaviour from #1197 keeps priority.
- The marker is a static text node with no mouse handling of its own. Hover highlight and navigation belong to the whole row, so there is a single interaction surface instead of a tiny nested click target.
- Selection state lives in the background job board and reaches the TUI through a small reactive projection into tui-state, since the TUI runs in a separate process from the plugin server. The projection rebuilds only on board mutations, so the TUI never polls.
- Aborted sessions (cancelled or errored) join the history when their transcript is still reachable; stop-retained sessions without one do not.

## Also in this branch

`scripts/cache-smoke.ts` left an orphaned `opencode serve` process behind when it failed or was interrupted. That stray server later blocked a Magic Context schema migration, because its fail-closed guard treats any live server as a concurrent instance. The serve subprocess now runs detached in its own process group, SIGINT, SIGTERM and SIGHUP tear the whole group down, and the failure path cannot re-enter itself. Separate commit, since it is unrelated to the sidebar.

## Validation

- New tests cover per-category latest selection (including the nested-parent case, resolved by keeping the highest last-used time), projection rebuilds on mutation, the restart sweep clearing stale sections, marker exclusivity with running tasks, and row click navigation for idle rows with history.
- Full suite: 3411 pass, 1 pre-existing failure (`foreground-fallback` dispose during retry backoff, also fails on current master). `bun run check:ci` clean.

Delta: +836/−49 production, +465/−0 tests, +48/−8 scripts.


dhaern:
@greptile-apps both findings are fixed in 7115f454 and 4ee25e93.

For the P1, you were right and the bug was worse than an inconvenience: cleanup could never deliver a signal, so the original orphan problem was still there. stopSpawnedServe now takes the captured serve record as a parameter instead of reading the module global that cleanup had already cleared. The script has no test harness, so verification was inspection plus the full biome check.

For the P2, latestReconciledByParentAgent now derives every parent-agent selection in one pass over the board instead of scanning once per parent. The comparison logic moved into a shared upsertSidebarSelection helper, so both projections keep a single implementation of the recency and tiebreak rules. Existing tests for both projections pass unchanged (169 focused, full suite 3411 with only the pre-existing foreground-fallback flake from master).

Could you re-review when you get a chance?

PR #1231, opened 2026-09-19
fix(foreground-fallback): promote foreground task() waiters to background before abort

dhaern:
## Motivation

When a subagent launched by a synchronous `task()` call hits a model
failover, the orchestrator receives `Task cancelled` while the child
keeps running on the fallback model as a headless, untracked session.
Its `tool.execute.after` hook never fires, so the early-registered
board placeholder stays provisional and invisible.

## Root cause

The fallback path aborts the failing run before replaying on the next
model. That abort settles the child's background job as `cancelled`,
and the waiting `task()` tool resolves its race as cancelled and fails
(opencode `task.ts` only kills the session on Effect interrupts, not on
a cancelled result). The replay then continues on a session nobody
tracks.

## Fix

Immediately before each fallback abort (the failover abort and the
busy-session re-prompt abort), promote the parent's foreground job via
`POST /experimental/session/{sessionID}/background`. The host resolves
the parent's wait through `backgroundResult`, so the tool completes
normally, its after-hook attributes the child, and the host's own
`onPromote` path delivers the superseded run's result.

v2 plugin inputs carry no `_client`, so the same request falls back to
a plain fetch against the input's `serverUrl`. A reload that disposes
the generation mid-promotion now fences the abort instead of cancelling
live work through a stale client, on both abort sites.

By endpoint design this promotes every foreground job of the parent, so
healthy parallel waiters also detach to background in the same failover
window instead of dying with the broken account. Promoted runs acquire
no new concurrency admission: they were already admitted as foreground
calls, and the promotion reclassifies the same run for tracking and
delivery rather than launching a new one.

Fail-soft throughout: an unknown parent, a host without the endpoint, a
2s promotion timeout, or a transport error all degrade to the previous
behavior. `task()` launch outputs that report `metadata.background:
true` mark the board record as background, even when call identity was
unresolved.

Stacked on #1232: the branch carries that board fix so CI runs green
(master currently fails a board visibility test this PR does not
touch).

## Validation

- foreground-fallback: 124/124 — promotion ordering before both abort
  sites, parent targeting, v2 fetch fallback, dispose-during-promotion
  fencing, no-promotion without a known parent, and fail-soft on
  transport errors and timeout.
- task-session-manager: 173/173 — a launch output with
  `metadata.background: true` registers the board record as background,
  including the unresolved-identity path.
- typecheck, lint, and the full suite green apart from one known timing
  flake that passes in isolation.


dhaern:
All three findings addressed in the new head:

1. **V2 Skips Waiter Promotion** — fixed. When the client exposes no
   `_client` (v2 plugin inputs), the promotion falls back to a plain
   `fetch` against the input's `serverUrl`: same host, same endpoint,
   same fail-soft envelope. Covered by a test driving the fallback path
   end-to-end and asserting the target URL and the promote → abort →
   replay ordering.

2. **Reload Can Trigger Stale Abort** — fixed. Both abort sites now
   check `abandonedByDispose` after the promotion await and before the
   abort, so a generation disposed mid-promotion can no longer cancel
   live work through its stale client. Covered by a test that disposes
   the manager inside the promotion request and asserts neither the
   abort nor the replay runs.

3. **Promotion Misses Background Ownership** — two parts:
   - Board registration: fixed. Host-reported `metadata.background:
     true` is now authoritative for the child that output describes, so
     the record registers as background even when call identity was
     unresolved.
   - Concurrency admission: intentionally unchanged. A promoted run is
     a reclassification of the same already-admitted foreground call —
     the promotion does not launch a new child, so it does not owe a
     new admission slot. Acquiring a ticket post-hoc would bind a
     possibly never-ready ticket against a saturated pool. Documented
     in the commit and PR description.

CI was red for a reason unrelated to this diff: a master regression in
the board tests, fixed by #1232 — this branch now stacks on it, so
checks run green. @greptile-apps could you re-review?

dhaern:
Concurrency ownership addressed: a promoted run now acquires the same ticket a native background launch holds at promotion time — mirroring the nested-task exemption and without blocking on capacity, since the child is already running. Registration binds the ticket and the terminal path releases it, so promoted runs stay inside the configured limits. The formatter drift that broke CI is also fixed. @greptile-apps could you re-review?

dhaern:
Fixed: the promotion-time acquisition is fire-and-forget accounting (nobody awaits `ticket.ready` there), so the queue-cancelled rejection now gets an explicit no-op handler — a granted, queued, or released ticket all behave as before. @greptile-apps could you re-review?

PR #1232, opened 2026-09-19
fix(board): keep reconciled provisional placeholders out of reusable sessions

dhaern:
## Motivation

`background-job-board.test.ts > keeps a completed placeholder hidden
before and after acknowledgement` fails on current master: after
`markReconciled`, a completed placeholder surfaces under
"Reusable Sessions" even though it was never attributed.

## Root cause

`listReusable` predates the provisional provenance contract (#1226).
`formatForPromptWithMetadata` applies the provisional exclusion to the
active and retained sections but reads the reusable section straight
from `listReusable`, so a reconciled completed placeholder leaks into
it and the sentinel stops returning `undefined` for parents of
unacknowledged placeholders.

## Fix

Apply the same provisional filter to the reusable section. `resolve()`
and `get()` remain untouched — placeholders stay resolvable, just not
advertised.

## Validation

- background-job-board: 108/108, both variants of the failing test
  included.
- Full suite: the regression disappears; no other behavior change.


dhaern:
Fixed — same gap in the sidebar projection: `isSidebarHistory` now excludes provisional records, so both sidebar queries (the per-parent dot selection and the global sweep) apply the same exclusion as the prompt's reusable section. A completed placeholder that reaches `markReconciled` no longer shows up there, while an attributed sibling still selects normally — covered by a regression test in the sidebar describe.

PR #1233, opened 2026-09-19
fix(multiplexer): probe /session/status with the child project directory

### LINKED ISSUE #1214: [Bug]: Multiplexer panes never spawn, spawn in the wrong tab, or are disabled entirely under a shared server (opencode serve + attached clients)
## Summary

With a single long-lived OpenCode server (`opencode serve`) shared by several
attached TUI clients (`opencode attach --dir <project>`), the plugin's
multiplexer integration fails in three ways:

1. **No pane is ever created** — the readiness gate is stuck forever.
2. **Panes spawn next to the wrong client** — beside the server's pane or
   wherever focus happens to be, not beside the parent session.
3. **No pane at all** for many `multiplexer.type` / environment combinations.

## Symptom 1 — no pane is ever created

The log repeats for the child's whole lifetime, and never self-heals:

```
[multiplexer-session-manager] child session not ready, skipping spawn
[multiplexer-session-manager] child session not ready, skipping respawn
```

Resolved by explicitly adding the session's `directory` to the
`/session/status` probe — without it the endpoint answers for the server's
own default project directory, so the child session never appears.

## Symptom 2 — panes spawn next to the wrong client

| Server cwd | Client A cwd | Client B cwd | Subagent dispatched by | Where the pane was created |
| --- | --- | --- | --- | --- |
| /proj-1 | /proj-1 | /proj-2 | B | In A's TUI — not in B's |
| /proj-1 | /proj-1 | /proj-2 | A | In A's TUI (correct only by coincidence) |

## Symptom 3 — no pane at all

The subagent still runs headless; only the dedicated pane is missing:

| `multiplexer.type` | Server started in | Client attached from | Observed |
| --- | --- | --- | --- |
| `tmux` |  tmux | tmux | OK |
| `tmux` | anywhere except tmux | tmux |  No pane is ever created|
| `auto` | herdr | tmux | No pane is ever created |
| `auto` | outside every multiplexer | tmux | No pane is ever created — pane management is disabled altogether |

dhaern:
Fixes #1214.

In shared `opencode serve` setups, with one server and panes for projects in different directories, every multiplexer spawn logged "child session not ready, skipping spawn" after the 2s readiness deadline while the child session was alive, so the pane never opened.

The readiness gate polled `GET /session/status` without any directory information. The server resolves status requests without a directory to its own process cwd, so sessions spawned in other project directories never appeared in the payload the gate polls.

The probe now carries the child's directory as the `x-opencode-directory` header, pre-encoded with `encodeURIComponent`. This is the same contract the official SDK applies to every request: the workspace-routing middleware reads the header raw and the instance context decodes it exactly once, so percent-bearing and non-ASCII paths resolve correctly, and since the header is not part of any endpoint schema, older servers ignore it instead of rejecting the request. The query parameter had both problems at once (400 on pre-May 2026 servers, double decode on percent paths); neither has a client-side workaround, which is why the probe uses the header.

Both spawn paths of the session manager provide the directory: the fresh-spawn local on first attach, and the known-session record on respawn. The cmux gate takes the same directory it already received and reuses the shared readiness parser. Adapters with a custom `checkSessionReady` keep working: the headers travel as an optional argument they already receive.

Diff: +50/−32 in production code, +179/−13 in tests.

Known follow-up, out of scope here: the idle-close pollers (`fetchSessionStatuses`, cmux session lifecycle) still poll without a directory; the poll fallback stays blind to sessions in other directories, while the primary `session.idle` events cover them. The query `directory` parameter also keeps a double-decode bug in the core for its own users (searchParams.get plus the instance-context decode); worth an upstream report.

Validation: multiplexer suite 270/270, including tests for the header routing, the spawn-path wiring, percent-bearing and non-ASCII encoding; typecheck and `check:ci` clean; CI green on this PR.


dhaern:
Follow-ups from an adversarial audit of this PR:

- The cmux readiness gate built the same status URL without the directory it already received; it now passes it, same as the generic gate.
- New test coverage asserts the child's directory reaches the readiness probe on both spawn paths (fresh attach and busy respawn).
- Documented for later: the idle-close pollers (`fetchSessionStatuses`, cmux session lifecycle) poll without a directory, so their poll fallback stays blind to sessions in other project directories; the primary `session.idle` events still cover them. The opencode core also double-decodes the `directory` query param in `instance-context.ts` (pre-existing, newly reachable by this param) and `/session/status` only declared the query in its schema since 2026-05-09, older servers would 400 it.

Diff for this commit: +1/−0 in production code, +12/−3 in tests. Multiplexer suite 266/266, typecheck and `check:ci` clean.

dhaern:
Both findings addressed:

- Older servers: a 400 on the status probe now retries without the directory param and keeps it off for the remaining attempts, so pre-schema servers fall back to the old behavior instead of timing out a live session (the cmux gate gets the same fallback).
- Percent-bearing paths: the param is skipped when the directory contains a literal `%`, since the core decodes the value twice and would route to the wrong instance. Those paths keep the pre-PR behavior; the real fix is the double decode upstream (`instance-context.ts`), worth an upstream report.

The cmux default readiness parser turned out to be a byte-for-byte duplicate of the shared one, so the cmux gate now reuses the shared check and helper.

Diff for this commit: +26/−23 in production code, +65/−0 in tests. Multiplexer suite 269/269, typecheck and `check:ci` clean.

dhaern:
Both P1s are fixed at the root instead of mitigated: the probe no longer sends the `directory` query parameter. It routes by the `x-opencode-directory` header, pre-encoded with `encodeURIComponent`, which is the exact contract the official OpenCode SDK applies to every request (`packages/sdk/js/src/client.ts` and `v2/client.ts`).

- Older servers: the header is read by the workspace-routing middleware and is not part of any endpoint schema, so no server version can reject the request with a 400; versions predating the header ignore it and keep the old behavior. The 400 fallback is deleted because there is nothing left to reject.
- Percent-bearing and non-ASCII paths: the core reads the header raw and decodes it exactly once, and our value is encoded exactly once, so every path resolves to itself (`decodeURIComponent(encodeURIComponent(x)) === x`). The skip guard is deleted because those paths now route correctly.

The cmux gate reuses the same helper and the shared readiness parser (its byte-for-byte duplicate is gone), and the session-manager wiring coverage is unchanged.

Diff for this commit: +30/−24 in production code, +116/−55 in tests. Multiplexer suite 270/270, typecheck and `check:ci` clean.

PR #1234, opened 2026-09-19
fix(apply-patch): align pre-native matching with native trim semantics

### LINKED ISSUE #1207: [Bug]: apply_patch pre-verification rejects leading-whitespace matches accepted by native OpenCode

The `apply_patch` pre-native verification hook rejects a patch when the expected old lines differ from the file only by leading whitespace. Native OpenCode v1.18.31 accepts the same sequence through its full `trim()` matching pass.

Because the slim hook fails with `verification_failed` before native execution and uses `failOpen: false`, native OpenCode never gets an opportunity to process the patch.

Expected: the pre-native hook should preserve compatibility with native matching, or the intentional divergence should be documented/configurable. A safer fix than broad fail-open may be to run native-compatible matching passes before the additional stale-patch rescue logic.

## Steps to reproduce

1. Create `fixture.txt` with leading indentation:

   ```text
     alpha
     beta
     gamma
   ```

2. Verify an update whose expected old lines omit that indentation:

   ```diff
   *** Begin Patch
   *** Update File: fixture.txt
   @@
   -alpha
   -beta
   -gamma
   +alpha changed
   +beta
   +gamma
   *** End Patch
   ```

3. Compare the matchers:
   - Native OpenCode v1.18.31: accepts via the full `trim()` pass.
   - oh-my-opencode-slim 2.2.20: rejects with `verification_failed`.
   - oh-my-opencode-slim 2.2.21: rejects with `verification_failed`.

This was reproduced in a fresh temporary harness that loaded the matcher/rewrite code directly from each versioned bundle. The fixture was recreated for every case and no long-lived OpenCode/plugin process or cache state participated.

## oh-my-opencode.json

Not applicable. The reproduction invokes the matcher directly and is independent of user configuration.

## OpenCode version

1.18.31

## oh-my-opencode-slim version

2.2.20 and 2.2.21

## Operating system

macOS 26.6.2

## Logs, screenshots, or extra context

Observed slim error:

```text
ApplyPatchError
kind: verification
code: verification_failed
rewriteStage: before-native
failOpen: false
```

Native reference:

- Tag: `v1.18.31`
- Commit: `014614d35b397775e5d397a490fc72368c894ec2`
- File: `packages/opencode/src/patch/index.ts`
- `seekSequence` performs exact, `trimEnd()`, `trim()`, then Unicode-normalized `trim()` passes.

Slim currently keeps full-trim comparators out of automatic canonicalization because indentation changes can be semantically unsafe. That safety rationale is understandable, but it creates a pre-native behavioral divergence: slim blocks a patch that native OpenCode would accept. Prefix/suffix and LCS rescue did not recover this fixture.

Related but not duplicate: #918 covers aggregate model-correlated patch failures, not this deterministic matcher divergence. PR #406 adds stale-patch rescue and trailing-whitespace/Unicode tolerance but does not cover leading whitespace.

dhaern:
Fixes #1207.

The apply_patch pre-verification rejected patches whose expected lines differed from the file only in leading whitespace. Native OpenCode accepts those through its trim() matching pass, and since the hook runs with failOpen: false, the patch failed with verification_failed before native could apply it.

The comparator chain used by verification excluded the full-trim comparators. matching.ts already defined `trim` and `unicode-trim`, but they sat in a separate test-only chain; `trimEnd` ignores trailing whitespace only, so an unindented pattern could never match an indented file line.

This PR adds both comparators to the verification chain and to the prepared-target forms, so the gate accepts what native accepts. The rewrite machinery then anchors the match to the real file line: the rewritten patch inserts the new content and removes the actual line with its original indentation.

One behavior change to call out: the old chain was stricter than native on purpose and pinned a "blocks a dangerous indented case" test. That strictness is what produced the false verification failures in the issue, so the gate now follows native semantics. Content drift is still caught; the exact and unicode passes are untouched.

The duplicate test-only comparator chain (`comparatorEntries` / `permissiveComparators`) is gone as part of this.

Diff: +23/−12 in production code, +56/−50 in tests.

Validation: apply-patch suite 151/151, full suite 3437 passing (the one failure is the known flake from #1218, green in isolation), typecheck and `check:ci` clean.


dhaern:
Fixed: the prefix/suffix rescue loop now iterates a conservative comparator slice (exact, unicode, trim-end, unicode-trim-end), so fuzzy rescues can no longer anchor across indentation levels. Direct contiguous matching keeps the full native-compatible chain, and a new test pins the conservative behavior on indentation-only edges (`rescueByPrefixSuffix` returns a miss there).

Diff for this commit: +9/−1 in production code, +11/−0 in tests. Apply-patch suite 152/152, typecheck and `check:ci` clean.

PR #1235, opened 2026-09-20
feat(tui): emergency kill-all keybind for running subagents

dhaern:
## What it does

`alt+w` (or `/killall`, or the command palette) aborts every running subagent of the conversation you are looking at. Each target is aborted on its own, so one failure never blocks the rest, and a toast reports the summary. Aborted panes close through the multiplexer idle path, the same route a normally-finished subagent takes. Reconciliation afterwards is the existing event pipeline's job; this adds no state, polling, or endpoints.

## Why alt+w

The first keybind was shift+escape. Testing showed the host keymap matches modifier-less escape bindings against shift+escape presses, so every kill also armed the native "press escape again to abort" on the main pane. That is host behavior a plugin cannot intercept, so the key moved to alt+w, which has no bindings in any mode layer.

## Targets

- Only sessions with status busy or retry, from the sidebar targets of the visible conversation root
- The root itself is never a target. tui-state.json has carried self-referencing parent entries in real data (`ses -> ses`), and with one of those on the main session the parent-defined filter let the kill abort the orchestrator before the subagents. Regression test included with that exact shape
- Other conversations and the home route kill nothing

## Client shapes

The v1 and v2 clients take different abort paths, discriminated at runtime by the `v2` accessor (the only marker exclusive to the v2 SDK): v2 aborts take flat `{sessionID, directory}` params, v1 takes nested `{path}`.

## Diff

Production: +248 / −2 (`src/tui-kill.ts` +125 new, `src/v2/tui.ts` +68/−1, `src/tui.ts` +55/−1)
Tests: +347 / −0 (`src/tui.test.ts` +262, `src/v2/tui.test.ts` +85)

## Validation

- 90/90 tests in the focal suites (tui + v2/tui), full suite clean except the known #1218 flake, `check:ci` green
- Live test with three subagents running (two explorers, one librarian): one alt+w press aborted all three while the main conversation stayed alive mid-turn, with no interrupt warning
- Earlier live runs confirmed aborted panes close via multiplexer idle events


dhaern:
@greptile-apps Both findings are addressed in d65f11f5:

**P1 (failed aborts count as success):** you were right that the SDKs resolve rejected aborts rather than throwing. There is now a shared `sessionAbortFailure` helper in `src/utils/session.ts` that reads error envelopes and explicit false results, and both abort paths validate through it: the v2 flat-params call in `tui-kill.ts` checks the resolved envelope before counting the kill, and `abortSessionWithTimeout` (the v1 path) now throws on a rejected envelope too, so every consumer of that helper gets the same guarantee. A resolved error envelope, a false result, a thrown error, and a missing client all count as failures now.

**P2 (abort failures not logged):** the single per-session catch logs the failed session id together with the cause (`[tui-kill] abort failed for <sessionID>: <reason>`), so every failure the toast reports is identifiable in the plugin log.

New tests cover the three rejection shapes on both client shapes (resolved error envelope on v1 and v2, explicit false), unit tests cover `sessionAbortFailure` itself, and `abortSessionWithTimeout` has a rejection test for the error-envelope case. Focal suites 101/101, `check:ci` green. Please re-review.

PR #1236, opened 2026-09-20
fix(task-session-manager): drop unknown task_id and spawn a new session

ScoFan-official:
## Summary

`task()` no longer refuses unknown aliases with refuse-without-spawn. If `task_id` is not on the board and is not a `ses_*` session id, Slim deletes the id and lets OpenCode create a new child session.

This fixes the OpenChamber empty Agent Task card (`No subagent session id on task metadata.`) when the orchestrator redispatches `@fixer` with a stale/unknown alias.

## Behavior

- Unknown alias (`fix-99`, UUID, non-`ses_*`) → drop `task_id`, spawn new session
- Unknown `ses_*` still treated as resume (board miss / rehydrate lag)
- Known `running` / `stopped` / unreconciled jobs still refuse
- `task_result` / `task_revive` / `task_message` still refuse unknown ids

## Test plan

- [x] `bun test src/hooks/task-session-manager/index.test.ts src/v2/setup.e2e.test.ts src/tools/task-message.test.ts src/tools/task-status.test.ts` — 203 pass
- [x] Live: Slim log `dropped unknown task_id; spawning new session` for `fix-smoke-unknown-alias`, child `ses_*` created
- [x] Live: `task_result` on unknown alias still errors

Based on `v2.2.21`.


dhaern:
Rebased onto `master` (2.2.22) — the branch now keeps #1228's UUID-hallucination branch, so hallucinated UUIDs drop as before and only unknown aliases reach the new drop path.

We also addressed the P1: after a plugin restart the board can lose an alias mapping while the original child session is still running, and silently spawning a fresh child in that window would duplicate live work and lose the specialist's context. `task()` now consults a host probe (session list + status map) before dropping an unknown alias — if a child of this conversation is still busy, the call refuses with an actionable message instead, and the probe fails closed. Hosts without the list API degrade to the plain drop.

Verification: 212 tests passing across the TSM suite, v2 e2e and task-message/task-status; `bun run check:ci` clean.

dhaern:
Both points addressed in `572df5de→` (amended):

- **Tracked children no longer block spawns**: the probe now excludes children the board already tracks (they were already covered by the known-running refusal upstream) and only considers untracked ones.
- **Retrying children count**: the status check now treats `busy` and `retry` as alive, so a retrying lost child still refuses the drop instead of being duplicated.

Same verification as before: 212 tests across TSM/v2-e2e/task-message/task-status, `bun run check:ci` clean. Diff stays minimal — no new tests added for this adjustment.

dhaern:
The two earlier P1s were already fixed in `66609dd2` (probe excludes board-tracked children and counts `retry` as alive) — that review landed on the previous head.

The latest head (`c347db84`) additionally removes a machine-local `node_modules` symlink that accidentally slipped into a commit. Current diff is exactly the 4 source files, nothing else.

@greptile-apps could you re-review the current head?

PR #1240, opened 2026-09-21
feat(multiplexer): move pane lifecycle into the TUI client

### LINKED ISSUE #1214: [Bug]: Multiplexer panes never spawn, spawn in the wrong tab, or are disabled entirely under a shared server (opencode serve + attached clients)

expnn:

Subagent panes are now owned by the **TUI client that displays the parent session**. The OpenCode server no longer creates, closes, or positions panes, and never reads multiplexer environment variables.

This rewrites the multiplexer integration for the deployment where one long-lived server (`opencode serve`) is shared by several attached TUI clients (`opencode attach --dir <project>`), possibly for different projects and on different machines. The previous design ran pane management inside the server process. That process has no UI and is not inside the user's multiplexer, so it cannot know where a pane should go, or reach the terminal at all. The result was the failure family in #1214: panes that never appear, panes that appear next to the wrong client or at the focused pane, and whole `multiplexer.type` x environment combinations that silently do nothing.

67 files changed, +8856/-11344 against master. The server-side session manager and the cross-process pane registry are deleted; the lifecycle core now lives in `src/multiplexer/client/`.

## Why #1233 was not a complete fix for #1214

PR #1233 (merged; it closed #1214) scoped the readiness probe to the child's project directory, and nothing else. It is a real fix for one symptom, but the other two symptoms, and part of the first, remain reproducible on master:

| #1214 symptom | After #1233 on master | Why the fix cannot cover it |
|---|---|---|
| 1. Readiness gate stuck - no pane ever created | Fixed for the initial-spawn and busy-respawn readiness probes | The idle-close pollers (`MultiplexerSessionManager.fetchSessionStatuses`, `CmuxSessionLifecycle.loadStatuses`) still query `/session/status` without a directory. On a shared server their fallback path cannot see children in other project directories: panes are retained instead of closed, and a queued idle close can no longer be cancelled when the child turns busy again. |
| 2. Panes spawn next to the wrong client | Unchanged | Placement still resolves in the server: herdr uses the server's `HERDR_PANE_ID` or `--current`; zellij uses the tab of the server's `ZELLIJ_PANE_ID` or the focused tab; kitty needs `KITTY_LISTEN_ON`, which only exists in a client process; cmux anchors on an in-process previous/root surface; tmux reads a client-written registry, but only when the server itself runs inside tmux, and it is a single last-writer-wins entry. |
| 3. No pane for many `type`/environment combinations | Unchanged | Enablement is decided once, in the server process (`shouldEnableMultiplexer`, `getMultiplexer`, and the manager's self-gate): `tmux` while the server runs outside tmux -> feature disabled; `auto` while the server is outside every multiplexer -> disabled; `auto` while the server sits in herdr and the client in tmux -> the wrong adapter is chosen. |

The root cause is not in any adapter: the addressing information needed to drive a multiplexer (its control plane) exists only in a process running inside that multiplexer, and the server is not that process. A server-side probe can be patched case by case; the class cannot. This PR moves the executor to where the information and the control plane actually are.

## Architecture: client-owned pane lifecycle

Two physical facts drive the design:

1. **The control plane is process-local.** `TMUX` / `TMUX_PANE`, `ZELLIJ_SESSION_NAME` / `ZELLIJ_PANE_ID`, `HERDR_PANE_ID`, `KITTY_LISTEN_ON`, `CMUX_TUI_SOCKET` are only present in a process running inside that multiplexer. A remote server can never have them.
2. **A pane is a view.** Dispatch happens only on the server (the host `task` tool creates the child session); a pane is one viewer's window onto that child. Views and dispatch are independent - N panes never mean N dispatches.

Invariants:

- The server never creates, closes, or positions a pane and never reads multiplexer environment variables; the server entry's import graph must not reach the client lifecycle core or any adapter (enforced by `src/dependency-contract.test.ts`).
- No cross-process files at runtime: nothing writes pane locations for another process to read.
- Fail-closed: anything a client cannot determine ends in "no pane" plus a structured reason - never a guess.
- The pane command is a pure view: `opencode attach <serverUrl> --session <childSessionId> --dir <directory>`. It never creates a session and never sends a prompt.

Module layout (`src/multiplexer/client/`):

- `lifecycle.ts` - the pane lifecycle core: event filtering, readiness, per-client dedup, stable-idle close, respawn, reconnect backfill. All IO (clock, status reader, session list, adapter factory, server URL) is injected.
- `tui-wiring.ts` - the TUI host adapter: event subscription and projection, admission, config load, server-URL reflection, status/list readers through the host client (carrying the session directory), adapter factory, pane titles.
- `sweep.ts` + `pane-title.ts` - best-effort cleanup of panes a crashed client left behind, identified by title metadata (`omosc:<owner pid>:<child session id>`).
- `diagnostics.ts` - structured no-pane reasons and creation records.
- `ports.ts`, `types.ts` - the injected IO surface and shared types.

The tmux / zellij / herdr / kitty / cmux adapters keep their command surface but now run in the client, anchoring on the client's own environment and addressing their own socket or session explicitly (`-S <socket>` taken from `TMUX`; `--session <name>` taken from `ZELLIJ_SESSION_NAME`).

## How a pane is created now

1. **Admission (client-local).** The client reads the project's `multiplexer.type` and its own environment: `none` -> off; `auto` -> whichever supported multiplexer this client is inside; an explicit adapter that does not match the client's environment -> off with exactly one diagnostic.
2. **Event filter.** A `session.created` event is eligible only if the child belongs to this client's project directory, its `parentID` is the session this client currently displays, and admission passed. Status/idle/deleted events are still processed for children this client already holds a pane for, so switching the displayed session can never strand a pane.
3. **Readiness.** Before spawning, the client polls `/session/status` with the child's directory, with bounded retries (~2 s total). Timeout -> no pane, reason `readiness-timeout`.
4. **Anchor and split.** The pane is created by splitting **the pane/tab/window this client itself is running in**, re-resolved at spawn time - never a remembered position and never the globally focused one. If the parent session is moved and the child turns busy again, the rebuilt pane follows the new position.
5. **Dedup.** One pane per child per client, enforced by process-local state; duplicate events, event replay, and reconnect backfill cannot produce a second pane. Different clients each keep their own pane for the same child (per-client view semantics), with no coordination and no extra dispatch.
6. **Close.** Session deleted -> immediate close. Stable idle -> close after a debounce (default 5 s) with a final re-check; turning busy inside the window cancels the close.
7. **Rebuild.** A closed child that turns busy again is rebuilt, if its parent is still displayed by this client, at the then-current display position.
8. **Reconnect.** After an event-stream gap, the client reconciles against `session.list` (filtered by `parentID`): missing eligible children are rebuilt, terminated ones closed.
9. **Diagnostics.** Every skipped pane logs a distinguishable reason (`admission-none`, `admission-mismatch`, `host-unreachable`, `readiness-timeout`, `adapter-not-found`, ...); every success logs child, parent, adapter, pane id, and anchor.

Configuration surface: `multiplexer.type` (admission), `multiplexer.layout`, `multiplexer.main_pane_size`. Invalid values disable pane management (fail-closed) with a once-per-process diagnostic. The legacy top-level `tmux` block is ignored with a warning.

## Multi-client view sync (verified on tmux, zellij, herdr)

With one server and several clients attached to the same session - even clients sitting in different multiplexers - a subagent dispatched from any of them appears in every viewing client at once:

- dispatch still happens exactly once, on the server; which client sent the prompt is irrelevant;
- every client that displays the parent session receives the same `session.created` event and opens its own pane for the child, split from its own parent pane, inside its own multiplexer;
- the same rule drives the lifecycle: each client closes its own pane on stable idle, and rebuilds it when the child turns busy again while the parent is still displayed.

Verified by hand with three clients in tmux, zellij, and herdr attached to one session on one machine: a subagent dispatched from any one of them opened a pane in all three, each anchored next to its own parent pane. The previous server-side design could not do this - it resolved placement from the server process's own environment and created at most one pane, in whatever terminal the server happened to live in; clients elsewhere saw nothing, or the wrong terminal received the pane.

## Behavior changes and removals

- `multiplexer.zellij_pane_mode` and the agent-tab mode are removed. Zellij panes always open in the tab containing the parent pane; the old key is ignored with a one-time deprecation warning.
- The tmux pane registry is gone; no runtime cross-process files remain.
- Background-job deferred close is replaced by the stable-idle debounce.
- cmux is rewritten for the new-generation TUI (`CMUX_TUI_SOCKET`); the old-generation macOS app is out of scope. Cooldowns, close budgets, deferred spawns, and `equalize` are removed.
- Layout and close operations only ever touch panes this client created; kitty never modifies the active tab.
- Bare `opencode` (embedded server, no TCP listener) is fail-closed with exactly one `host-unreachable` diagnostic - start with `--port <port>` or use `serve` + `attach`.
- Per-client views: the same child can have one pane per client that displays the parent session.

`docs/multiplexer-integration.md` documents all of this, including a full behavior-changes and migration table.

## Response to review findings

All five findings were re-checked against the code. Four were valid and are fixed in `0ab0d431`; one is a real limitation that this PR does not fix, now documented instead of left implicit.

| Finding | Verdict | Resolution |
|---|---|---|
| P1 - reachability probe sends the directory as a `?directory=` query parameter | Valid | **Fixed**: the probe now sends the pre-encoded `x-opencode-directory` header, which the server reads raw and decodes exactly once (`instance-context` middleware) and which has been accepted since v1.0.74. The `?directory=` query parameter is decoded twice (`URLSearchParams.get` plus the instance-context decode), which corrupts directories containing literal `%XX` sequences (verified against the v1.18.30 source and reproduced on v1.14.30 / v1.14.44 / v1.18.31). |
| P1 - pane wiring keeps the old project after a TUI project switch | Valid, but gated behind an experimental feature | In v1.18.x `api.state.path.directory` never changes on a default install (`/sessions` is scoped to the current directory; `/move` moves the session, not the TUI). It changes only with experimental workspaces (`OPENCODE_EXPERIMENTAL_WORKSPACES`): `/warp`, deleting the current workspace via `/workspaces`, or opening a session from another workspace. Re-scoping needs config/admission re-evaluation plus per-pane directory ownership; recorded under Known Limitations with a follow-up. |
| P1 - reconcile uses a stale displayed session | Valid | **Fixed**: `runReconcile` re-reads the displayed session from the route before diffing children, so navigating during a quiet event stream cannot reconcile against the previous parent. Regression test added. |
| P1 - busy event cannot stop an in-flight idle close | Valid | **Fixed**: every held-pane event bumps an activity epoch; the pre-close status read captures it and aborts the close when the epoch moved, so a stale idle snapshot cannot close a pane whose child just turned busy. Regression test added. |
| P2 - doctor misses non-object `multiplexer` shapes | Valid | **Fixed**: doctor re-validates any present `multiplexer` value, including the non-object shapes the runtime sanitizer turns into `type: "none"`, so the diagnostic can explain why panes are disabled. Regression test added. |

## Testing

- `bun test`: 3627 pass, 2 skip, 0 fail (177 files).
- `bun run typecheck` and `bun run check:ci`: clean.
- New suites cover the lifecycle core (dedup, stable-idle debounce, respawn, reconnect backfill, fail-closed paths), TUI wiring (event projection, admission, status/list readers, server-URL handling), the crash sweep, pane titles, and diagnostics. A dependency-contract test pins the server/client process boundary with a documented mutation check. Review-driven regression tests cover the probe header encoding, the reconcile displayed-session refresh, the busy-during-final-read close abort, and the doctor non-object `multiplexer` shape.

## Related

Fixes #1214. #1233 addressed only the readiness probe; this PR replaces the server-side execution model that it patched.


expnn:
Superseded by a fresh PR opened from the same branch to trigger a new review pass. The code is unchanged (one commit message reworded).

PR #1241, opened 2026-09-21
fix: reconcile V2 terminal outcomes

kumar-shivang:

OpenCode V2 persists a completed session's terminal outcome as idleOutcome (and some host shapes expose idle_outcome), while Slim only inspected outcome. This left completed child sessions in the background-job board as running, so task_message could accept a noReply message that would never run.

This change accepts both V2 outcome field spellings, allowing successful children to reconcile into reusable sessions. It also returns actionable task_message guidance for terminal and retained tasks, and adds the correct continuation path to the orchestrator prompt.

Verification:
- bun test src/agents/orchestrator.test.ts src/utils/background-job-terminal-gate.test.ts src/tools/task-message.test.ts (123 passing)
- bun run typecheck
- bun run build
- bun run check:ci
- Independent review: approved


mhenke:
@kumar-shivang  please pull in master/push and the failing ci pipeline will pass.

Issue #1247, opened 2026-09-21
[Bug]: image-hooks.ts doesn't work with opencode v2.0.11


haoliangwu:
### What happened, and what did you expect?

I upgraded opencode v2 today and it seems like everything of omo-slim works fine with it besides the `observer` subagent. 

The `observer` works well with opencode v1 and can be dispatched normally once the user prompt contains any images but fails with 400 error in opencode v2 if the model is text only type.

After debugging for a while, I realize the root cause is related to `image-hook.ts` and image part of message payload.

### Steps to reproduce

1. upgrade to opencode v2 with omo-slim installed
2. enable the observer
3. start a session and paste image
4. enter
5. 400 error occurs, the error message complains the model only supports text type message balabala...

### oh-my-opencode.json

```jsonc
"opencode-go": {
      "orchestrator": {
        "model": "opencode-go/glm-5.3-flash",
        "variant": "high",
      },
      "oracle": {
        "model": "opencode-go/glm-5.3-flash",
        "mcps": ["*"],
      },
      "council": {
        "model": ["opencode-go/glm-5.3-flash"],
        "variant": "high",
      },
      "librarian": {
        "model": "opencode-go/mimo-v2.5",
        "mcps": ["websearch", "context7", "grep_app", "cdp-bridge", "anysearch"],
      },
      "explorer": {
        "model": "opencode-go/mimo-v2.5",
        "mcps": ["codegraph"],
      },
      "designer": {
        "model": "opencode-go/deepseek-v4-flash-vision-exp",
      },
      "fixer": {
        "model": "opencode-go/deepseek-v4-flash",
        "variant": "high",
      },
      "observer": {
        "model": "opencode-go/mimo-v2.5",
      },
    }
```

### OpenCode version

2.0.11

### oh-my-opencode-slim version

2.2.22

### Operating system

macOS Tahoe 26.6.2 

### Logs, screenshots, or extra context

```shell
## Before (OpenCode v1)

Image parts were self-describing data URLs on a single `url` string:


{ "type": "image", "url": "data:image/png;base64,<...>", "filename": "shot.png" }

File parts could also carry images, detected via the mime field:

{ "type": "file", "mime": "image/png", "..." : "..." }


`isImagePart()` correctly matched both `(type === 'image', or
type === 'file'` with an `image/*` mime / image file extension), and bytes
were recovered by parsing the data URL (decodeDataUrl).

## After (OpenCode v2)

[...]

PR #1248, opened 2026-09-21
feat: intercept OpenCode v2 media image parts in image hook

haoliangwu:
v1 path unchanged (guarded by regression test)，v2 path is additive。

see more information at https://github.com/alvinunreal/oh-my-opencode-slim/issues/1247.


alvinunreal:
thank @haoliangwu

PR #1253, opened 2026-09-21
fix(foreground-fallback): fall back on provider billing/quota rejections

dhaern:
Follow-up to #1135. That PR made deterministic content-policy rejections fallbackable. This extends the same treatment to billing and quota rejections, which currently kill the whole foreground task.

## Symptoms

When an xAI (Grok) account hits its spending limit, the provider returns a deterministic rejection and the task dies without walking the fallback chain:

```
AI_APICallError: personal-team-blocked:spending-limit: You have run out of credits or need a Grok subscription. Add credits at https://grok.com/?_s=usage or upgrade at https://grok.com/supergrok.
```

The OAuth token is valid (it refreshed on the same request), so this is billing, not an auth failure. Retrying the same model can never succeed, but the next model in the chain runs on a different provider that does not share the balance, so it should be tried instead.

## Why they are fallbackable

Same reasoning as the content-policy case in #1135: the error is deterministic for the same account, and a different provider does not share the quota. Only the error class changes: billing instead of policy.

## Changes

`RETRYABLE_ERROR_PATTERNS` gains:

- xAI billing: `/\bpersonal-team-blocked\b/`, `/\bspending.?limit\b/i`, `/\b(?:ran|run) out of credits\b/i`. The third one also matches the official xAI fixture wording ("You ran out of credits. [WKE=personal-team-blocked:spending-limit]").
- Zhipu GLM quota codes `/"1113"/`, `/"1308"/`, `/"1309"/`, `/"1310"/` (balance exhausted, coding-plan usage limit, expired plan, weekly/monthly limit). The English messages already match the existing quota wording, so the quoted JSON codes cover the Chinese wire variants (`余额不足或无可用资源包`) and Anthropic-style `{"type":"1113"}` envelopes where no English text survives. Sources: the official error-code table at docs.z.ai and real wire captures from public issues.
- GLM English wording not covered yet: `/package has expired/i`, `/limit.?exhausted/i`.

`isFailoverError` also accepts `402` (Payment Required) as a retryable status.

One note on xAI: the status code is unreliable there. The same billing error has been observed as 402, 403, and body `code: 429`, so the string signature does the classification. 402 stays in the status set as the standard billing class for providers that use it consistently.

## Tests

- xAI spending-limit rejection as a raw `AI_APICallError` string and as a `data.statusCode: 400` shape
- The official xAI fixture variant (`ran out of credits` with the `WKE=` code)
- 402 with an innocuous message, pinning the status-code path on its own
- GLM codes `1113` and `1308` with Chinese messages, the expired-plan English message, and `Weekly/Monthly Limit Exhausted`
- Negative case: ordinary wording that merely mentions "credits" stays a hard error

Validation: `bun test src/hooks/foreground-fallback/index.test.ts` passes 127 tests, full suite 3666 pass, biome clean.


dhaern:
@greptile-apps Fair point. Both generic phrases are now anchored to the provider wording they were added for:

- `/package has expired/i` → `/\bcoding plan package has expired\b/i` (the GLM 1309 message)
- `/limit.?exhausted/i` → `/\b(?:weekly|monthly) limit exhausted\b/i` (the GLM 1310 message)

Unrelated failures such as "file descriptor limit exhausted" or an expired TLS certificate stay hard errors now, and both are pinned with negative tests. The structured codes and the xAI signatures are unchanged. Validated with the focal suite (128 pass), full suite (3667 pass), and biome clean on ab0fd0a6. Please re-review.

PR #1257, opened 2026-09-22
perf(foreground-fallback): read only the transcript tail for the fallback replay

dhaern:
Follow-up to #1253 (same failover flow; the touched region is disjoint from the billing patterns, so both merge cleanly).

## Symptoms

On long-lived sessions, every failover waited ~20 s between the provider error and the model switch. Host log of a live months-old orchestrator session:

```
[02:20:54.908] stream error  modelID=gpt-6-astra  "All 1 account(s) are rate-limited..."
[02:20:56.438] loop step=0   (fallback replay run starts)
```

The delay was not the fallback logic (the abort fired 0.09 s after the error). The cause is in `execFallback`: it re-read the FULL transcript (`session.messages({ path })`) just to find the last replayable user message. On that session the endpoint serves 463 MB in 11.7 s, and the JSON parse lands on top.

## Change

`execFallback` now reads only the transcript tail:

- `session.messages({ path, query: { limit: FALLBACK_REPLAY_TAIL_MESSAGES } })` with `FALLBACK_REPLAY_TAIL_MESSAGES = 50`
- If the tail contains no replayable user message, it pays the full read instead (hosts that ignore `limit`, exotic transcripts). Never worse than before.
- `requestError` in the "no user message found" log aggregates tail and full-read errors.
- The observation-handoff `baselineMessageID` (#1231) reads from the same array; it only needs the trailing message id, which the tail always contains.

The host paginates `/session/{id}/message` with `desc(time_created), desc(id)` plus a reverse, so `limit` is by contract the LAST N messages in ascending order.

## Measured impact

| Read | Time | Size |
| --- | --- | --- |
| Full transcript (before) | 11.74 s | 463 MB |
| Tail `limit=50` (after) | 0.015 s | 19.7 KB |

Live failover on the same session went from ~20 s error-to-switch to ~1.5 s.

## Tests

- Tail-only path: exactly one `messages` call, `query.limit === 50`, the replay uses the tail's user prompt, no full read issued.
- Fallback path: tail without a replayable user message, then a second full read, replay uses the full read's user prompt.
- The mock client gained a `messagesImpl` override mirroring the existing `promptAsyncImpl` pattern.

Validation: 126/126 focal tests, 3665 tests across the suite, `bun run check:ci` clean. An independent audit pass re-ran both and verified the pagination contract against the host source.


dhaern:
@greptile-apps Both findings are fixed in 51f9442e:

**P1 (V2 Discards Transcript Limit):** the shim now honors the bound. When `session.messages` is called with `query.limit` and the host exposes the paginated v2 route, `src/v2/client-shim.ts` calls `session.messages({ sessionID, limit, order: 'desc' })` and reverses the page back to the v1 ascending shape, so the transfer stays bounded on the v2 path too. Hosts without that route keep the full context read, and a call without `query.limit` still reads the full transcript as before. `V2Context['session']` now declares the optional `messages` member (runtime-probed capability). Two new shim tests pin the contract: the bounded call forwards `{ limit: 50, order: 'desc' }` to the paginated route and never touches `context`, returning items in ascending order; the no-limit call stays on `context`.

**P2 (Tail Error Gets Lost):** `requestError` no longer overwrites the tail error. When both reads fail, the "no user message found" log payload becomes `[tailError, fullError]`; a single failure keeps its plain form. A new test drives the dual-error path end to end (two reads issued, no replay).

Validation: 175 tests across the touched suites and 3668 across the repository, `bun run check:ci` clean. Please re-review.

PR #1276, opened 2026-09-23
fix(v2): emit host-canonical *-resource rules from the permission bridge

### LINKED ISSUE #1244: v2: per-agent permission denies are not enforced in spawned subagent sessions (schema not stripped, execution not gated)
## Environment

- opencode v2.0.11
- oh-my-opencode-slim @latest (verified against source, v2.2.22)
- linux, CLI

## Summary

On v2 hosts, per-agent `permission` denies are **not enforced in spawned subagent sessions**: denied tools remain in the child's tool schema and are not blocked at execution time. Two mechanisms interact:

1. The per-child session permission bridge emits **exact-match rules that pin `resource` to the tool-name literal** — these never match real call resources (file paths), and in core's `whollyDisabled` findLast evaluation they **shadow the wildcard-resource deny** (`resource: "*"`) that static agent registration emits, so denied tools are not stripped from the schema.
2. The child's execution does not appear to be governed by the static agent permissions for non-matching resources, although the bridge's doc comment states they "keep governing everything the exact-match ruleset cannot express".

## Repro (no special config needed — omos's own defaults)

1. Default config. Spawn any non-orchestrator subagent (e.g. `explorer`).
2. omos `applyDefaultPermissions` denies `task_status` / `task_result` / `task_message` / `task_cancel` / `task_revive` / `wait_for_user` for every non-orchestrator agent.
3. Ask the subagent to list the tools in its function list.

**Actual:** all six denied tools are still present in the subagent's tool schema. Log shows the rules were applied:

```
[v2][permission-rules] applied exact-match rules to child session {"agent":"explorer","count":9}
```

4. Additionally, with explicit user config `presets.<p>.agents.<name>.permission = { "edit": "deny", "write": "deny", "apply_patch": "deny", "task": "deny", "webfetch": "deny" }` on `explorer`, the spawned subagent could still **successfully call `edit`** (the edit was applied to a file). So the "static agent permissions govern unmatched resources after the bridge replaces the inherited list" promise did not hold at execution time on this host.

## Mechanism analysis (source-cited)

1. Static registration emits wildcard-resource denies into the agent draft: `applyAgentToDraft` → `adaptPermissions` → `{action:"task_status", resource:"*", effect:"deny"}`.
2. Core `Tool.snapshot` (packages/core/src/tool.ts:231, `whollyDisabled` at :292) strips a tool only when the **last matching rule** (findLast, action-matched) has `resource === "*"` and `effect === "deny"`.
3. Core merges agent rules first, session rules second (`Permission.merge` = `rulesets.flat()`, permission.ts:99).
4. The per-child bridge `deriveExactPermissionRules` re-emits every whole-tool string effect as `{action, resource: <tool-name literal>, effect}`. In the merged ruleset those literal-resource rules come **after** the agent's wildcard deny, so for tool `task_status` the last action-matching rule is the bridge rule with `resource === "task_status" ≠ "*"` → `whollyDisabled` returns false → **not stripped**.
5. At execution, the literal resource never matches real call resources (file paths) → the bridge deny never fires. Whether the static agent deny then fires depends on how the host consults agent permissions vs session permissions for spawned children — empirically it did **not** (step 4 of the repro).

## Design note (why the literal-resource emission exists)

The test suite documents an incident (`explorer derived 0 rules`): read-only agents whose maps carry only whole-tool effects used to derive an empty ruleset, so the bridge left the parent's inherited (allow-all) session rules in place. The whole-tool emission exists to guarantee a **non-empty derived ruleset** so the bridge replaces the inherited list. Any fix must not regress that incident.

Jiajun0413:
## Summary

Whole-tool permission effects were emitted from the child-session bridge with the declared tool key as the literal resource (`{action, resource: tool}`). The literal never matches a real call resource, and core's `whollyDisabled` (packages/core/src/tool.ts:292) requires the last action-matching rule to carry `resource: '*'` before it strips a tool — so denied tools stayed in the child's schema, and inherited parent session rules (merged after agent rules, packages/core/src/session/context.ts:131) shadowed the agent's static denies at execution time (issue #1244, repro steps 3/4).

## Fix

Whole-tool effects now derive `{action, resource: '*', effect}` — the host-canonical form of the declaration's true scope, identical in shape to the static agent registration (`adaptPermissions`) and the Agent default ruleset. Nested pattern entries keep their exact resources. The drift-prone mirrored mapping (`exactActionsForV1Key`) is deleted in favor of the shared `v1PermKeyToV2`.

Not a widening: the `'*'` resource restates what a whole-tool declaration already means (allows/denies/asks all apply tool-wide); the previous literal form was the narrowed-but-inert expression. The non-empty-derivation guarantee (explorer-derived-0-rules incident) is unchanged — wildcard rules count toward non-empty, the bridge still replaces the inherited child list.

Fixes #1244.


alvinunreal:
thanks @Jiajun0413

Issue #1279, opened 2026-09-23
Auto-updater: parallel server processes race bun install on the same pending version (TOCTOU)


JoJohanse:
Follow-up from #1278 — the windowsHide fix removes the user-visible symptom; this tracks the wasted-work amplifier behind it.

## Mechanism

- `hasChecked` (`src/hooks/auto-update-checker/index.ts`) dedups only **within** a single server process.
- The desktop app runs one opencode server process per project/window. When a new plugin version is pending, each process's first top-level `session.created` passes the gate and runs its own `bun install` into staging.
- `getCachedVersion` (`src/hooks/auto-update-checker/checker.ts`) reads the **running module's** package.json, so every process still on the old version re-detects the update — including after another process already staged/published it — until each server restarts.

Observed on Windows 11 / OpenCode Desktop 2.0.14: 5 parallel `Update available` checks within the same second (5 open project windows), followed by further reinstall rounds *after* a successful publish, from servers still on the old version.

## Harm

- Redundant npm traffic and staging churn, multiplied by open windows
- Cross-window version flapping until all servers restart
- Each install is a real CPU/network cost even when the console window is hidden

## Suggested directions

- Cross-process install lock (precedent: `acquirePidFileLockWithRetry` in `src/companion/manager.ts`)
- After acquiring the lock, re-check whether the staged/published version already equals `latestVersion` before installing

PR #1280, opened 2026-09-23
fix(windows): hide child-process console windows on GUI hosts

### LINKED ISSUE #1278: Windows desktop: visible console window per bun.exe spawn (auto-update fires on session.created) — add windowsHide
## Environment

- OS: Windows 11 (win32), OpenCode **Desktop** 2.0.14
- oh-my-opencode-slim: 2.2.22 → 2.2.24 (observed on both)
- bun installed via npm; the real `bun.exe` is resolved by `crossSpawn`/`resolveWindowsCommand` (thanks to #1045 — that fix is what makes these spawns succeed, and incidentally what now makes them visible, see below)

## Symptom

In the desktop app, creating a new conversation and sending the first message makes several **visible console windows pop up**, one per spawned `bun.exe` process. With multiple project windows open they appear as a batch — plugin logs show 5 parallel `Update available` checks within the same second (one per opencode server process). The storm repeats on every new conversation until each running server has restarted onto the updated plugin version.

## Root cause

1. `auto-update-checker` triggers on the first top-level `session.created` per server process (`src/hooks/auto-update-checker/index.ts`).
2. If npm has a newer version, it runs `bun install` via `runBunInstallSafe` → `crossSpawn(["bun", "install"], { cwd, stdout: "pipe", stderr: "pipe" })`.
3. `crossSpawn` builds `spawnOptions = { stdio, cwd, env }` — **no `windowsHide: true`**.
4. The opencode server inside the desktop app is a GUI-subsystem process with **no console**. On Windows, spawning a console executable from a console-less parent allocates a new console → one visible window per spawn. (TUI users never see this because the child inherits the terminal's existing console.)

Amplifier: servers still running the old version re-detect the update on every `session.created` and re-run `bun install` until restarted, so windows keep popping on every new conversation during the "restart pending" window.

Evidence from one machine (plugin logs): 5× `Update available (latest): 2.2.22 → 2.2.24` at 01:13:36Z, 5× `bun install failed` ~11s later; success at 02:17:55Z; then again `Update available` + `bun install failed` at 02:39Z / 02:47Z from servers still running 2.2.22.

## Other spawns of the same class

- companion: `spawn(bin, [], { detached: true, stdio: "ignore" })` — on Windows a detached console child gets its own console window
- interview `openBrowser` on win32: `spawn("cmd", ["/c", "start", "", url], { detached: true, stdio: "ignore" })`
- pwsh detection: `spawnSync("where", ["pwsh"], { stdio: ["ignore", "pipe", "pipe"] })`

## Suggested fix

Add `windowsHide: true` to `crossSpawn`'s `spawnOptions` (it propagates to both the direct `nodeSpawn` branch and the `cmd.exe /d /s /c` branch), and to the three spawns listed above. It is a no-op on POSIX. Verified locally: patching the installed `dist/index.js` on 2.2.24 stops the windows completely.

Optionally, dedupe the reinstall: before running `bun install`, skip when the staged/published version already equals `latestVersion`, so parallel server processes don't each run their own install during the restart-pending window.

Happy to send a PR if you'd like.

JoJohanse:
The opencode desktop server is a GUI-subsystem process without a console. On Windows, spawning a console-subsystem executable (bun.exe, cmd.exe, where.exe) from a console-less parent makes the OS allocate a NEW console, so every plugin auto-update popped a visible console window — one per `bun install`, multiplied by open project windows (parallel server processes) and re-firing on each new conversation while an update was pending (#1278).

Set `windowsHide: true` (CREATE_NO_WINDOW on Windows, no-op elsewhere) at every runtime spawn site:

- **crossSpawn** (`src/utils/compat.ts`): covers all bun/tar/where/ast-grep/multiplexer spawns, including the auto-updater's `bun install` and the `cmd.exe /d /s /c` shim branch from #1045
- **ACP agent spawn** (`src/tools/acp-run.ts`): agents ran with a visible console for the whole session on desktop hosts
- **companion spawn** (`src/companion/manager.ts`): detached daemon
- **interview browser opener** (`src/interview/service.ts`): `cmd /c start`
- **pwsh detection** (`src/utils/zip-extractor.ts`): `where pwsh`

stdio stays piped/ignored everywhere; no child needs a visible console or a TTY. CLI-installer spawns (`src/cli/`) are intentionally untouched: they run attached to an interactive console.

Notes:
- `windowsHide` does not interfere with `detached`'s outlive-parent behavior (DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP), and libuv applies CREATE_NO_WINDOW only when no stdio fd is inherited — true at every site here (all pipe/ignore).
- The parallel-install amplifier (multiple server processes racing `bun install` on the same pending version) is a separate TOCTOU issue, filed as follow-up: #1279

## Verification

- New test `src/utils/compat.spawn-options.test.ts` locks the chokepoint: asserts `windowsHide: true` reaches both the direct spawn branch and the win32 `cmd.exe /d /s /c` shim branch (alongside `windowsVerbatimArguments: true`). `bun test src/utils/compat.spawn-options.test.ts src/utils/compat.test.ts` → 23 pass; `tsc --noEmit` clean; biome clean on changed files.
- Empirically verified on Windows 11 (OpenCode Desktop 2.0.14): applying the same flags to the installed dist stopped the window storm completely.

Closes #1278


alvinunreal:
thanks @JoJohanse

